PR #13, opened 2023-01-14
enhancement: Catch better errors in deploy.yaml

scart88:
I tried to improve the validations; I think there are plenty of better ways to do it, but this is one of the first times I'm doing a PR and working on a gem, so I hope everything is ok.

This is a start for the issue: https://github.com/rails/mrsk/issues/6

Signed-off-by: Drew <drew@xmail8.com>

### DIFF
--- a/lib/mrsk/configuration.rb
+++ b/lib/mrsk/configuration.rb
@@ -23,7 +23,7 @@ class Mrsk::Configuration
 
   def initialize(config, validate: true)
     @config = ActiveSupport::InheritableOptions.new(config)
-    ensure_required_keys_present if validate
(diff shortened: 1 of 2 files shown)


scart88:
@dhh I started a few validations for the `deploy.yml` and I'm not sure if you had any chance to review this before I closed the PR. I was wondering if it was anything good here and if I should start again in a new PR.

Thanks

dhh:
At first gloss, this looks a little too heavy to me. It also does puts within the model, which is a no-no. Let's hold off for a bit to revive.

scart88:
I understand. Thank you very much for the insights.

PR #21, opened 2023-01-21
Added support for volumes

chrisdebruin:


lib/mrsk/configuration.rb
```diff
@@ -61,7 +60,6 @@ def traefik_hosts
     roles.select(&:running_traefik?).flat_map(&:hosts)
   end
 
-
```
dhh, line 64: Extra spaces are intentional to separate sections of related methods.

lib/mrsk/configuration.rb
```diff
@@ -87,6 +84,12 @@ def env_args
     end
   end
 
+  def volumes
+    return unless config.volumes.present?
```
dhh, line 88: Turn into full conditional, like the other usage, so `if config.volumes.present?`.

lib/mrsk/configuration.rb
chrisdebruin: alright

lib/mrsk/configuration.rb
chrisdebruin: done

lib/mrsk/configuration.rb
```diff
@@ -87,6 +87,14 @@ def env_args
     end
   end
 
+  def volumes
```
dhh, line 90: This actually needs to be `volume_args` to match the others.

PR #23, opened 2023-01-22
Accessories

dhh:
Operate accessory services like MySQL and Redis via MRSK.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,9 +262,9 @@ accessories:
     volumes:
       - /var/lib/mysql:/var/lib/mysql
   redis:
-    image: redis:8.0
+    image: redis:latest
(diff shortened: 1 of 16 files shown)


northeastprince:
@dhh is this something 37signals is using in production and/or plans to? For something like a big MySQL database this doesn't seem complex enough to be a good solution for managing it.

dhh:
What complexity are you craving?

northeastprince:
Good point 😆, I guess it just seems too simple to handle something like a core database.

dhh:
Why? You mount your data using volumes that you should of course make sure are properly backed up. You can setup multiple, if you want full replication. Complexity does not make something more robust. Quite the contrary.

Issue #7, opened 2023-01-13
Allow bastion SSH server


dhh:
SSHKit can configure Netssh to use a bastion SSH server. We should change our setup so that passes through. We could change the top level ssh_user configuration option to be:

```yaml
ssh:
  user: app
  # whatever is needed for bastion
```

PR #24, opened 2023-01-27
Allow use of bastion host

chrisdebruin:
fix for #7


dhh:
Nice. Went with a single proxy definition. But we also need to deal with this in relation to the specific calls to ssh, like in follow_logs. Can you have a look at that?

chrisdebruin:
> Nice. Went with a single proxy definition. But we also need to deal with this in relation to the specific calls to ssh, like in follow_logs. Can you have a look at that?

Yep I'll look into it

chrisdebruin:
@dhh added proxy for specific calls to ssh

PR #27, opened 2023-01-28
added docker logging driver and options

azolf:


dhh:
Hmm. Think this is a little too low level. Lemme think about how best to expose. Also, need to have some sensible defaults.

azolf:
> Hmm. Think this is a little too low level. Lemme think about how best to expose. Also, need to have some sensible defaults.

Adding default values is easy, and we could put them there. also, we could define methods to use them as equivalent docker options.

dhh:
Appreciate the work here, but I really like the simplicity of the local log driver. Great defaults, all we need. Done here: https://github.com/mrsked/mrsk/commit/a80289d0460f6757a752186bfae42de10b19d6ec

PR #28, opened 2023-01-30
15 prune based on number

azolf:


dhh:
I know I suggested this one as well, but seeing what it takes, it feels like we're cutting against the grain. Docker wants to prune on time and use/dangling. I think we should just stick to that. The 3 day prune cut-off for containers seems fine, but we should lower the 30 day on images to 7 days. I'll do that. These are all duplicate images anyway. Everything lives in the registry, so they're only there to speed things up a little.

PR #29, opened 2023-01-30
Allow customizing traefik args

anoldguy:
Defaults aside, we want to configure specific options for traefik. This PR is a very crude passthru for command line arguments to traefik. 

```yaml
traefik:
  - --accesslog=true
  - --accesslog.format=json
  - --metrics.prometheus=true

# Generates
# --accesslog=true --accesslog.format=json --metrics.prometheus=true
```

Ultimately, we might want to restrict which ones we can pass via some more intelligent parsing of options like so:

```yaml
traefik:
  accesslog:
    format:json
  metrics:
    prometheus: true

# Would generate
# --accesslog=true --accesslog.format=json --metrics.prometheus=true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,18 @@ volumes:
   - "/local/path:/container/path"
 ```
 
+### Customizing traefik
+
+You can customize the traefik command line using `traefik`:
+
+```yaml
+traefik:
+  - --accesslog=true
+  - --accesslog.format=json
+  - --metrics.prometheus=true
+  - --metrics.prometheus.buckets=0.1,0.3,1.2,5.0
+```
+
 ### Using different roles for servers
 
 If your application uses separate hosts for running jobs or other roles beyond the default web running, you can specify these hosts in a dedicated role with a new entrypoint command like so:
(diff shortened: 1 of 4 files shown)


dhh:
Let's structure these like build args. Use a hash, argumentize, etc. See https://github.com/rails/mrsk#configuring-build-args-for-new-images and https://github.com/rails/mrsk/blob/main/lib/mrsk/commands/builder/base.rb#L10

anoldguy:
That's how I originally had it, using a hash, but the args don't all take the same name, so `argumentize` isn't a good fit as it exists right now. [`MRSK::Utils#argumentize`](https://github.com/rails/mrsk/blob/master/lib/mrsk/utils.rb#L4) explicitly takes an argument and repeats it with the various k/v pairs. Volumes and build_args are perfect examples.

```bash
--volumes "/foo:/bar" --volumes "/baz:/bat"
```

But traefik is, of course, a special case. Here, the access logging needs to be turned on with `--accesslog=true`, and then further configured with `--accesslog.format=json`. Same with several other options, like the aforementioned prometheus metrics. Turn on with `--metrics.prometheus=true`, but then further configure it with `--metrics.prometheus.buckets=0.1,0.3` or `--metrics.prometheus.addEntryPointsLabels=true`.

I had some special casing in traefik_args to handle that, but I don't think it's as straightforward as just using argumentize. See the CLI [docs for accesslog](https://doc.traefik.io/traefik/observability/access-logs/#access-logs) for what I mean, and why I opted for the more crude pass-thru.

Holding off on this, like we discussed offline while a more elegant way is investigated.

PR #31, opened 2023-02-02
Align config class name with file name

fschueller:
`Mrsk::Configuration::Assessory` -> `Mrsk::Configuration::Accessory` thus aligning with the name of the file.


dhh:
Good catch 👌

robzolkos:
Wonder if we should get Zeitwerk check in CI to catch these sorts of issues going forward?

dhh:
Haven't added Zeitwerk to MRSK yet. Happy to take a PR that does that.

PR #33, opened 2023-02-02
Change version terminology to tag

fschueller:
The rollback feature (and other things) are based on the respective image's
tag. While the tag can represent the app's version, we're still targeting an
image's tag and not an image version. This aligns with upstream terminology and
can help understand what's actually being targeted from an operating perspective.

https://github.com/rails/mrsk/issues/32


dhh:
Yes, we use the version as a tag, but we also use the version as part of the image name. And the version is derived from the specific _git hash_ we're treating as a version, not just any old tag. So the image tag is but a subset and a derivative of our version concept. And in another way, the version is a subset of what a tag can be, when in MRSK land it can only be one thing: the version 😄.

But I do think we should describe these fine distinctions in the README!

PR #34, opened 2023-02-02
Add zeitwerk to check file names match classes

robzolkos:
In https://github.com/rails/mrsk/pull/31 class names had to be updated as they didn't match the file name.

To prevent this from happening again we should do a sanity check in CI with Zeitwerk.  The Zeitwerk loader will error if classes are not named as per the filename.

I'm not sure if we should be explicit in requiring Zeitwerk as a dependency?  Zeitwerk is a transitive dependency of`railties` which is included in MRSK already.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -3,10 +3,23 @@ require "active_support/test_case"
 require "active_support/testing/autorun"
 require "debug"
 require "sshkit"
+require "zeitwerk"
 
 ActiveSupport::LogSubscriber.logger = ActiveSupport::Logger.new(STDOUT) if ENV["VERBOSE"]
 
 SSHKit.config.backend = SSHKit::Backend::Printer
 
 class ActiveSupport::TestCase
+  test "Zeitwerk compliance" do 
+    loader = Zeitwerk::Loader.for_gem
+    loader.setup
+
+    begin
+      loader.eager_load(force: true)
+    rescue Zeitwerk::NameError => e
+      flunk e.message
+    else
+      assert true
+    end
+  end
 end


lib/mrsk/cli/base.rb
```diff
@@ -1,5 +1,5 @@
 require "thor"
-require "mrsk/sshkit_with_ext"
+require "ssh_kit/backend/abstract"
```
robzolkos, line 2: I'm not sure if this is the right thing to do here?

We get a warning:

```
❯ bin/test
WARNING: Zeitwerk defines the constant SshKit after the directory

    ..../mrsk/lib/ssh_kit

To prevent that, please configure the loader to ignore it:

    loader.ignore("#{__dir__}/ssh_kit")

Otherwise, there is a flag to silence this warning:

    Zeitwerk::Loader.for_gem(warn_on_extra_files: false)
Run options: --seed 58503

# Running:

............................................................................................

Finished in 0.041901s, 2195.6517 runs/s, 3269.6117 assertions/s.
92 runs, 137 assertions, 0 failures, 0 errors, 0 skips
```

lib/mrsk.rb
```diff
@@ -1,3 +1,8 @@
+require "zeitwerk"
+
+loader = Zeitwerk::Loader.for_gem
```
intrip, line 3: From the [docs](https://github.com/fxn/zeitwerk#for_gem) adding `Zeitwerk::Loader.for_gem` should automatically introduce autoloading of the files under the `lib/` folder so I wonder if we still need to require them manually?

Also, I'm not 100% sure if we want to autoload all the files under `lib/` because from what I see [in the current entry point ](https://github.com/rails/mrsk/blob/35b42cc8859bd94eeadc985a68c96d3f2dc41090/lib/mrsk.rb#L4-L5) we don't require the `lib/cli` files so adding `Zeitwerk::Loader.for_gem` would change the existing behavior.

robzolkos: Do you think we should just get Zeitwerk to autoload when running tests and leave it out of the lib code altogether?  As mentioned in the docs [here](https://github.com/fxn/zeitwerk#testing-compliance)

robzolkos:
I'm not added Zeitwerk to a gem like this before so could use some pointers on getting this right 🙏🏻

fxn:
Hey!

No, this is not a good idea, let me explain.

If the gem loaded with Zeitwerk, then it would make sense to let CI verify compliance. This is just a safety net, if a file in CI fails the validation, it means the file has no test coverage. If the file is exercised by the test suite and does not abide to the convention, then you'll find out in the test itself already.

However, this gem does not load with Zeitwerk, therefore it does not make sense to check it _could_ be loaded by Zeitwerk.

fxn:
That said, on a quick check, this gem seems to be structured conventionally. Loading with Zeitwerk seems to be just a couple lines of code and deleting all requires. This is in the context of modern Rails, so would be coherent too. @dhh I could write the patch myself if you liked the idea.

dhh:
Would be happy to see the explicit requires disappear! Thanks @fxn ✌️

robzolkos:
Closing pending patch from Xavier.

PR #35, opened 2023-02-03
Load with Zeitwerk

fxn:
This patch lets MRSK load with Zeitwerk:

* The project was in excellent shape re conventions.
* There are about 60 `require` calls deleted in code and tests.
* The loader eager loads the project because Thor commands need to be in memory. I inspected `$LOADED_FEATURES` in `main` and the existing `require` calls were basically doing the same thing in cascade anyway.
* `bin/mrsk` now loads the entrypoint instead of `mrsk/cli` to have the loader in place. With that, the subsequent reference to `Mrsk::Cli::Main` just works. Note that `lib/mrsk/cli.rb` was loading the entrypoint, so in a sense this setup simplifies and orders bootstrapping in a natural way.
* Similarly, `test/test_helper.rb` loads the entrypoint as well.
* The file `lib/mrsk/sshkit_with_ext.rb` is special and this is fine, we just tell the loader to ignore it. This extension is loaded with an ordinary `require` where is needed (same original spot, not edited).
* There's some extra diff lines due to whitespace cleanup (automatic in my editor).

References #34.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,6 +6,7 @@ PATH
       dotenv (~> 2.8)
       sshkit (~> 1.21)
       thor (~> 1.2)
+      zeitwerk (~> 2.5)
 
 GEM
   remote: https://rubygems.org/
@@ -91,6 +92,7 @@ PLATFORMS
   arm64-darwin-22
   x86_64-darwin-20
   x86_64-darwin-21
+  x86_64-darwin-22
   x86_64-linux
 
 DEPENDENCIES
--- a/bin/mrsk
+++ b/bin/mrsk
@@ -4,7 +4,7 @@
 Thread.report_on_exception = false
 
 require "dotenv/load"
-require "mrsk/cli"
+require "mrsk"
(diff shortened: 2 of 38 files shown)


fxn:
There are some failures, but they seem to come from an updated `main` branch.

dhh:
Lovely work, @fxn! Shows just what a treat Zeitwerk is to all Ruby libraries 🤘❤️

fxn:
Awesome @dhh!

Let me add that this patch eager loading the gem is the KISS version, provided the `require`s were already eager loading in practice.

However, if the day arrives that you feel like you'd like to eager load less (even if that day is today :)), there is also the option of bump the version constraint for Zeitwerk, and eager load only the commands using [namespace eager loading](https://github.com/fxn/zeitwerk#eager-load-namespaces), for example. The rest of the gem would be loaded on demand, and in particular when running an individual test, only what it is used (plus commands).

Just drop me a line in that case, it's an easy patch.

PR #45, opened 2023-02-10
Set .env files to 0600 so they're not world-readable

anoldguy:
If we don't want to keep the 'generate credentials' step inside the mrsk execution paths, let's be careful with the credentials we do pull down, and not world-readable.

I'm intentionally touching the file first and setting permissions, so that credentials are never in a world-readable file, but the actual chance of that happening is extremely minimal, in which case, the `File.chown` call could be moved after the `File.write`.

### DIFF
--- a/lib/mrsk/cli/main.rb
+++ b/lib/mrsk/cli/main.rb
@@ -107,13 +107,9 @@ class Mrsk::Cli::Main < Mrsk::Cli::Base
   desc "envify", "Create .env by evaluating .env.erb (or .env.staging.erb -> .env.staging when using -d staging)"
   def envify
     if destination = options[:destination]
-      File.write(".env.#{destination}", "")
-      File.chown(0600, ".env.#{destination}")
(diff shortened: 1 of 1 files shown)


lib/mrsk/cli/main.rb
```diff
@@ -107,8 +107,12 @@ def init
   desc "envify", "Create .env by evaluating .env.erb (or .env.staging.erb -> .env.staging when using -d staging)"
   def envify
     if destination = options[:destination]
+      File.write(".env.#{destination}", "")
+      File.chown(0600, ".env.#{destination}")
       File.write(".env.#{destination}", ERB.new(IO.read(Pathname.new(File.expand_path(".env.#{destination}.erb")))).result)
```
jeremy, line 112: Can pass `:perm` to set permissions if a file is created:
```suggestion
      File.write(".env.#{destination}", ERB.new(IO.read(Pathname.new(File.expand_path(".env.#{destination}.erb")))).result, perm: 0600)
```

dhh:
Ah, just saw this after I already took a swing.

PR #53, opened 2023-02-20
fix(escape-cli-args): Always use quotes to escape CLI arguments

pagbrl:
## What's the identified issue ?

Mrsk currently does not escape the command line arguments it concatenates for the docker commands.

This make mrsk very susceptible to unescaped characters errors, as the following example demonstrates with an environment string that contains the symbol `&` : 
```
env:
  clear:
    - THIS_WILL_FAIL=fdjksl&fejkl
```

Which subsequently creates this docker run issue : 
```
  INFO [a427b772] Running docker run -d --name healthcheck-hello-symfo -p 3999:80 --label service=healthcheck-hello-symfo THIS_WILL_FAIL=fdjksl&fejkl redacted:latest on redacted
  ERROR (SSHKit::Command::Failed): docker exit status: 127
docker stderr: bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
bash: line 1: -e: command not found
bash: line 1: fejkl: command not found
"docker run" requires at least 1 argument.
See 'docker run --help'.

Usage:  docker run [OPTIONS] IMAGE [COMMAND] [ARG...]

Run a command in a new container
```

This is due to the `argumentize` function not escaping the characters.

## What fix does this PR introduce ?
This PR forces all `argumentized` CLI arguments to be escaped, thus making the produced commands safe to run.

## What's important to note
- Labels are argumentized the same way that environment variables are, which means I also had to change the way we use traefik labels that use backticks in the code. Result should be 1:1 with what existed before, but I also had to revamp most of the tests to account for this change. 
We might have wanted to make two different `argumentize` functions, but I don't love it because of code duplication and because I think it's always better to escape characters.

- In the future we might want to create a special case for integers, but iirc everything that bash passes is a string so I didn't do that here and noticed no behavior

Tests are passing and this has been tested on a mrsk deploy.

### DIFF
--- a/lib/mrsk/utils.rb
+++ b/lib/mrsk/utils.rb
@@ -1,13 +1,14 @@
 module Mrsk::Utils
   extend self
 
-  # Return a list of shell arguments using the same named argument against the passed attributes (hash or array).
+  # Return a list of escaped shell arguments using the same named argument against the passed attributes (hash or array).
   def argumentize(argument, attributes, redacted: false)
-    Array(attributes).flat_map do |k, v|
-      if v.present?
-        [ argument, redacted ? redact("#{k}=#{escape_bash_string v.to_s}") : "#{k}=#{escape_bash_string v.to_s}" ]
+    Array(attributes).flat_map do |key, value|
+      if value.present?
+        escaped_pair = [ key, value.to_s.dump.gsub(/`/, '\\\\`') ].join("=")
+        [ argument, redacted ? redact(escaped_pair) : escaped_pair ]
       else
-        [ argument, k ]
+        [ argument, key ]
       end
     end
   end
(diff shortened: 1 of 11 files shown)


lib/mrsk/utils.rb
```diff
@@ -5,7 +5,7 @@ module Mrsk::Utils
   def argumentize(argument, attributes, redacted: false)
     Array(attributes).flat_map do |k, v|
       if v.present?
-        [ argument, redacted ? redact("#{k}=#{v}") : "#{k}=#{v}" ]
+        [ argument, redacted ? redact("#{k}=\"#{v}\"") : "#{k}=\"#{v}\"" ]
```
dhh, line 8: Don't we need to escape any potential "'s inside the v?

pagbrl: Yes thank you ! I've used the `dump` function which does exactly that, and added a gsub escape for backticks which are not escaped by this function.

PR #54, opened 2023-02-21
Print container logs when HealthCheck response_code != 200

intrip:
The Healthcheck container is shut down right after performing the check, this makes it harder to troubleshoot configuration issues in the healthcheck endpoint, e.g [DNS rebinding error](https://edgeapi.rubyonrails.org/classes/ActionDispatch/HostAuthorization.html). Print the container logs to help with the troubleshooting.

### DIFF
--- a/lib/mrsk/cli/healthcheck.rb
+++ b/lib/mrsk/cli/healthcheck.rb
@@ -1,6 +1,8 @@
 class Mrsk::Cli::Healthcheck < Mrsk::Cli::Base
   MAX_ATTEMPTS = 5
 
+  class HealthcheckError < StandardError; end
+
   default_command :perform
 
   desc "perform", "Health check current app version"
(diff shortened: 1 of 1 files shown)


lib/mrsk/cli/healthcheck.rb
```diff
@@ -18,6 +18,7 @@ def perform
           if status == "200"
             info "#{target} succeeded with 200 OK!"
           else
+            error capture_with_info(*MRSK.healthcheck.logs)
```
dhh, line 21: We already print the logs on the other rescue. Better to change the raise to be a specific, but unique exception, then catch that in the clause that prints the logs as well. Can add a dedicated HealthCheckError exception.

intrip: Good point 👍. https://github.com/mrsked/mrsk/pull/54/commits/689b81014bb5914c9f45f14e18ed0c4731bee6a5

PR #55, opened 2023-02-21
Change default healthcheck port to 3000

matharvard:
The documentation states that the default healthcheck port is 3000, but mrsk seems to be hitting 3999. This PR changes mrsk's default to match the documentation again!

~~Noticed the healthcheck tests are failing and I'm looking into it now :(~~ Fixed.

In retrospect I probably should have created an issue first and worked off that. If you'd prefer that just let me know! I'm actively using mrsk on soon-to-be production environments, so I'm very motivated to help out where I can!

### DIFF
--- a/lib/mrsk/commands/healthcheck.rb
+++ b/lib/mrsk/commands/healthcheck.rb
@@ -1,5 +1,5 @@
 class Mrsk::Commands::Healthcheck < Mrsk::Commands::Base
-  EXPOSED_PORT = 3999
+  EXPOSED_PORT = 3000
 
   def run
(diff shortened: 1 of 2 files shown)


matharvard:
I think I misunderstood what was going on here and why my application failed to deploy. From looking at the deployment output I saw the `curl` against port 3999 and figured that was the problem. My understanding now is that mrsk starts a `healthcheck-service` container to perform the `curl` command against on port 3999. The Rails application I am working with doesn't seem to boot up fast enough for this check to complete successfully.

PR #57, opened 2023-02-23
Example on how to set up Cron

intrip:


README.md
```diff
@@ -330,6 +330,21 @@ accessories:
...
+
+You can use a specific container to run your Cron jobs:
+
+```yaml
+servers:
+  cron:
+    hosts:
+      - 192.168.0.1
+    cmd:
+      bash -c "cat config/crontab | crontab - && cron -f"
```
glaucocustodio, line 343: are you running you cron container as root? Because there seems to be some massage to make this command work as non-root user, I couldn't figure out..

glaucocustodio: I am also struggling to make logs of cron jobs available to `docker logs`, I tried a couple of things (like [this one](https://betterstack.com/community/questions/how-to-run-cron-jobs-inside-docker-container/) for example) but none seems to work

glaucocustodio: I ended up running as root, you can check my write-up on this [blog post](https://glaucocustodio.github.io/2024/01/23/migrating-from-dokku-to-kamal-scheduling-cron-jobs)

corvuszero: What other people experiencing this issue don't mention is the fact that they are most likely running using the root user. I tried absolutely everything until I followed your advice and used root. Crontab started to magically work, thanks for your blog post.

PR #67, opened 2023-03-02
Add devcontainer configuration

htunnicliff:
This PR adds a configuration for a [devcontainer](https://containers.dev). I added [Ubuntu](https://github.com/devcontainers/images/tree/main/src/base-ubuntu) as the base image and included the [Ruby](https://github.com/devcontainers/features/tree/main/src/ruby) feature set. 

Other base images—such as Debian or Alpine—could work as well, but Ubuntu seemed like a good one to start with.

--- 

For background, a devcontainer configuration helps create a container-based development environment that includes all the tools needed to work in a given repository, without the need to modify a user's regular computing environment.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,23 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/ruby
+{
+  // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
+
+  // Features to add to the dev container. More info: https://containers.dev/features.
+  "features": {
+    "ghcr.io/devcontainers/features/ruby:1": {}
+  }
+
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate the effort, but MRSK doesn't have any convoluted development requirements. If "has Ruby on system" is too high a bar for someone, I think MRSK is probably not a good first project for them to engage with.

htunnicliff:
> Appreciate the effort, but MRSK doesn't have any convoluted development requirements. If "has Ruby on system" is too high a bar for someone, I think MRSK is probably not a good first project for them to engage with.

Based on your initial introduction to MRSK (simplicity!) this makes a lot of sense. I feel quite similar about adding unnecessary complexity, especially in a simple project.

That being said, quite a few projects on GitHub are adding in devcontainer configurations to make it easy for engineers to work on projects however they like, be it locally using a traditional IDE, in a cloud environment like GitHub Codespaces, CodeSandbox, or GitPod, or even on an iPad.

To play devil's advocate even further, it appears that rails/rails has had a devcontainer configuration for over a year: https://github.com/rails/rails/commits/main/.devcontainer.

PR #71, opened 2023-03-03
Destination aware container names

tbuehlmann:
I started looking into having role aware container names but that turned out to be quite a big undertaking. So I started with destinations instead to see where it could head.

This PR is a start to have destination aware container names so **different destinations** of the **same service** can run on the **same server**.

This doesn't work right now:

```yml
# config/deploy.yml
service: app

# config/deploy.staging.yml
web:
  - 1.1.1.1

# config/deploy.demo.yml
web:
  - 1.1.1.1
```

… as both docker containers would have the same name `app-<version>` on 1.1.1.1. The PR adds - if present - the destination as part of the container name, ending up as `app-staging-<version>` and `app-demo-<version>`.

I started with `Mrsk::Commands::App` and before changing all of the other commands: Is this the way we want it to be? It'll include some if conditions on whether or not `MRSK.destination` (or `config.destination`) is truthy or not.

Another approach could be to default the destination to `"production"` if it's missing. This would probably be simpler but is more chatty in regards to the container names (if no destination is used it would be `app-production-<versions>` instead of `app-<version>`).

### DIFF
--- a/lib/mrsk/commands/auditor.rb
+++ b/lib/mrsk/commands/auditor.rb
@@ -21,7 +21,11 @@ class Mrsk::Commands::Auditor < Mrsk::Commands::Base
 
   private
     def audit_log_file
-      "mrsk-#{config.service}-audit.log"
+      if config.destination
+        "mrsk-#{config.service}-#{config.destination}-audit.log"
+      else
+        "mrsk-#{config.service}-audit.log"
+      end
     end
 
     def tagged_record_line(line)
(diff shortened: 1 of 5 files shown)


lib/mrsk/commands/auditor.rb
```diff
@@ -21,7 +21,11 @@ def reveal
 
   private
     def audit_log_file
-      "mrsk-#{config.service}-audit.log"
+      if config.destination
+        "mrsk-#{config.service}-#{config.destination}-audit.log"
+      else
+        "mrsk-#{config.service}-audit.log"
+      end
```
dhh, line 28: Can probably use the `compact.join("-")` approach here too.

dhh:
This is great. I'm not sure it's right for accessories to take destinations into considerations. You really want to be explicit with those, since it's usually about stable data.

PR #77, opened 2023-03-04
Allow setting the Dockerfile and the Docker build context when building

calmyournerves:
Hey there 👋 

We're using a monorepo and gems get included in the apps via path (e.g. `gem 'mygem', path: '../mygem'`). As Docker limits everything to the context (default `.`), including gems outside of the context does not work. I'm no Docker expert, but setting the context to `..` and specifying the Dockerfile with `--file Dockerfile` seems to do the trick. 

This PR:
- adds a `context` option to the builder config
- adds a `dockerfile` option to the builder config
  The default value for this is `Dockerfile` and it will always be appended to the build command. This is helpful if someone wants to use different Dockerfiles (e.g. `Dockerfile.development`). I'm not sure if this is feasible as a config option at all, but to support passing the context, the `--file` argument seems to be needed in the build command (and it mayber is a little more explicit). Otherwise, Docker would look for the Dockerfile in the root of the context.

 The build configuration would look as follows:
 
```yaml
# Set context
builder:
  context: ".."

# Use different Dockerfile
builder:
  dockerfile: Dockerfile.xyz

# Set context and Dockerfile
builder:
  context: ".."
  dockerfile: "../Dockerfile.xyz" # Dockerfile is in the parent directory
```

It would also be possible to just allow configuring the `context` and making sure the `--file Dockerfile` argument always gets passed to the build command (without it being a configuration option).

### DIFF
--- a/lib/mrsk/commands/builder/base.rb
+++ b/lib/mrsk/commands/builder/base.rb
@@ -10,7 +10,8 @@ class Mrsk::Commands::Builder::Base < Mrsk::Commands::Base
   end
 
   def build_options
-    [ *build_tags, *build_labels, *build_args, *build_secrets ]
+    [ *build_tags, *build_labels, *build_args, *build_secrets, *build_dockerfile ]
+  end
   end
 
   private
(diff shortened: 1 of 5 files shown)


dhh:
Good stuff. Can you add docs too?

calmyournerves:
Thanks for merging! Sure, will open another PR 👍

PR #85, opened 2023-03-05
Customizable Traefik host port

clowder:
Allow users to free up port 80 on the host machine, without losing Traefik's Docker routing super-powers. 

My usecase is using MRSK to deploy a Rails app to a server already running Caddy on port 80.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,15 @@ traefik:
 
 This will start the traefik container with `--accesslog=true accesslog.format=json`.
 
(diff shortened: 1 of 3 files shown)


tbuehlmann:
Would be nice to also be able to publish :8080 for the insecure api dashboard.

rainerborene:
@clowder are you using traefik behind caddy? need to deploy caddy as well but don't want to serve our traffic through traefik.

Issue #62, opened 2023-03-01
Release the CLI as downloadable binary


lambrospetrou:
It would be nice to be able to download and use the tool as a CLI without depending on Ruby being on the system (requiring `gem install`).

Any plans on releasing this as a standalone CLI tool?

dhh:
Please Do Investigate!

99linesofcode:
I just stumbled upon mrsk so might not be able to see the bigger picture here (ie. making some incorrect assumptions) but what about containerizing mrsk itself? You might even be able to publish it as a Docker extension, integrating it into Docker Desktop. Assuming Docker is a core dependency.

dhh:
Yes, definitely possible to do that. Same strategy we used with Docked Rails. Happy to see that explored.

kjellberg:
By adding the Dockerized app to [Homebrew](https://brew.sh/), it can be installed with ease on OSX and Linux (and possibly WSL). A global `mrsk` alias is created, allowing easy access to the `mrsk` command:

```console
brew install mrsk
mrsk --version
```

latest version, please..
```console
brew upgrade mrsk 
```

Brew formula could look something like this:

```ruby
class MrskApp < Formula
  desc "Mrsk"
  homepage "https://example.com/mrsk"
  url "https://example.com/mrsk.tar.gz"
  version "1.0.0"

  depends_on "docker"

  def install
    system "docker", "pull", "mrsk:latest"
    system "docker", "tag", "mrsk:latest", "mrsk:#{version}"
    system "docker", "save", "-o", "mrsk-#{version}.tar", "mrsk:#{version}"
    bin.install "mrsk"
  end

  test do
    system "#{bin}/mrsk", "--version"
  end
end
```

dhh:
That sounds great! Please make it so.

kjellberg:
PR #73 creates a Dockerfile, but we need to build, tag and upload the Docker image somewhere on every release.

Here's what I propose we do:

1. Add a new GitHub Actions workflow to the repository that triggers on each release tag.
2. In the workflow, build a Docker image of our application and tag it with the release version number.
3. Push the Docker image to a public Docker registry, such as Docker Hub or GitHub Packages.

Or just some extra rows in `./bin/release`. It's up to you @dhh.

dhh:
I've added the GH action to trigger builds on pushes to main. Right now it just pushes latest. It lives in GH Packages 👍.

[...]

PR #86, opened 2023-03-05
Installation script for mrsk in docker

kjellberg:
This PR introduces a new installation script that simplifies the installation of `mrsk`. The script downloads and installs the latest Docker image, creates a ~/.mrsk/bin directory, adds it to the `PATH` environment variable, and creates a `mrsk` binary file that runs the Docker image with the appropriate flags and arguments. The script can be used on any platform that supports Docker, without requiring Ruby to be installed.

## Motivation
The main motivation behind this installation script is to provide a simple and easy way to install mrsk with as few dependencies as possible. By using a Docker image, we can eliminate the need for installing Ruby and any other dependencies that may be required. This makes the installation process simpler and more streamlined, and ensures that `mrsk` can be installed and used on any platform that supports Docker. #62 

## Test
To test this new installation script, run the following command in your terminal:

```console
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/kjellberg/mrsk/docker-install-script/scripts/install.sh)"
```

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+set -e
+
+# Define functions for printing messages to the user
+function print_msg() {
+  echo -e "\n\033[1m\033[34m${1}\033[0m\n"
+}
+
+function print_error() {
+  echo -e "\n\033[1m\033[31mError: ${1}\033[0m\n"
+  exit 1
+}
+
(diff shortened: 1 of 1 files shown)


kjellberg:
I'll make a PR to `mrsked/mrsk-site` instead. `https://mrsk.dev/install.sh` is a much cooler URL.

PR #88, opened 2023-03-05
Ensure curl is installed on hosts during bootstrapping

simonrand:
Debian minimal (and others) doesn't come with `curl` installed out of the box, MRSK should check and resolve this during bootstrapping.

This is the root cause of the issue - but possibly worth opening an issue for the lack of information provided during failing healthchecks due to things like curl missing – i.e. no obvious error is raised beyond `INFO Health check against /up failed to respond, retrying in Ns...` – happy to open an issue for this if people would like.


dhh:
Please do explore better communicating errors on the healthcheck as well 👍

Issue #98, opened 2023-03-10
Multiple traefik ports


stepbeekio:
Hey,

I'd love to have a way to configure traefik to listen on multiple ports for multiple entrypoints to enable services that might listen for other kinds of traffic - e.g. a port open for HTTP traffic and another for TCP traffic. My current thinking is that this would be an escape hatch from the default configuration so mrsk doesn't need to support this as packaged up feature.

Would something like this:

```yaml
traefik:
  host_port: 8080
  additional_ports:
     - 9000
  entrypoints.mytcp.address: ':9000'
```

be of interest? 

Alternatively, maybe it  could look like this:

```yaml
traefik:
  host_port: 8080
  additional_entrypoints:
    myentrypointname: ':9000'
```

producing a traefik arg like `--entrypoints.myentrypointname.address=:9000`. 

This would also enable the traefik admin dashboard for interested users.

stepbeekio:
I've done a little work on this locally. Using the additional_ports suggestion above, I've ended up with a config yaml that includes a lot of excess effort. I've omitted anything not strictly related.

```yaml
service: logline

labels:
  traefik.tcp.routers.logstash.rule: 'HostSNI(`*`)'
  traefik.tcp.routers.logstash.entrypoints: logstash
  traefik.tcp.services.logstash.loadbalancer.server.port: 10516
  traefik.http.routers.logline.entrypoints: web
  traefik.http.services.logline.loadbalancer.server.port: 8080

traefik:
  additional_ports:
    - 10516
  args:
    'entrypoints.web.address=:80': true
    'entrypoints.logstash.address=:10516': true
    accesslog: true
```

I have a PR ready to go if this is an acceptable UX for now - or am interested in hearing about how 37s feels about the tradeoff between deeper traefik integration vs an escape hatch like this @dhh.

I'm particularly interested in this use-case as a motivating feature for us to move from platforms like Heroku et al where the ability to open a plain ol' TCP port isn't offered. The gap between PaaS and managed K8S or Nomad etc. is pretty sparse so I'd love to have mrsk fill this for us.

stepbeekio:
Opened a PR.

stepbeekio:
Updated PR in response to feedback from @kjellberg. This is the final result:

```yaml

labels:
  traefik.tcp.routers.logstash.rule: 'HostSNI(`*`)'
  traefik.tcp.routers.logstash.entrypoints: logstash
[...]

PR #100, opened 2023-03-10
Added the additional_ports configuration for traefik

stepbeekio:
See #98 for more detailed proposal. This change adds the `additional_ports` configuration for traefik. Includes documentation to give entrypoint examples for this escape hatch. The ruby code impacted is relatively small - we essentially just allow a user to add other published ports to the traefik container.

The obvious downside to this is that by using this configuration in isolation a user could bring down their service. Traefik wouldn't know which port to bind as the default entrypoint and would fail to serve traffic.

Added a unit test and tested locally against a running deploy.


kjellberg:
This can also be used to enable port 443 for https. Just a thought, why not rename `additional_ports` to just `ports`, which defaults to [80]. People messing with ports should be smart enough to include port 80 if they configure it.

```yaml
traefik:
  ports:
    - 80
    - 443
    - 9000
```

stepbeekio:
@kjellberg That's a good point. I think it would fit better with moving away from traefik defaults too. I'll get time later today to edit.

kjellberg:
Actually I need this right now. Was trying to enable https for a website but I had to patch mrsk to add port 443, just 5 seconds before you created this PR. Cloudflare messes up with session cookies, using flexible ssl, so would be nice to support additional ports.

If this get merged, I'd be happy to create a patch that enables ssl with as simple as:

```yaml
servers:
  web:
    hostname: example.com
    hosts:
      - 206.189.112.50
traefik:
  ports:
    - 80
    - 443
  letsencrypt:
     email: me@acme.com
```

**Another thought**:

What if two different projects deploys to the same server. How do we keep the traefik config consistent? Should `mrsk` support this?

##### Project A
```yaml
traefik:
  ports:
    - 80
    - 9000
    - 9001
```

##### Project B
```yaml
traefik:
  ports:
    - 80
    - 8080
```

Deploying *Project B* will override ports config from *Project A*. I guess we already have the same problem with `args`.

lib/mrsk/commands/traefik.rb
```diff
@@ -54,6 +54,22 @@ def port
...
+      end
+    end
+
+    def volumes
+      if volumes = config.raw_config.dig(:traefik, "options", "volumes")
+        volumes.collect { |value| "--volume #{value}" }.compact
+      else
+        ["--volume /var/run/docker.sock:/var/run/docker.sock"]
+      end
+    end
```
dhh, line 71: I was looking to have all this be generic. That there's nothing specific about publish or volumes. That it should all just be options that turn into this. Like we do for app.

stepbeekio:
@dhh I've edited the `optionize` method to allow for nested lists inside arguments that get expanded. I figured that making `optionize` more flexible in what it accepted would be better than introducing a one-off for traefik options. I'm not a ruby dev so have no idea if there's a smarter way to do this.

nerdinand:
I ran into problems with this feature with the ActiveSupport main-branch. There, `docker_option_args` returns `[]`, because `config.raw_config.dig(:traefik, "options")` returns `nil` for some reason.

PR #101, opened 2023-03-10
fix: Github Workflow: Docker Publish

davegudge:
The workflow was failing with:

```
The workflow is not valid. .github/workflows/docker-publish.yml (Line: 22, Col: 14): Unexpected symbol: '|'. Located at position 12 within expression: github.ref | replace('refs/tags/', '')
```

The `set-output` command is deprecated, so the issue has been fixed by utilising the `github.ref_name` context to retrieve the version tag that triggered the workflow.

> `github.ref_name`: The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub. For example, `feature-branch-1`.

https://docs.github.com/en/actions/learn-github-actions/contexts

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -17,10 +17,6 @@ jobs:
         name: Checkout
         uses: actions/checkout@v3
       -
-        name: Extract Version Number
-        id: extract_version
-        run: echo "::set-output name=version::${{ github.ref | replace('refs/tags/', '') }}"
-      -
         name: Set up QEMU
         uses: docker/setup-qemu-action@v2
       -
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -16,10 +16,6 @@ jobs:
       -
         name: Checkout
         uses: actions/checkout@v3
```
davegudge, line 18: According to the [`docker/build-push-action` docs](https://github.com/docker/build-push-action#git-context), the "Checkout" step can also be removed from this workflow:

> By default, this action [`docker/build-push-action`] uses the [Git context](https://docs.docker.com/engine/reference/commandline/build/#git-repositories), so you don't need to use the [actions/checkout](https://github.com/actions/checkout/) action to check out the repository as this will be done directly by [BuildKit](https://github.com/moby/buildkit).

```suggestion

```

PR #103, opened 2023-03-10
Install buildx inside container

99linesofcode:
Title says it all really. With this change in place I was able to run `mrsk setup` and succesfully pull in my image on the server. Here is the docker run command I used to get that working:

`docker run -it --rm -v "${PWD}:/workdir" -v "${SSH_AUTH_SOCK}:/ssh-agent" -v /var/run/docker.sock:/var/run/docker.sock -e "SSH_AUTH_SOCK=/ssh-agent" ghcr.io/99linesofcode/mrsk:latest setup`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,9 @@
 # Use the official Ruby 3.2.0 Alpine image as the base image
 FROM ruby:3.2.0-alpine
 
+# Install  docker/buildx-bin
+COPY --from=docker/buildx-bin /buildx /usr/libexec/docker/cli-plugins/docker-buildx
+
 # Set the working directory to /mrsk
 WORKDIR /mrsk
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ env:
     - RAILS_MASTER_KEY
 ```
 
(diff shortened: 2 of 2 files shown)


99linesofcode:
Ok so I actually intended to push this to my fork while trying to get mrsk working in the context of a Laravel application using the newly added mrsk package..

So far the build step failed for me but I might run into more issues along the way. Let me do some more testing here before we merge this.

99linesofcode:
image

> A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and hyphens. A tag name may not start with a period or a hyphen and may contain a maximum of 128 characters.

99linesofcode:
As you can see from the docker run command in the description, I also needed to volume mount the docker UNIX socket into the container or else MRSK would not get past the push stage. I feel we should explain why mounting the socket is necessary as well but frankly, I'm not entirely sure why Docker needs this. Anyone that could put that into words for me?

99linesofcode:
Apparently, mounting the docker socket into the container is risky as it gives the container access to the host machine. If I better understood why I had to mount it in the first place, I might be able to figure out how we can do without.

PR #111, opened 2023-03-12
Add option `--use-prebuilt-image` to skip building and pushing the image

calmyournerves:
Hey 👋 

MRSK always builds the image from the local tree when deploying. Ideally, we want to build our images on CI and then deploy them locally with passing the image tag. This would help that we only deploy images/commits that have passed CI before. Also, if we forget to pull the latest changes or have uncommitted changes, these would be deployed (which can be dangerous). 

This PR adds an option `--use-prebuilt-image` to skip building the image and just deploy a given version (MRSK already supports passing `--version`).

This is more of a proposal (haven't looked at tests yet). I added it as a global option for now but also looked at adding it as a individual options for the commands that need it (`deploy`, `redeploy`, `build deliver`). I don't like it too much, but it seems easier than checking and passing options around.

What are your opinions about this? How do you handle building/CI/deployment?

### DIFF
--- a/lib/mrsk/cli/base.rb
+++ b/lib/mrsk/cli/base.rb
@@ -12,6 +12,7 @@ module Mrsk::Cli
     class_option :quiet, type: :boolean, aliases: "-q", desc: "Minimal logging"
 
     class_option :version, desc: "Run commands against a specific app version"
+    class_option :use_prebuilt_image, type: :boolean, default: false, desc: "Use prebuilt image, skip building"
 
     class_option :primary, type: :boolean, aliases: "-p", desc: "Run commands only on primary host instead of all"
     class_option :hosts, aliases: "-h", desc: "Run commands on these hosts instead of all (separate by comma)"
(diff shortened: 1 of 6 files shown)


dhh:
I think we just need it to be a CLI option to deploy/redeploy. Then it can use that to fork on whether to call deliver or just pull. Let's go with "--skip-push/-P".

calmyournerves:
Hey @dhh, thanks for the feedback! Will update the PR accordingly. 

Curious on how you folks handle this when deploying HEY. Is the image just always built/deployed on the developers machine using `mrsk deploy`? Or do you have some kind of shared deployment service?

dhh:
We just build locally and push directly at 37signals. Not using an external CI server. But happy to support MRSK supporting the latter.

calmyournerves:
Yep, looking into it! Seems a little tricky as `SSHKit.config.backend = SSHKit::Backend::Printer` does only seem to work for remote commands, not for those using `run_locally` (those will just get executed). It tried this with monkey-patching SSHKit which works, but it just feels wrong. Am I missing something?

dhh:
Ah, interesting. I think monkey-patching is fine in this case. Can always try to upstream afterwards. Or you can just mock what's being executed locally.

calmyournerves:
As `deploy`/`redeploy` just invoke other commands, I specifically added tests to check those invocations. I'll look into testing more commands (also testing commands run by `run_locally` with the mentioned SSHKit limitation) in the scope of another PR.

What do you think?

PR #117, opened 2023-03-14
Add tests for main CLI commands

calmyournerves:
This adds tests for all CLI commands in `cli/main.rb`. 

It also includes a monkey-patch for `SSHKit` to make `run_locally` use the printer backend too as `SSHKit.config.backend` only applies to remote commands. The backend for `run_locally` is [hard coded](https://github.com/capistrano/sshkit/blob/master/lib/sshkit/dsl.rb#L9).

I'm happy to open additional PRs to test more of the (sub)commands. I just want to make sure this is the way to go.


dhh:
Please do fill in all the missing tests 🙏

PR #118, opened 2023-03-14
Add docker in docker to Dockerfile for container dev

rstarmer:
The Dockerfile for MRSK is fine in it's own way, but it doesn't really give you a docker environment that can actually run mrsk!  A Dockerfile is added by this pull request to specifically to support the local docker build commands against the local docker daemon.

The README.md is updated to include instructions for launching via Docker (for those without a local Ruby enviornment), and an alias is provided to share .ssh and the docker socket into the container.

A tweak to the readme is included for those who don't login to remote servers as root to point out the need for this when supporting the "install docker" aspect of a new server configuration.


rstarmer:
I noted that in the latest main branch, that docker build was added to the Dockerfile, which is a better approach than mine. I've removed the un-necessary Dockerfile.dind, and updated the README to include a cleaned up build instruction.

dhh:
We should reference the container distributed here via ghcr.io/mrsked/cli rather than having to build your own.

rstarmer:
Updated to point to `ghcr.io/mrsked/mrsk` which I presume is the git target proposed (there does not seem to be a `ghcr.io/mrsked/cli` image.

PR #127, opened 2023-03-17
Add hint to use Docker Hub access tokens

ediathome:
Adding a hint to use Docker Hub access tokens instead of the Docker Hub user account password.

I specified my Docker Hub user account password in the .env file and mrsk failed with the following message:

    ERROR (SSHKit::Command::Failed): docker exit status: 256
    docker stderr: WARNING! Using --password via the CLI is insecure. Use --password-stdin.
    Error response from daemon: Get "https://registry-1.docker.io/v2/": unauthorized: incorrect username or password

I resolved the issue by creating a Docker Hub access token and specifying the access token in the .env file:

   MRSK_REGISTRY_PASSWORD=my-docker-hub-access-token

Adding the comment might save any newbies like me to run into the same error.

### DIFF
--- a/lib/mrsk/cli/templates/deploy.yml
+++ b/lib/mrsk/cli/templates/deploy.yml
@@ -13,6 +13,8 @@ registry:
   # Specify the registry server, if you're not using Docker Hub
   # server: registry.digitalocean.com / ghcr.io / ...
   username: my-user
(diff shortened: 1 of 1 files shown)


dhh:
Added as a general point in f73c526890430f1653e79a973119789bcb2f94d7 👍. Applies to all registries that use tokens.

PR #131, opened 2023-03-19
Global logging configuration

calmyournerves:
Docker supports a variety of logging drivers, such as `json-file`, `awslogs`, and `fluentd`. Some of them, like `awslogs` or `fluentd`,  can push logs directly to a external service. In our case, we would like to be able to send all container logs directly to AWS CloudWatch with `awslogs` driver. After configuring an instance profile on AWS with the necessary permissions, it is as simple as passing `--log-driver awslogs --log-opt awslogs-region=<region> --log-opt awslogs-group=<cloudwatch log group>` to the `docker run` command.

As MRSK currently always passes `--log-opt max-size=MAX_LOG_SIZE` to `docker run` commands, defining the logging options mentioned above in the `options` of the role does not work (MRSK supports passing options to the `docker run` command via role `options:` in the config file) with the following error: `docker stderr: docker: Error response from daemon: unknown log opt 'max-size' for awslogs log driver.`

I thought about just removing the default `max-size` log option which would make passing `options` via role work. But being able to control logging behaviour globally might be even better, as we can push logs of Traefik and accessories to the desired backend as well.

This PR adds a global configuration block called `logging:` that passes the configured log options to all running containers.

Example:

```yaml
logging:
  driver: "awslogs"
  options:
    awslogs-region: "eu-central-2"
    awslogs-group: "my-app"
    # On CloudWatch, this results in a log stream per container with the naming of `<image-name>-<container-id>`:
    tag: '{{ with split .ImageName ":" }}{{join . "_"}}{{end}}-{{.ID}}'
```

~This PR would also change the currently defined default log size of 10M (the Docker default is `100M`). To get this behaviour back, the following config would need to be used:~

If nothing is configured, the current default of `max-size=10m` is used.


calmyournerves:
Updated it to always use `max-size=10m` as default like before.

PR #143, opened 2023-03-22
Default to deploying the config version

djmb:
If we don't supply a version when deploying we'll use the result of `docker image ls` to decide which image to boot. But that doesn't necessarily correspond to the one we have just built.

E.g. if you do something like:

```
mrsk deploy        # deploys git sha AAAAAAAAAAAAAA
git commit --amend # update the commit message
mrsk deploy        # deploys git sha BBBBBBBBBBBBBB
```

In this case running `docker image ls` will give you the same image twice (because the contents are identical) with tags for both SHAs but the image we have just built will not be returned first. Maybe the order is random, but it always seems to come second as far as I have seen.

i.e you'll get something like:

```
REPOSITORY    TAG              IMAGE ID       CREATED          SIZE
foo/bar       AAAAAAAAAAAAAA   6272349a9619   31 minutes ago   791MB
foo/bar       BBBBBBBBBBBBBB   6272349a9619   31 minutes ago   791MB
```

Since we already know what version we want to deploy from the config, let's just pass that through.


dhh:
Couldn't we just get rid of the whole `docker image ls` stuff together too, then? If we're always passing a version?

djmb:
If you do `mrsk app boot` or `mrsk app exec` I think you would want the most recently deployed version, not what you have locally.

With https://github.com/mrsked/mrsk/pull/143/commits/1ed4a37da2b73125c37044d04e5a9a5ba0efd62a we pull the `latest` tag onto the hosts and use it instead - how does that sound?

dhh:
Way nicer! Excellent.

PR #145, opened 2023-03-22
Commander needn't accumulate configuration

jeremy:
Commander had version/destination solely to incrementally accumulate CLI options. Simpler to configure in one shot.

Clarifies responsibility and lets us introduce things like `abbreviated_version` in one spot - Configuration.


jeremy:
Rebased

PR #148, opened 2023-03-23
Fix Discord link

osbre:
Generated a new link that wouldn't expire.


jeremy:
Thanks! However, @dhh should be the inviter so he'll need to generate the link.

PR #150, opened 2023-03-23
Remove accessory images using tags rather than labels

jeremy:
Accessory images are removed with `docker image prune` and a label filter currently, but accessory images are not labeled like app images, so no images are actually removed.

Fix by removing the tagged image, which prunes untagged parents.


dhh:
Some failing tests.

PR #155, opened 2023-03-24
Customizable stop wait time

intrip:
Configurable via a global `stop_wait_time` option. The default is `10` which matches Docker defaults.


intrip:
TIL on non passing default!

dhh:
Oh? I took that from your explanation 😄. That `-t 10` was already the default in Docker?

intrip:
Yes, `10s` is default: That's ok!

Now I see how you implemented, Initially I thought that "docker" command would automatically handle `nil` options but that's not the case 😅

PR #159, opened 2023-03-25
Add cache-from option

nzwsch:
Allowing the option `--cache_from` to the builder in this PR will reduce the time to build the Docker image.

https://lipanski.com/posts/speed-up-your-docker-builds-with-cache-from

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,6 +439,16 @@ RUN --mount=type=secret,id=GITHUB_TOKEN \
   rm -rf /usr/local/bundle/cache
 ```
 
+### Using cache at image build
+
+Specifying a Docker tag allows you to use the cache to save container build time:
+
+```yaml
+builder:
+  cache_from:
+    - latest # 37s/hey:latest
+```
+
 ### Using command arguments for Traefik
 
 You can customize the traefik command line:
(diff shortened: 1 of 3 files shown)


nzwsch:
I ran the command locally and was able to use the cached image from the build as you suggested.
In my use case, I wanted to push the command to GitLab to build, run tests based on the image tag, and if the tests are successful, deploy with the mrsk command using the tag.
However, in this case, I can work around this by doing an empty build of the image once pulled on CI, but I think it would be hard to support bypassing all docker options with the mrsk command, so this PR can be closed.

bdeak4:
Hi @nzwsch, did you manage to get cache working in CI builds?

I've tried something along the lines:
```
docker pull $DOCKER_IMAGE:latest
docker build -t $DOCKER_IMAGE:latest --cache-from $DOCKER_IMAGE:latest .
mrsk deploy
```

But mrsk does not seem to pick up cache from docker build

nzwsch:
@bdeak4
DHH mentioned above that the build image should be used as a cache if an image layer exists. It is not necessary to run the build in the local environment.
```
docker pull $DOCKER_IMAGE:latest
mrsk deploy
```
Please make sure that the image tag is one that you have pulled into your mrsk config file.

PR #171, opened 2023-04-01
Check if env_template exists before envify runs

tannakartikey:
At the moment, if the `mrsk envify` is run when `env.erb` file does not exist, the CLI throws Ruby error:
> $ mrsk envify
> ERROR (Errno::ENOENT): No such file or directory @ rb_sysopen - .env.erb

With this change, Mrsk will first check if the env template file exists and then either continue or print a helpful error message.


dhh:
Hmm, I'm not sure we're improving the error message here materially. This seems actually pretty clear to me.

PR #177, opened 2023-04-03
Update README.md: adds information about what kind of Linux to use; adds requirement to SSH into your distro before running `mrsk deploy`

jasonfb:


README.md
```diff
@@ -19,6 +19,12 @@ gem install mrsk
 ```sh
 alias mrsk='docker run --rm -it -v $HOME/.ssh:/root/.ssh -v /var/run/docker.sock:/var/run/docker.sock -v ${PWD}/:/workdir  ghcr.io/mrsked/mrsk'
 ```
+## Linux Distro
+
+- MRSK works on any Debian-based Linux distribution, like Debian, Mint, or Ubuntu. 
+- Start by confirming you can SSH into your Linux distribution. This will store the key in `~/.ssh/known_hosts` and is a prerequesite for MRSK to deploy.
```
dhh, line 25: This shouldn't be a prerequisite, and has not been so in my testing.

tannakartikey: @jasonfb I removed my server from `known_hosts` and it worked fine for me. In fact, the server got added to the `known_hosts` file automatically. Seems SSHKit takes care of it.

jasonfb: does not make sense. 

How can MRSK know about the location of the `.pem` file without it being passed explicitly or first being added to your ssh agent?

Unless I SSH into my instance first, I get this result when setting up a new EC2 instance (Ubuntu).

% mrsk deploy
Acquiring the deploy lock
  INFO [5ecabcdc] Running /usr/bin/env mkdir mrsk_lock && echo "TG9ja2VkIGJ5OiBKYXNvbiBGbGVldHdvb2QtQm9sZHQgYXQgMjAyMy0wNC0x
MSAxODo1Mjo1OSBVVEMKVmVyc2lvbjogNDMwNGNkNTQ3ZTlhOGRkZTI4Yzcw
OTU3MWI3N2ZkZDgwNGQ5OTc0MwpNZXNzYWdlOiBBdXRvbWF0aWMgZGVwbG95
IGxvY2s=
" > mrsk_lock/details on 3.89.125.213
  ERROR (Mrsk::Cli::Base::LockError): Deploy lock found
jason@MBA22 /Users/jason/Work/_LEARNING/MRSK/mrsk5 [main]
% mrsk lock release
  INFO [5d0ed061] Running /usr/bin/env rm mrsk_lock/details && rm -r mrsk_lock on 3.89.125.213
  ERROR (Net::SSH::AuthenticationFailed): Authentication failed for user ubuntu@3.89.125.213


The steps to reproduce are:
- Make a new EC2 Instance, Ubuntu, generating a new key and downloading the `.pem` file

- Make a brand new Rails app, then `gem install mrsk` and `mrsk init`


- Edit config file with IP address of new instance,

- Run `mrsk deploy`, you get incorrect error message `ERROR (Mrsk::Cli::Base::LockError): Deploy lock found`


- Run `mrsk lock release` you get: 

  ERROR (Net::SSH::AuthenticationFailed): Authentication failed for user ubuntu@3.89.125.213


I think the keyfile (`.pem`) needs to be set in the ssh-agent, and the only way I know how to do that is to ssh first. Is there a way to add the `.pem` file into the MRSK configuration. is that possible?

The only way I can get it authenticated is to first run this, which seems to add it to the ssh agent, otherwise the ssh agent doesn't know about my keyfile

what part is missing here?

dhh: I think something else is going on. MRSK just ssh on your behalf. So there shouldn't be any difference between you doing it by hand first or letting MRSK do it. But you do need to ssh-add the relevant key first of course. Maybe that's what you're missing?

jasonfb: thank you, this is done easily one of two ways:
1) ssh -i <with identity file>
or
2) ssh-add <with identity file>

PR #179, opened 2023-04-03
Update README.md — edits "vision" section to remove some hype (written by Chat-GPT)

jasonfb:
I told ChatGPT to "rewrite the following paragraph but remove the oversold promises and hype language" and here's what I got


dhh:
Don't see an improvement here.

PR #182, opened 2023-04-05
Only redact the non-sensitive bits of build args and env vars.

### LINKED ISSUE #96: "mrsk config" exposes env secrets
There will be cases when a user is asked to copy the output of `mrsk config` for debugging. Currently it's a bad idea since all secrets are exposed in `env_args`, which also makes the following statement in README false:

`Marking an ENV as secret currently only redacts its value in the output for MRSK.`

jeremy:
* `-e [REDACTED]` → `-e SOME_SECRET=[REDACTED]`
* Replaces `Utils.redact` with `Utils.sensitive` to clarify that we're indicating redactability, not actually performing redaction.
* Redacts from YAML output, including `mrsk config` (fixes #96)


dhh:
Test failures due to Ruby install?

intrip:
https://github.com/mrsked/mrsk/pull/184 should fix CI

PR #183, opened 2023-04-05
Clear stale containers

intrip:
By stopping all the older containers matching `/#{service}-#{role}-#{dest}-.*/` running on the same host.

### DIFF
--- a/lib/mrsk/cli/app.rb
+++ b/lib/mrsk/cli/app.rb
@@ -22,7 +22,7 @@ class Mrsk::Cli::App < Mrsk::Cli::Base
                 execute *MRSK.app(role: role).rename_container(version: version, new_version: tmp_version)
               end
 
-              old_version = capture_with_info(*MRSK.app(role: role).current_running_version).strip
+              old_version = capture_with_info(*MRSK.app(role: role).current_running_version, raise_on_non_zero_exit: false).strip
               execute *MRSK.app(role: role).run
               sleep MRSK.config.readiness_delay
               execute *MRSK.app(role: role).stop(version: old_version), raise_on_non_zero_exit: false if old_version.present?
@@ -260,7 +260,7 @@ class Mrsk::Cli::App < Mrsk::Cli::Base
 
     def list_versions(host:, role: nil, status: nil)
       versions = nil
-      on(host) { versions = capture_with_info(*MRSK.app(role: role).list_versions(status: status)).split("\n").map(&:strip) }
+      on(host) { versions = capture_with_info(*MRSK.app(role: role).list_versions(status: status), raise_on_non_zero_exit: false).split("\n").map(&:strip) }
       versions
     end
 
--- a/lib/mrsk/cli/app.rb
+++ b/lib/mrsk/cli/app.rb
@@ -134,13 +134,7 @@ class Mrsk::Cli::App < Mrsk::Cli::Base
         roles = MRSK.roles_on(host)
 
         roles.each do |role|
-          stale_versions = \
-            capture_with_info(*MRSK.app(role: role).list_versions, raise_on_non_zero_exit: false)
-            .split("\n")
-            .map(&:strip)
-            .drop(1)
-
-          stale_versions.each do |version|
+          stale_versions(role: role).each do |version|
             if stop
               puts_by_host host, "Stopping stale container for role #{role} with version #{version}"
               execute *MRSK.app(role: role).stop(version: version), raise_on_non_zero_exit: false
--- a/lib/mrsk/cli/app.rb
+++ b/lib/mrsk/cli/app.rb
@@ -130,11 +130,13 @@ class Mrsk::Cli::App < Mrsk::Cli::Base
     with_lock do
       stop = options[:stop]
 
+      cli = self
+
       on(MRSK.hosts) do |host|
         roles = MRSK.roles_on(host)
 
         roles.each do |role|
-          stale_versions(role: role).each do |version|
+          cli.stale_versions(host: host, role: role).each do |version|
             if stop
               puts_by_host host, "Stopping stale container for role #{role} with version #{version}"
               execute *MRSK.app(role: role).stop(version: version), raise_on_non_zero_exit: false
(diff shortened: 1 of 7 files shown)


lib/mrsk/commands/app.rb
```diff
@@ -82,20 +82,22 @@ def execute_in_new_container_over_ssh(*command, host:)
...
+    list_versions("--latest", status: :running)
+  end
+
+  def list_versions(*docker_args, status: nil)
     pipe \
-      docker(:ps, *filter_args, "--format", '"{{.Names}}"'),
-      %(sed 's/-/\\n/g'),
-      "tail -n 1"
+      docker(:ps, *filter_args(status: status), *docker_args, "--format", '"{{.Names}}"'),
+      %(grep -oP "(?<=\-)[^-]+$") # Extract SHA from "service-role-dest-SHA"
```
jeremy, line 100: `-P` is a GNU grep arg. Got a more portable alternative?

lib/mrsk/cli/app.rb
```diff
@@ -50,12 +50,32 @@ def start
...
         roles = MRSK.roles_on(host)
 
         roles.each do |role|
-          execute *MRSK.auditor(role: role).record("Stopped app"), verbosity: :debug
-          execute *MRSK.app(role: role).stop, raise_on_non_zero_exit: false
+          stale_versions = list_versions(host: host, role: role).drop(1)
+          stale_versions.each do |version|
+            say "Stopping stale container with version #{version}", :red
+            on(host) { execute *MRSK.app(role: role).stop(version: version), raise_on_non_zero_exit: false }
+          end
```
jeremy, line 78: Possible to share implementation with `stop`? Both are stopping containers, just with different versions.

lib/mrsk/cli/main.rb
```diff
@@ -37,6 +37,9 @@ def deploy
         say "Ensure app can pass healthcheck...", :magenta
         invoke "mrsk:cli:healthcheck:perform", [], invoke_options
 
+        say "Stop stale containers...", :magenta
+        invoke "mrsk:cli:app:stop_stale_containers", [], invoke_options
+
```
jeremy, line 42: Perhaps `boot` should stop all old containers rather than just the previous version?

lib/mrsk/commands/app.rb
intrip: TIL. The reason for using `-P` was to rely on lookbehind to capture only the SHA. I come up with a combination of grep + cut to achieve it instead.

lib/mrsk/cli/app.rb
intrip: Good idea 🙏

lib/mrsk/cli/main.rb
intrip: Yes!

intrip:
CI failure will be fixed in https://github.com/mrsked/mrsk/pull/184

lib/mrsk/cli/app.rb
```diff
@@ -8,24 +8,30 @@ def boot
 
         cli = self
 
-        on(MRSK.hosts) do |host|
+        MRSK.hosts.each do |host|
```
intrip, line 11: This might introduce a perf regression , will take a deeper look.

dhh:
Hmm, I don't think the complexity here is proportionate to the problem. I think it'd be worth looking at a simpler solution where we simply warn about duplicates and then perhaps having an explicit command to cleanup.

lib/mrsk/cli/app.rb
```diff
@@ -22,7 +22,7 @@ def boot
                 execute *MRSK.app(role: role).rename_container(version: version, new_version: tmp_version)
               end
 
-              old_version = capture_with_info(*MRSK.app(role: role).current_running_version).strip
+              old_version = capture_with_info(*MRSK.app(role: role).current_running_version, raise_on_non_zero_exit: false).strip
```
intrip, line 25: To avoid errors when a deployment run after a stop.

intrip:
> Hmm, I don't think the complexity here is proportionate to the problem. I think it'd be worth looking at a simpler solution where we simply warn about duplicates and then perhaps having an explicit command to cleanup.

I've introduced a specific command to detect and optionally clear stale containers. By default only detects stale containers during a deploy.

dhh:
Much nicer!

lib/mrsk/cli/app.rb
```diff
@@ -124,6 +124,35 @@ def containers
...
+  option :stop, aliases: "-s", type: :boolean, default: false, desc: "Stop the stale containers found"
+  def stale_containers
+    with_lock do
+      stop = options[:stop]
+
+      on(MRSK.hosts) do |host|
+        roles = MRSK.roles_on(host)
+
+        roles.each do |role|
+          stale_versions = \
```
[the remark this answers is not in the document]
intrip: Handled in https://github.com/mrsked/mrsk/pull/183/commits/12e3a562c4cae631045a4952ce6c197d40af93f7
The reason to put it in the SSHKIT layer is that otherwise, we can't leverage the `on(hosts)` parallelism.

lib/mrsk/sshkit_with_ext.rb
```diff
@@ -1,9 +1,19 @@
 require "sshkit"
 require "sshkit/dsl"
 
+module AppHelper
+  def stale_versions(role:)
```
dhh, line 5: Let's use the `cli = self` pattern that you can find a few other places in the codebase instead 👍

intrip: We could also make it private but the issue is that at that point I'm not sure on how to leverage the on(host) parallelism.

intrip: I used the `no_commands` block to make it explicit that it's not a command

lib/mrsk/cli/app.rb
```diff
@@ -219,6 +244,19 @@ def version
     on(MRSK.hosts) { |host| puts_by_host host, capture_with_info(*MRSK.app.current_running_version).strip }
   end
 
+  no_commands do
```
dhh, line 247: imo better just to make this private and then use #send.

intrip: Changed to `private`

dhh:
Great! Just needs a rebase.

intrip:
Conflicts fixed now

PR #186, opened 2023-04-06
Add labels to custom command containers

djmb:
Add the --labels (aliased to -l) option to `mrsk app exec` and `mrsk accessory exec`. This allow us to add custom labels to the containers.

Example labels:
- The user who started the container
- Mark containers that we expect to run for a long time
- Add a time by which the command is expected to finish

Labels are passed verbatim to `docker run`.

Additionally adds an `custom` label to all app and accessory containers. The value is true for ones started via `mrsk app|accessory exec` and false otherwise.

### DIFF
--- a/lib/mrsk/cli/app.rb
+++ b/lib/mrsk/cli/app.rb
@@ -76,7 +76,10 @@ class Mrsk::Cli::App < Mrsk::Cli::Base
   desc "exec [CMD]", "Execute a custom command on servers (use --help to show options)"
   option :interactive, aliases: "-i", type: :boolean, default: false, desc: "Execute command over ssh for an interactive shell (use for console/bash)"
   option :reuse, type: :boolean, default: false, desc: "Reuse currently running container instead of starting a new one"
+  option :labels, aliases: "-l", type: :array, default: [], desc: "Labels to apply to the container (not compatible with --reuse)", banner: "label1=value1 label2=value2 label3"
   def exec(cmd)
+    labels = options[:labels]
+    raise ArgumentError, "Cannot add labels to reused containers" if options[:reuse] && labels.any?
     case
     when options[:interactive] && options[:reuse]
       say "Get current version of running container...", :magenta unless options[:version]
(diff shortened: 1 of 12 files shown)


lib/mrsk/cli/app.rb
```diff
@@ -76,7 +76,10 @@ def details
   desc "exec [CMD]", "Execute a custom command on servers (use --help to show options)"
   option :interactive, aliases: "-i", type: :boolean, default: false, desc: "Execute command over ssh for an interactive shell (use for console/bash)"
   option :reuse, type: :boolean, default: false, desc: "Reuse currently running container instead of starting a new one"
+  option :labels, aliases: "-l", type: :array, default: [], desc: "Labels to apply to the container (not compatible with --reuse)", banner: "label1=value1 label2=value2 label3"
```
djmb, line 79: Thor option arrays are greedy for parameters that follow the option flag, so you can't have the command listed straight after them, because it will treat it as a label.

So instead of:
```
$ mrsk app exec -i -l foo=bar bash
ERROR: "mrsk app exec" was called with no arguments
Usage: "mrsk app exec [CMD]"
```

You'd need one of:
```
mrsk app exec bash -i -l foo=bar
mrsk app exec -i bash -l foo=bar
mrsk app exec -l foo=bar -i bash
```

lib/mrsk/commands/accessory.rb
```diff
@@ -57,12 +58,14 @@ def execute_in_existing_container(*command, interactive: false)
...
   end
 
-  def execute_in_new_container(*command, interactive: false)
+  def execute_in_new_container(*command, interactive: false, labels: [])
     docker :run,
       ("-it" if interactive),
       "--rm",
       *env_args,
       *volume_args,
+      "--label", "custom=true",
```
djmb, line 67: I think we should also add `config.label_args` here so the custom containers are labeled with the service and destination. 

We'll need to wait for https://github.com/mrsked/mrsk/pull/183 to land first though so we can provide a clear upgrade path.

We could also consider giving the container a meaningful name, maybe using `custom` as a special role and appending a random string to allow multiple instances.

dhh:
Will drive this from specific features like auto-cleanup of consoles instead 👍

PR #193, opened 2023-04-08
Don't throw error if primary_web_host is not found

### LINKED ISSUE #191: Config without web role is not supported
My config files look as described below. It does not have `web` role. Since I am running everything on a single server, it makes more sense to have the roles defined as `staging` and `production` or `web-staging` and `web-production`.

Here, we need to make the check optional:
https://github.com/mrsked/mrsk/blob/2146f6d0ecbdc675226daabd5a3dcaa0ce389e0f/lib/mrsk/configuration.rb#L82-L84

I think it will affect the `--primary` option if we modify the code to `role(:web)&.primary_host`

My config files:

```yml
# config/deploy.yml
service: myapp
image: user/myapp
```

```yml
# config/deploy.staging.yml
servers:
  staging:
    traefik: true
    hosts:
      - 8.8.8.8
```

```yml
# config/deploy.production.yml
servers:
  production:
    traefik: true
    hosts:
      - 8.8.8.8
```

tannakartikey:
Resolves #191

### DIFF
--- a/lib/mrsk/configuration.rb
+++ b/lib/mrsk/configuration.rb
@@ -80,7 +80,7 @@ class Mrsk::Configuration
   end
 
   def primary_web_host
-    role(:web).primary_host
+    role(:web)&.primary_host
   end
 
   def traefik_hosts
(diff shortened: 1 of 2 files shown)


dhh:
That just throws the problem further down the chain. There needs to always be a web primary host, and MRSK should be able to depend on that.

PR #197, opened 2023-04-09
Traefik service name to be derived from role and destination

tannakartikey:
Resolves https://github.com/mrsked/mrsk/issues/190

Provides the scope for granular control over Traefik rules. Traefik rules can not be targetted towards any role or destination and are not limited just to the service anymore.


northeastprince:
This looks great! Also really needed for multiple destinations (prod, staging, etc.) on the same server cause I thought this would come with #71.

PR #205, opened 2023-04-10
Bootstrap: multi-OS Docker install

jeremy:
* Limit auto-install to root users; otherwise, give manual install guidance
* Support non-Debian/Ubuntu with the multi-OS get.docker.com installer
* Drop native curl dependency by running it with a small Docker image
* Fix up healthcheck test mocks to reflect actual operations

### DIFF
--- a/lib/mrsk/cli/main.rb
+++ b/lib/mrsk/cli/main.rb
@@ -17,9 +17,6 @@ class Mrsk::Cli::Main < Mrsk::Cli::Base
       invoke_options = deploy_options
 
       runtime = print_runtime do
-        say "Ensure curl and Docker are installed...", :magenta
-        invoke "mrsk:cli:server:bootstrap", [], invoke_options
-
         say "Log into image registry...", :magenta
         invoke "mrsk:cli:registry:login", [], invoke_options
 
--- a/lib/mrsk/cli/server.rb
+++ b/lib/mrsk/cli/server.rb
@@ -1,17 +1,21 @@
 class Mrsk::Cli::Server < Mrsk::Cli::Base
-  desc "bootstrap", "Ensure curl and Docker are installed on servers"
+  desc "bootstrap", "Set up Docker to run MRSK apps"
   def bootstrap
-    with_lock do
-      on(MRSK.hosts + MRSK.accessory_hosts) do
-        dependencies_to_install = Array.new.tap do |dependencies|
-          dependencies << "curl" unless execute "which curl", raise_on_non_zero_exit: false
-          dependencies << "docker.io" unless execute "which docker", raise_on_non_zero_exit: false
-        end
+    missing = []
 
-        if dependencies_to_install.any?
-          execute "apt-get update -y && apt-get install #{dependencies_to_install.join(" ")} -y"
+    on(MRSK.hosts | MRSK.accessory_hosts) do |host|
+      unless execute(*MRSK.docker.installed?, raise_on_non_zero_exit: false)
+        if execute(*MRSK.docker.superuser?, raise_on_non_zero_exit: false)
+          info "Missing Docker on #{host}. Installing…"
+          execute *MRSK.docker.install
+        else
+          missing << host
         end
       end
+    end
+
+    if missing.any?
+      raise "Docker is not installed on #{missing.join(", ")} and can't be automatically installed without having root access and the `curl` command available. Install Docker manually: https://docs.docker.com/engine/install/"
     end
   end
 end
--- a/test/cli/server_test.rb
+++ b/test/cli/server_test.rb
@@ -1,11 +1,30 @@
 require_relative "cli_test_case"
 
 class CliServerTest < CliTestCase
-  test "bootstrap" do
+  test "bootstrap already installed" do
+    SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:docker, "-v", raise_on_non_zero_exit: false).returns(true).at_least_once
+
+    assert_equal "", run_command("bootstrap")
+  end
+
+  test "bootstrap install as non-root user" do
+    SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:docker, "-v", raise_on_non_zero_exit: false).returns(false).at_least_once
+    SSHKit::Backend::Abstract.any_instance.expects(:execute).with('[ "${EUID:-$(id -u)}" -eq 0 ]', raise_on_non_zero_exit: false).returns(false).at_least_once
+
+    assert_raise RuntimeError, "Docker is not installed on 1.1.1.1, 1.1.1.3, 1.1.1.4, 1.1.1.2 and can't be automatically intalled without having root access and the `curl` command available. Install Docker manually: https://docs.docker.com/engine/install/" do
+      run_command("bootstrap")
+    end
+  end
+
+  test "bootstrap install as root user" do
+    SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:docker, "-v", raise_on_non_zero_exit: false).returns(false).at_least_once
+    SSHKit::Backend::Abstract.any_instance.expects(:execute).with('[ "${EUID:-$(id -u)}" -eq 0 ]', raise_on_non_zero_exit: false).returns(true).at_least_once
+    SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:curl, "-fsSL", "https://get.docker.com", "|", :sh).at_least_once
+
     run_command("bootstrap").tap do |output|
-      assert_match /which curl/, output
-      assert_match /which docker/, output
-      assert_match /apt-get update -y && apt-get install curl docker.io -y/, output
+      ("1.1.1.1".."1.1.1.4").map do |host|
+        assert_match "Missing Docker on #{host}. Installing…", output
+      end
     end
   end
 
--- a/lib/mrsk/commander.rb
+++ b/lib/mrsk/commander.rb
@@ -84,6 +84,10 @@ class Mrsk::Commander
     @builder ||= Mrsk::Commands::Builder.new(config)
   end
 
+  def docker
+    @docker ||= Mrsk::Commands::Docker.new(config)
+  end
+
   def healthcheck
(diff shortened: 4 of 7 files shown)


lib/mrsk/cli/healthcheck.rb
```diff
@@ -1,7 +1,11 @@
 class Mrsk::Cli::Healthcheck < Mrsk::Cli::Base
   MAX_ATTEMPTS = 7
 
-  class HealthcheckError < StandardError; end
+  class HealthcheckError < RuntimeError
+    def initialize(message)
+      super "Health check against #{MRSK.config.healthcheck["path"]} #{message}"
```
jeremy, line 6: Obviates passing `target` around

lib/mrsk/cli/healthcheck.rb
```diff
@@ -11,36 +15,33 @@ def perform
...
 
-        target = "Health check against #{MRSK.config.healthcheck["path"]}"
         attempt = 1
-
         begin
-          status = capture_with_info(*MRSK.healthcheck.curl)
-
-          if status == "200"
-            info "#{target} succeeded with 200 OK!"
+          case output = capture_with_info(*MRSK.healthcheck.curl, raise_on_non_zero_exit: false)
```
jeremy, line 20: We were raising on non-200-OK curl responses previously, bypassing the status parsing conditions below.

Instead, don't raise, eliminating `SSHKit::Command::Failed` handling, giving us just retries on `HealthcheckError` to worry about.

lib/mrsk/cli/healthcheck.rb
```diff
@@ -11,36 +15,33 @@ def perform
...
+            raise HealthcheckError, "failed with status #{output}"
           else
-            raise HealthcheckError, "#{target} failed with status #{status}"
+            raise HealthcheckError, "failed to run: #{output}"
           end
-        rescue SSHKit::Command::Failed
+        rescue HealthcheckError => error
           if attempt <= MAX_ATTEMPTS
-            info "#{target} failed to respond, retrying in #{attempt}s..."
+            info "#{error.message}, retrying in #{attempt}s..."
```
jeremy, line 32: Now indicates why the health check failed.

lib/mrsk/cli/healthcheck.rb
```diff
@@ -11,36 +15,33 @@ def perform
...
           end
         end
-      rescue SSHKit::Command::Failed, HealthcheckError => e
-        error capture_with_info(*MRSK.healthcheck.logs)
 
-        if e.message =~ /curl/
-          raise SSHKit::Command::Failed, "#{target} failed to return 200 OK!"
-        else
-          raise
-        end
```
jeremy, line 43: YAGNI

lib/mrsk/cli/healthcheck.rb
```diff
@@ -11,36 +15,33 @@ def perform
...
+        info "Health check succeeded with 200 OK!"
```
jeremy, line 41: Indicate success at the end, if nothing raised/failed, rather than within output status parsing.

lib/mrsk/cli/main.rb
```diff
@@ -17,9 +17,6 @@ def deploy
       invoke_options = deploy_options
 
       runtime = print_runtime do
-        say "Ensure curl and Docker are installed...", :magenta
-        invoke "mrsk:cli:server:bootstrap", [], invoke_options
-
```
jeremy, line 22: No longer need slow/noisy per-deploy bootstrapping if we aren't auto-installing Docker, considering any subsequent Docker usage is equivalent to an "is Docker ready?" check.

lib/mrsk/cli/server.rb
```diff
@@ -1,17 +1,14 @@
...
-        if dependencies_to_install.any?
-          execute "apt-get update -y && apt-get install #{dependencies_to_install.join(" ")} -y"
-        end
-      end
+    on(MRSK.hosts + MRSK.accessory_hosts) do |host|
+      unready << host unless execute(:docker, :info, raise_on_non_zero_exit: false)
+    end
+
+    if unready.any?
+      raise "Docker is not running on #{unready.join(", ")}. Install it first: https://docs.docker.com/engine/install/"
```
jeremy, line 11: Indicate how to install Docker rather than automating it. That only worked with SSH root users—which most hosts strongly discourage—and on Debian/Ubuntu hosts. Clearer to indicate what needs to be set up. Allows MRSK to remain indifferent to deploy target platform.

lib/mrsk/commands/healthcheck.rb
```diff
@@ -18,7 +20,7 @@ def run
   end
 
   def curl
-    [ :curl, "--silent", "--output", "/dev/null", "--write-out", "'%{http_code}'", "--max-time", "2", health_url ]
+    [ *config.healthcheck.fetch("curl"), *CURL_ARGS, health_url ]
```
jeremy, line 23: `curl` option now defaults to a `docker` invocation, limiting MRSK deps to Docker alone. No native curl install required.

test/cli/healthcheck_test.rb
```diff
@@ -15,44 +18,61 @@ class CliHealthcheckTest < CliTestCase
...
     run_command("perform").tap do |output|
-      assert_match "Health check against /up failed to respond, retrying in 1s...", output
-      assert_match "Health check against /up failed to respond, retrying in 2s...", output
-      assert_match "Health check against /up succeeded with 200 OK!", output
+      assert_match "Health check against /up failed to connect, retrying in 1s...", output
+      assert_match "Health check against /up failed with status 500, retrying in 2s...", output
+      assert_match "Health check against /up failed to run: unknown, retrying in 3s...", output
+      assert_match "Health check succeeded with 200 OK!", output
+    end
+  end
```
jeremy, line 33: Reworked tests to demonstrate actual failure output and sequencing.

test/cli/healthcheck_test.rb
```diff
@@ -15,44 +18,61 @@ class CliHealthcheckTest < CliTestCase
...
+      .with(:docker, :run, "--rm", "curlimages/curl:latest", "--silent", "--output", "/dev/null", "--write-out", "'%{http_code}'", "--max-time", "2", "http://localhost:3999/up", raise_on_non_zero_exit: false)
+      .returns("000")
     SSHKit::Backend::Abstract.any_instance.stubs(:capture_with_info)
       .with(:docker, :container, :ls, "--all", "--filter", "name=^healthcheck-app-999$", "--quiet", "|", :xargs, :docker, :logs, "--tail", 50, "2>&1")
 
     exception = assert_raises SSHKit::Runner::ExecuteError do
       run_command("perform")
     end
-    assert_match "Health check against /up failed to return 200 OK!", exception.message
+    assert_equal "Exception while executing on host 1.1.1.1: Health check against /up failed to connect", exception.message
```
jeremy, line 66: We no longer have "missing curl" errors to worry about. Add a test for `000` status, indicating the curl didn't get an HTTP response at all.

test/cli/server_test.rb
```diff
@@ -2,10 +2,16 @@
...
+    SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:docker, :info, raise_on_non_zero_exit: false).returns(true).at_least_once
+
+    assert_equal "", run_command("bootstrap")
+  end
+
+  test "bootstrap with Docker not running" do
+    SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:docker, :info, raise_on_non_zero_exit: false).returns(false).at_least_once
+
+    assert_raise RuntimeError, "Docker is not running on 1.1.1.1, 1.1.1.3, 1.1.1.4, 1.1.1.2. Install it first: https://docs.docker.com/engine/install/" do
+      run_command("bootstrap")
```
jeremy, line 14: New bootstrap output indicates where Docker isn't running and how to install it.

test/commands/healthcheck_test.rb
```diff
@@ -39,15 +39,23 @@ class CommandsHealthcheckTest < ActiveSupport::TestCase
...
-      "curl --silent --output /dev/null --write-out '%{http_code}' --max-time 2 http://localhost:3999/healthz",
+      "docker run --rm curlimages/curl:latest --silent --output /dev/null --write-out '%{http_code}' --max-time 2 http://localhost:3999/healthz",
+      new_command.curl.join(" ")
+  end
+
+  test "curl with custom curl command" do
+    @config[:healthcheck] = { "curl" => [ :curl, :args ] }
+
+    assert_equal \
+      "curl args --silent --output /dev/null --write-out '%{http_code}' --max-time 2 http://localhost:3999/up",
```
jeremy, line 58: Can override the curl invocation.

lib/mrsk/cli/server.rb
dhh: Hmm, Digital Ocean and Hetzner are two examples where you can just use the root user to do the install. I don't mind that we catch missing docker better, but it seems like a step backwards to me without an upside to not automate this step when possible.

lib/mrsk/commands/healthcheck.rb
dhh: Love this.

lib/mrsk/cli/server.rb
javierav: IMHO, the option proposed by @jeremy is better, otherwise mrsk has to detect if the user is root and implement different installation methods to support systems based on apt, yum, etc. which complicates the logic a lot. And in the case of ubuntu, the recommended option is to use the binaries from the docker repository instead of the ones that ubuntu brings by default in its repositories. If you want to automate the docker installation process when creating the instance, you can always use cloud-init to provision it.

jeremy: Updated to attempt an auto-install for root users only using the [multi-OS install script](https://get.docker.com).

It's not foolproof, but it relieves MRSK of trying to drive multiple OS and packaging systems.

jeremy:
Updated to attempt an auto-install—but for root users only—using the [multi-OS install script](https://get.docker.com).

It's not foolproof, but it relieves MRSK of trying to drive multiple OS and packaging systems.

dhh:
Think we need a rebase after the new health check strategy right?

PR #208, opened 2023-04-12
Labels can be added to Traefik

tannakartikey:
If we need to configure the Traefik container with labels, we need the option to attach labels to the Traefik container itself. 

For example, we need this configure secure dashboard for Traefik as described in Traefik docs:
https://doc.traefik.io/traefik/operations/dashboard/#dashboard-router-rule


tannakartikey:
If you try to set Traefik Dashboard basic auth then it will not work because of the wrong escape for the `$` symbol. https://github.com/mrsked/mrsk/pull/144/files fixes that.

PR #219, opened 2023-04-13
Move health checks into Docker

kevinmcconnell:
Replaces our current host-based HTTP healthchecks with Docker healthchecks, and adds a new `healthcheck.cmd` config option that can be used to define a custom health check command. Also removes Traefik's healthchecks, since they are no longer necessary (Traefik watches the Docker health change events).

When deploying a container that has a healthcheck defined, we wait for it to report a healthy status before stopping the old container that it replaces. Containers that don't have a healthcheck defined continue to wait for `MRSK.config.readiness_delay`.

There are some pros and cons to using Docker healthchecks rather than checking from the host. The main advantages are:

- Supports non-HTTP checks, app-specific check scripts provided by a container, etc.
- When booting a container, allows MRSK to wait for it to be healthy before shutting down the old container it replaces. This should be safer than relying on a timeout after start.
- Containers with healthchecks won't become active in Traefik until they reach a healthy state, which prevents any traffic from being routed to them before they are ready.
- Health status becomes visible in other tooling, like `docker ps` output, `mrsk app details`, etc.

The main _disadvantage_ is that containers are now required to provide some way to check their health. MRSK's default check assumes that `curl` is available in the container which, while common, won't always be the case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,9 +677,11 @@ That'll post a line like follows to a preconfigured chatbot in Basecamp:
 [My App] [dhh] Rolled back to version d264c4e92470ad1bd18590f04466787262f605de
 ```
 
-### Custom healthcheck
+### Healthcheck
 
-MRSK defaults to checking the health of your application again `/up` on port 3000 up to 7 times. You can tailor the behaviour with the `healthcheck` setting:
+MRSK uses Docker healtchecks to check the health of your application during deployment. Traefik uses this same healthcheck status to determine when a container is ready to receive traffic.
+
+The healthcheck defaults to testing the HTTP response to the path `/up` on port 3000, up to 7 times. You can tailor this behaviour with the `healthcheck` setting:
 
 ```yaml
 healthcheck:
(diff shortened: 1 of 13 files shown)


lib/mrsk/commands/healthcheck.rb
```diff
@@ -11,14 +11,15 @@ def run
...
       *web.option_args,
       config.absolute_image,
       web.cmd
   end
 
-  def curl
-    [ :curl, "--silent", "--output", "/dev/null", "--write-out", "'%{http_code}'", "--max-time", "2", health_url ]
+  def status
+    pipe container_id, xargs(docker(:inspect, "--format", DOCKER_HEALTH_STATUS_FORMAT))
   end
```
kevinmcconnell, line 23: Healthcheck and App both define a `status` method to fetch container status. It's a bit same-y, but the container name/id differs in each, and letting them figure that out themselves ended up looking clearer.

lib/mrsk/configuration/role.rb
```diff
@@ -75,8 +90,6 @@ def traefik_labels
       if running_traefik?
         {
           "traefik.http.routers.#{traefik_service}.rule" => "PathPrefix(`/`)",
-          "traefik.http.services.#{traefik_service}.loadbalancer.healthcheck.path" => config.healthcheck["path"],
-          "traefik.http.services.#{traefik_service}.loadbalancer.healthcheck.interval" => "1s",
```
kevinmcconnell, line 79: We don't need Traefik to healthcheck since Docker is doing it. Also this wouldn't work for any custom `cmd` anyway.

lib/mrsk/utils/healthcheck_poller.rb
```diff
@@ -0,0 +1,39 @@
+class Mrsk::Utils::HealthcheckPoller
```
kevinmcconnell, line 1: Named this `HealthcheckPoller` rather than something like `Healthcheck`, since the actual healthcheck is provided for us by Docker. This class just needs to know how to wait for the status to reach something runnable.

README.md
```diff
@@ -677,9 +677,11 @@ That'll post a line like follows to a preconfigured chatbot in Basecamp:
 [My App] [dhh] Rolled back to version d264c4e92470ad1bd18590f04466787262f605de
 ```
 
-### Custom healthcheck
+### Healthcheck
 
-MRSK defaults to checking the health of your application again `/up` on port 3000 up to 7 times. You can tailor the behaviour with the `healthcheck` setting:
+MRSK uses Docker healthecks to check the health of your application during deployment. Traefik uses this same healthcheck status to determine when a container is ready to receive traffic.
```
intrip, line 682:
```suggestion
MRSK uses Docker healthchecks to check the health of your application during deployment. Traefik uses this same healthcheck status to determine when a container is ready to receive traffic.
```

intrip (approved):
Fantastic 👏👏

README.md
kevinmcconnell: Thanks @intrip! 🦅👀

tbuehlmann:
Oh this seems nice.

PR #222, opened 2023-04-13
Allow booting containers in groups for rolling restarts

kevinmcconnell:
Adds top-level configuration options for `group_limit` and `group_wait`. When a `group_limit` is present, we'll perform app boot operations on no more than `group_limit` hosts at a time, optionally sleeping for `group_wait` seconds after each batch.

We currently only do this batching on boot (including when that's part of a deployment). Other commands, like `app stop` or `app details` still work on all hosts in parallel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -831,11 +831,11 @@ mrsk lock acquire -m "Doing maintanence"
 mrsk lock release
 ```
 
-## Gradual restarts
+## Rolling deployments
 
 When deploying to large numbers of hosts, you might prefer not to restart your services on every host at the same time.
 
-MRSK's default is to start new containers on all hosts in parallel. But you can control this by configuring `group_limit` and `group_wait`.
+MRSK's default is to boot new containers on all hosts in parallel. But you can control this by configuring `group_limit` and `group_wait`.
 
 ```yaml
 service: myservice
@@ -844,9 +844,9 @@ group_limit: 10
 group_wait: 30
 ```
 
-When `group_limit` is specified, containers will be started on, at most, `group_limit` hosts at once. MRSK will pause for `group_wait` seconds between batches.
+When `group_limit` is specified, containers will be booted on, at most, `group_limit` hosts at once. MRSK will pause for `group_wait` seconds between batches.
 
-These settings only apply when starting containers (using `mrsk deploy`, `mrsk app boot` or `mrsk app start`). For other commands, MRSK continues to run commands in parallel across all hosts.
+These settings only apply when booting containers (using `mrsk deploy`, or `mrsk app boot`). For other commands, MRSK continues to run commands in parallel across all hosts.
 
 ## Stage of development
 
--- a/README.md
+++ b/README.md
@@ -841,7 +841,7 @@ MRSK's default is to boot new containers on all hosts in parallel. But you can c
 service: myservice
 
 boot:
-  group_limit: 10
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -831,6 +831,23 @@ mrsk lock acquire -m "Doing maintanence"
 mrsk lock release
 ```
 
+## Gradual restarts
+
+When deploying to large numbers of hosts, you might prefer not to restart your services on every host at the same time.
+
+MRSK's default is to start new containers on all hosts in parallel. But you can control this by configuring `group_limit` and `group_wait`.
```
jeremy, line 838: As top-level config, these options are pretty generic. It's hard to tell how they apply. Could they live within a new config section for app boot?

README.md
```diff
@@ -831,6 +831,23 @@ mrsk lock acquire -m "Doing maintanence"
...
+
+```yaml
+service: myservice
+
+group_limit: 10
```
jeremy, line 843: Would like to be able to specify # of groups (or % of hosts) rather than fixed group size so the rolling restart size will scale up/down with host count. In Capistrano, we do 25% of hosts in each group. (Likely out of scope for this PR.)

README.md
```diff
@@ -831,6 +831,23 @@ mrsk lock acquire -m "Doing maintanence"
...
+group_wait: 30
```
jeremy, line 844: Would dial this down since folks may treat it as a default to start with. We use 2-second waits for Capistrano rolling deploys, by contrast.

README.md
```diff
@@ -831,6 +831,23 @@ mrsk lock acquire -m "Doing maintanence"
...
+```
+
+When `group_limit` is specified, containers will be started on, at most, `group_limit` hosts at once. MRSK will pause for `group_wait` seconds between batches.
+
+These settings only apply when starting containers (using `mrsk deploy`, `mrsk app boot` or `mrsk app start`). For other commands, MRSK continues to run commands in parallel across all hosts.
```
jeremy, line 849: If we focus on rolling deployments, we can limit this to `app:boot` alone.

README.md
```diff
@@ -831,6 +831,23 @@ mrsk lock acquire -m "Doing maintanence"
 mrsk lock release
 ```
 
+## Gradual restarts
+
+When deploying to large numbers of hosts, you might prefer not to restart your services on every host at the same time.
```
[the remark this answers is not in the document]
kevinmcconnell: That's true. I'd included `start` because bringing up a set of new hosts could have a similar effect. But that's probably rare, and if they're actually new hosts you could also just add them in batches.

Will change to boot only 👍

README.md
kevinmcconnell: Yeah, good idea 👍

README.md
kevinmcconnell: Good call. Could have the setting accept either format, like `10` or `25%`.

kevinmcconnell: Added in https://github.com/mrsked/mrsk/pull/222/commits/5ea2e20e51408fcfde4cc813cb44de70bbe002d9

README.md
kevinmcconnell: I've moved these under a new `boot` section.

README.md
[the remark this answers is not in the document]
kevinmcconnell: Changed in https://github.com/mrsked/mrsk/pull/222/commits/100b72e4b48d4a887746ddf0f8981b1de1096dcf

PR #223, opened 2023-04-13
Allow customizing audit broadcast with env

kevinmcconnell:
When invoking the audit broadcast command, provide a few environment variables so that people can customize the format of the message if they want, or use the values to send structured messages to APIs, and so on.

We currently provide `MRSK_PERFORMER`, `MRSK_ROLE`, `MRSK_DESTINATION` and `MRSK_MESSAGE`.

Also adds the destination to the default message, which we continue to send as the first argument as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,6 +677,15 @@ That'll post a line like follows to a preconfigured chatbot in Basecamp:
 [My App] [dhh] Rolled back to version d264c4e92470ad1bd18590f04466787262f605de
 ```
 
+In addition to the formatted message, MRSK sets a number of environment variables with the components of the broadcast.
+You can use these (rather than the command argument) if you want more control over how the message is formatted.
+MRSK currently sets:
+
+- `MRSK_PERFORMER` - the user performing the command
+- `MRSK_DESTINATION` - the destination
+- `MRSK_ROLE` - the specific role being targetted, if any
+- `MRSK_EVENT` - text of the action (e.g. "Deployed app@150b24f")
+
 ### Custom healthcheck
 
 MRSK defaults to checking the health of your application again `/up` on port 3000 up to 7 times. You can tailor the behaviour with the `healthcheck` setting:
--- a/README.md
+++ b/README.md
@@ -677,14 +677,20 @@ That'll post a line like follows to a preconfigured chatbot in Basecamp:
 [My App] [dhh] Rolled back to version d264c4e92470ad1bd18590f04466787262f605de
 ```
 
-In addition to the formatted message, MRSK sets a number of environment variables with the components of the broadcast.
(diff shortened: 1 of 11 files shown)


lib/mrsk/cli/base.rb
```diff
@@ -73,7 +73,9 @@ def print_runtime
       end
 
       def audit_broadcast(line)
-        run_locally { execute *MRSK.auditor.broadcast(line), verbosity: :debug }
+        if broadcast = MRSK.auditor.broadcast(line)
+          system(MRSK.auditor.broadcast_environment(line), broadcast)
```
kevinmcconnell, line 77: I don't think there's a way to set the environment when using `run_locally`, so I switched to `system` for this one instead.

README.md
```diff
@@ -677,6 +677,15 @@ That'll post a line like follows to a preconfigured chatbot in Basecamp:
...
 ```
 
+In addition to the formatted message, MRSK sets a number of environment variables with the components of the broadcast.
+You can use these (rather than the command argument) if you want more control over how the message is formatted.
+MRSK currently sets:
+
+- `MRSK_PERFORMER` - the user performing the command
+- `MRSK_DESTINATION` - the destination
+- `MRSK_ROLE` - the specific role being targetted, if any
+- `MRSK_EVENT` - text of the action (e.g. "Deployed app@150b24f")
```
jeremy, line 687: Perhaps `MRSK_MESSAGE` or similar? `MRSK_EVENT` could later become something more specific like `deployed`.

lib/mrsk/cli/base.rb
dhh: @kevinmcconnell Did you want to switch to this #with approach?

jeremy:
Additions via https://github.com/basecamp/mrsk/pull/1:
- Audit logs and broadcasts accept `details` whose values are included as log tags and MRSK_* env vars passed to the broadcast command
- Commands may return execution options to the CLI in their args list
- Introduce `mrsk broadcast` helper for sending audit broadcasts

kevinmcconnell:
@jeremy this branch no longer seems to be working for me. The env vars aren't set in my broadcast command. It's also creating some empty files when it runs, with names that match the values from those vars (in other words, I get files named things like  `staging,` and `kevin,`). Am I doing something wrong?

Simplest case I have to reproduce is to run a `mrsk broadcast` with an empty broadcast command script.

jeremy:
Oops! Will fix that up @kevinmcconnell.

kevinmcconnell:
Thanks for fixing that up, @jeremy!

I've also resolved a conflict that cropped up with latest `main`, and tested this out on a couple of deploys. I think it's good to go now.

PR #225, opened 2023-04-14
load environment variables from multiple .env files as dotenv-rails

xiaohui-zhangxh:
Almost all of our projects are loading environment variables with dotenv-rails, so we always commit `.env` file to Git, our developers clone repository and create `.env.local` to override database/redis settings

I think mrsk can follow the best practice of dotenv-rails, not only read `.env` file, but also another files in priority as: 

1. .env.`[destination]`.local
2. .env.`[destination]`
3. .env.deploy.local
4. .env.local
5. .env.deploy
6. .env

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,19 @@ MRSK_REGISTRY_PASSWORD=pw
 DB_PASSWORD=secret123
 ```
 
+environment variables will override in the following order (highest defined variable overrides lower):
+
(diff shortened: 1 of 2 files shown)


tbuehlmann:
We're using .env files for local development and we already switched from .env to .env.local because of mrsk. With this change I wouldn't know how to have .env files for local development anymore.

xiaohui-zhangxh:
close this, create issue first #233

PR #235, opened 2023-04-20
Add support for ssh forwarding to docker

dilpreet92:
* this will enable clone private gem repositories or in general forwarding ssh connection to docker

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,6 +585,22 @@ ARG RUBY_VERSION
 FROM ruby:$RUBY_VERSION-slim as base
 ```
 
+### Forward your existing SSH agent connection to docker
+
+Clone your private repositories during build.
+
+```yaml
+  builder:
+    ssh: $HOME/.ssh/name_of_your_ssh_key
+```
+
+Then, include the following in your Dockerfile:
+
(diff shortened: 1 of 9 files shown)


dhh:
What benefit do you see here over using BUNDLE_GITHUB__COM with a Docker secret, as described in the README? https://github.com/mrsked/mrsk/#using-build-secrets-for-new-images

dilpreet92:
> What benefit do you see here over using BUNDLE_GITHUB__COM with a Docker secret, as described in the README? https://github.com/mrsked/mrsk/#using-build-secrets-for-new-images

there might be some cases where private gems are being accessed using `ssh` and not `https:`. 
`gem 'abc', git: 'git@github.com:dilpreet92/xxx.git'`
In order for SSH to work with Bundler, the user executing bundle install needs to have an authorized private key for the Github repo, so a common pattern is to use a [machine user](https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users) on CI, that has read access to all the repos it needs access to.

djmb:
Support for this was added in https://github.com/basecamp/kamal/pull/434

PR #242, opened 2023-04-25
Add optional healthcheck skip

acidtib:
Optional setting `skip` can be useful for deploying bots (Discord bots, Twitch bots, regular jobs) or anything that does not require an incoming HTTP connection. The default is false.

```yaml
healthcheck:
  skip: true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -686,11 +686,14 @@ healthcheck:
   path: /healthz
   port: 4000
   max_attempts: 7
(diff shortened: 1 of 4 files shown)


tbuehlmann:
I can see tabs instead of spaces, can you convert them?

acidtib:
going to close this pull request, #219 is a better option :)

PR #244, opened 2023-04-25
Get lock status by executing directly

djmb:
Getting the lock status with invoke passes through any options from the original command which will raise an exception if they are not also valid for the lock status command.

Fixes https://github.com/mrsked/mrsk/issues/239


djmb:
@dhh - this failed on the integration test being flaky - https://github.com/mrsked/mrsk/pull/245 should fix that.

PR #250, opened 2023-04-28
Integration test healthcheck wget

djmb:
The alpine nginx container doesn't contain curl, so let's override the
healthcheck command to use wget.

### DIFF
--- a/lib/mrsk/utils/healthcheck_poller.rb
+++ b/lib/mrsk/utils/healthcheck_poller.rb
@@ -1,5 +1,5 @@
 class Mrsk::Utils::HealthcheckPoller
-  TRAEFIK_HEALTHY_DELAY = 1
+  TRAEFIK_HEALTHY_DELAY = 2
 
   class HealthcheckError < StandardError; end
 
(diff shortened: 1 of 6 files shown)


lib/mrsk/utils/healthcheck_poller.rb
```diff
@@ -1,5 +1,5 @@
 class Mrsk::Utils::HealthcheckPoller
-  TRAEFIK_HEALTHY_DELAY = 1
+  TRAEFIK_HEALTHY_DELAY = 2
```
djmb, line 2: @kevinmcconnell - the integration tests were consistently failing in Github actions until I upped this to 2 seconds. Traefik was returning a 404 when we checked it the app was up.

Worked fine with 1 second on my M1, so probably a resource constraint issue - I think its worth sticking with a slightly longer default delay, but maybe this should be configurable.

kevinmcconnell: Thanks, and I agree, it would be worth making this configurable.

Issue #231, opened 2023-04-19
escape_shell_value doesn't support Chinese value of environment variable


xiaohui-zhangxh:
hi @dhh, we have some Chinese environment variable values placed in .env, such as text message template. When I deploy my app with mrsk, I noticed Chinese string becomes to unreadable chars, then passing this value to docker env and load to Rails, Rails can't get the same string as value in .env

Dig into code, I found  `escape_shell_value` calls `String#dump`, I don't know if generated quotes `"` by `String#dump` are important, can we escape to safe shell with [`Shellwords.escape` ](https://ruby-doc.org/3.2.1/stdlibs/shellwords/Shellwords.html) ,  it seems Shellwords  should return safe string as its design.

```ruby
  def escape_shell_value(value)
    value.to_s.dump
      .gsub(/`/, '\\\\`')
      .gsub(DOLLAR_SIGN_WITHOUT_SHELL_EXPANSION_REGEX, '\$')
  end
```

```ruby
╰─⠠⠵ irb -r'mrsk'
irb(main):001:0> Mrsk::Utils.escape_shell_value("中国")
=> "\"\\u4E2D\\u56FD\""
irb(main):002:0> require 'shellwords'
=> false
irb(main):003:0> Shellwords.escape('中国')
=> "\\中\\国"
irb(main):004:0> 
```

xiaohui-zhangxh:
I have made some change to this function here https://github.com/mrsked/mrsk/compare/main...xiaohui-zhangxh:feature/escape-shell-value-with-shellwords

After switched to Shellwords there are many tests are failed due to the quotes(`"`) issue, we have to fix those tests.

dhh:
Do open a PR 👍

xiaohui-zhangxh:
> Do open a PR 👍

done

xiaohui-zhangxh:
1.0 has been changed to pushing env to server, so there is no issue with shell args, closing this.

PR #260, opened 2023-05-02
Escape shell value with Shellwords instead of String#dump

xiaohui-zhangxh:
More detailed info, please see issue https://github.com/mrsked/mrsk/issues/231

This PR has fixed all tests


xiaohui-zhangxh:
since issue #231 is closed, this should be closed too.

PR #282, opened 2023-05-09
Update README.md

jasonfb:
add clarification about `RAILS_MASTER_KEY`, whether it should be in the `.env` file, and differences between Rails 7.0/Rails 7.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,26 @@ env:
     - RAILS_MASTER_KEY
 ```
 
-Then edit your `.env` file to add your registry password as `MRSK_REGISTRY_PASSWORD` (and your `RAILS_MASTER_KEY` for production with a Rails app).
+Edit your `.env` file to add your registry password as `MRSK_REGISTRY_PASSWORD` (unless you are committing the `.env` file to source control, in which case you'll want to pass it in as builder arg, see below). 
+
+The `RAILS_MASTER_KEY` might be needed during asset precompile, which happens on the build container (not the running app). 
+
+If migrating a legacy Rails app that requires env settings during precompile, then you'll need to make your `RAILS_MASTER_KEY` available to the deploy process. However, this is typcially bad practice and you probably should remove depenencies around requiring env settings during pre-compile. 
+
+In Rails 7.0, you'll need to make the `RAILS_MASTER_KEY` available to the precompile step. Starting with Rails 7.1, you can use the new `SECRET_KEY_BASE_DUMMY=1` setting to bypass the need for the `RAILS_MASTER_KEY` during precompile.
+
+If your precompile step does need `RAILS_MASTER_KEY` or you are on Rails 7.0, you have two choices:
+
+1) Add your `RAILS_MASTER_KEY` for production to `.env`
+2) Alternatively, you can pass the RAILS_MASTER_KEY in as a builder arg in your Dockerfile (be sure to do this before the asset:precompile line in your Dockerfile), like so:
+
+```
+ARG RAILS_MASTER_KEY
+ENV RAILS_MASTER_KEY=$RAILS_MASTER_KEY
+```
+You should not put your `RAILS_MASTER_KEY` into your `.env` file if the `.env` file is checked-in to your source control, but if that file is not checked into source control, you can. 
+
+For guideance about whether or not to commit your `.env` file, (https://github.com/bkeepers/dotenv#should-i-commit-my-env-file)[see this].
 
 Now you're ready to deploy to the servers:
 


jasonfb:
@tbuehlmann says on discord (please correct me if I've misrepresented you) you can do a secret base key one of four ways, and this README change does not explain the nuance:

1) a SECRET_KEY_BASE enrivonment variable
2) a secret_key_base credentials
3) a secret from config/secrets
4) automatically if Rails.env.local? or ENV["SECRET_KEY_BASE_DUMMY"] is set

further, he says: 
"just wanted to point out that SECRET_KEY_BASE_DUMMY doesn't help with the credentials issue"

jasonfb:
@tbuehlmann also says:

> one other thing. this is suggested in the PR:

2) Alternatively, you can pass the RAILS_MASTER_KEY in as a builder arg in your Dockerfile (be sure to do this before the asset:precompile line in your Dockerfile), like so:

ARG RAILS_MASTER_KEY
ENV RAILS_MASTER_KEY=$RAILS_MASTER_KEY

> this is a bad idea as you'd expose your master key.

jasonfb:
thank you @tbuehlmann for the feedback I think we should keep investigating because I think it's important to provide good guidance here, especially because we don't want to provide insecure suggestions in this critical setup area.

jasonfb:
I'm not totally (yet) convinced that the secret is in the tag. If it is, I agree this is not a good solution.

When I pull my latest tag (which was created by mrsk) from dockerhub to my local machine, examine it in Docker and run it, I get this :

```
2023-05-09 09:55:15 /usr/local/bundle/ruby/3.1.0/gems/railties-7.0.4.3/lib/rails/application.rb:581:in `validate_secret_key_base': Missing `secret_key_base` for 'production' environment, set this string with `bin/rails credentials:edit` (ArgumentError)
```

so if it's true that the secret is somewhere in the tag, how do I find it? it appears to me that when I pull the tag locally and run the image, Rails can't boot without the secret base key, suggesting to me that the key isn't in the tag. (although I could be confused about the difference between the image and the mrsk deploy container here) 

On the "Inspect" tab in Docker, I see no secret base key or RAILS_MASTER_KEY 

On the "Files" tab, I can examine the files of the image but here also I am unable to find the RAILS_MASTER_KEY in any file (the real place it lives --- `config/master.key` is not checked in so it isn't here in the image)

I'm basically just trying to confirm if it is true or not that "this is a bad idea as you'd expose your master key" — meaning if the master key gets into the Docker image itself. (if true, then I agree with @tbuehlmann this is a bad idea and shouldn't be recommended)

tbuehlmann:
I think things are being a little mixed up here.

So I'm reading about `RAILS_MASTER_KEY`, but not credentials. Why's that? And I'm reading about `SECRET_BASE_KEY_DUMMY` which isn't really connected to the issue at hand (that decrypting credentials fails because of a missing key (which is **not** the secret key base)).

I'm also not sure if we should mention all of this in the first place because it's really tied to Rails and the issue, to me, is out of scope of mrsk. You would have stumbled over this with capistrano and other tools as well. But I'm undecided here.

A smallish example how your secret (in the form of an environment variable) would end up in the image:

```
# Dockerfile
FROM alpine

ARG FOO
ENV FOO=$FOO
```

```sh
# build
$ docker build --build-arg FOO=some-value -t some-image .

# run
$ docker run --rm some-image env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=9e207cdf6bc2
FOO=some-value
HOME=/root
```

jasonfb:
@tbuehlmann -- your points are well taken.
let me clarify the goal:
The goal is to provide guidance on how to make the precompile work within the mrsk deploy container. 

If you're on Rails 7.0, there's no option for SECRET_KEY_BASE_DUMMY, so AFAIK, you'll need to provide the secret_key_base _in some manner_. (As you've said, that can be different ways). 

The Rails defaults are for using encrypted credentials and putting them into the encrypted credentials. To do that, the precompile will need access to the RAILS_MASTER_KEY. 

it's not just about what "connected to the issue at hand," it's about providing guidance for people who get tripped up in this area. 

So... for Rails 7.1, just use `SECRET_KEY_BASE_DUMMY`. easy. 

But for Rails 7.0, what's the solution you'd propose to the problem? 

The original problem is described here: https://stackoverflow.com/questions/76157925/rails-7-validate-secret-key-base-missing-secret-key-base-for-production-e/76190212

if you have a solution other than what's on that SO post, please suggest your solution. I will modify this PR accordingly based on new information, but as of yet, I don't see an alternative solution.

if the MRSK README is supposed to be Rails-less, maybe an alternative or supplemental README can be added that covers just this stuff for the MRSK-Rails setup.

nickhammond:
I know this is closed but I stumbled on after searching through the repository. @jasonfb to get this going with 7.0 you can utilize build secrets to get what you're looking for. There's a section in the readme about installing private gems, it's a similar process to get precompile working if you need the master key. Build secrets with mrsk utilize docker secrets under the hood so you don't end up leaving your secrets behind.

https://github.com/mrsked/mrsk#using-build-secrets-for-new-images

PR #285, opened 2023-05-10
Clean up traefik boot failure

### LINKED ISSUE #217: suggestion: failed traefik boot clean up
My traefik container originally failed to start due to a port conflict on port 80. However this leaves a container lying around called `traefik` that then causes subsequent boots to fail (even though I had fixed the port conflict).

Perhaps failed boots of traefik should be cleaned up? Or is there some state issue I'm missing?

Jberczel:
If user not deploying to a freshly provisioned server, there may be cases where traefik fails to initially boot.

For example, if there's an exisitng container binded to port 80, `mrsk traefik boot`  will create but **not** run a new traefik container, which will disrupt successive traefik boots (container name already exists).

This commit aims to clean up the initial traefik boot. If it cannot, it'll raise an exception for user to investigate and fix their docker environment.

Afterwards, they should be able to deploy as usual, and mrsk will take care of cleaning up any orphaned traefik container.

Fixes #217

### DIFF
--- a/lib/mrsk/cli/traefik.rb
+++ b/lib/mrsk/cli/traefik.rb
@@ -2,9 +2,23 @@ class Mrsk::Cli::Traefik < Mrsk::Cli::Base
   desc "boot", "Boot Traefik on servers"
   def boot
     with_lock do
+      cli = self
       on(MRSK.traefik_hosts) do
         execute *MRSK.registry.login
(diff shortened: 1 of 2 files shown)


djmb:
Should be covered by https://github.com/mrsked/mrsk/pull/414

PR #290, opened 2023-05-11
add healthcheck interval option

acidtib:
The current health check happens every `1s`, this PR introduces the `interval` option to customize docker `--health-interval`

```sh
healthcheck:
  path: /up
  port: 3000
  interval: 20s
```


obie:
Hallelujah, this is seriously needed.

PR #291, opened 2023-05-11
MRSK Hooks

djmb:
Adds hooks to MRSK. Currently just two hooks, pre-build and post-push.

We could break the build and push into two separate commands if we found the need for post-build and/or pre-push hooks.

Hooks are stored in `.mrsk/hooks`. Running `mrsk init` will now create that folder and add sample hook scripts.

Hooks returning non-zero exit codes will abort the current command.

Further potential work here:
- We could replace the audit broadcast command with a post-deploy/post-rollback hook or similar
- Maybe provide pre-command/post-command hooks that run after every mrsk invocation
- Also look for hooks in `~/.mrsk/hooks`


olivermt:
How would this work with running mrsk with a -c path/to/deploy.yml ? 
The hook paths would need to be configurable in the yml file itself for this to work with custom config paths I think.

dhh:
Agree it'd be nice to have the commands configurable. But also curious to explore the two types. What does the post-push one give us that pre-build doesn't? Seems like it just has the clean check, which we already use on build? Maybe pre-build is the only one we actually need?

djmb:
A second clean check confirms that nothing a snuck in between starting and finishing the build.

`post-push` might also be useful for running a security scan of the image, though strictly you might want to do that `post-build`, which would mean splitting build and push into two commands. 

Other hooks I think we'll want are:
1. pre-command - to be run before any command is run (for us needed to pre-warm DNS). Or if we want to be more specific we could have `pre-connect` that runs just before the first time a command tries to ssh.
2. post-deploy and post-rollback - to replace `audit_broadcast_cmd`.

dhh:
Hmm, the post push seems a bit speculative to me at this point. But I very much do like the idea of replacing the audit_broadcast_cmd with post-deploy and post-rollback. In fact, I think it'd probably make sense to roll that into this release. Then we'll at least have 3 solid hooks that we know we have use cases for. And fine to just drop the `audit_broadcast_cmd` and note in the release notes that it needs to be a hook going forward instead.

dhh:
Seems the CI tests are a bit unreliable.

djmb:
The CI test failure was caused by:
```
2023-05-22T12:06:13.8274060Z Error response from daemon: Get "https://registry-1.docker.io/v2/": net/http: TLS handshake timeout
2023-05-22T12:06:13.8416624Z Error response from daemon: No such image: nginx:1-alpine-slim
```

Maybe not surprising that we see issues like this with docker running in docker running in a Github action. I'll set --max-concurrent-downloads to 1 on the docker daemon to see if that helps.

dhh:
That worked!

dhh:
I was just looking at that, and I think we can drop it, @djmb. Don't think that broadcast is actually necessary/useful. It's being tracked by the auditor. That's enough.

lib/mrsk/cli/main.rb
```diff
@@ -76,49 +76,54 @@ def redeploy
...
-          say "The app version '#{version}' is not available as a container (use 'mrsk app containers' for available versions)", :red
+            rolled_back = true
+          else
+            say "The app version '#{version}' is not available as a container (use 'mrsk app containers' for available versions)", :red
+          end
         end
       end
     end
+
+    run_hook "post-rollback", runtime: runtime.round if rolled_back
```
dhh, line 126: I think we can actually just let this be post-deploy as well. Don't think we need a special case for rollback. The outcome is essentially the same: New version is now running.

dhh:
Merged your other cleanup PR. Can you rebase?

dhh:
So seems like we're missing skip broadcast now. Should we just add a generic "skip hooks" option as an alternative?

PR #294, opened 2023-05-13
Add README section to enable Traefik dashboard

duyanton:


tannakartikey:
I don't think we add a whole new section and example for this again. Just add the `api.dashboard` line under the "Customize the Traefik command line using args:" section

PR #310, opened 2023-05-24
Add a pre-lock hook

djmb:
This can be used for hooks that should run before connecting to remote hosts. An example use case is pre-warming DNS.


README.md
```diff
@@ -883,14 +883,18 @@ firing a JSON webhook. These variables include:
 - `MRSK_MESSAGE` - the full audit message, e.g. "Deployed app@150b24f"
 - `MRSK_SERVICE_VERSION` - an abbreviated version (for use in messages)
```
northeastprince, line 884: @djmb why does this say `MRSK_SERVICE_VERSION` when the example below uses `MRSK_VERSION`?

djmb: Ah both are available, I'll update the README to reflect that 🙏

djmb:
Closed in favour of https://github.com/mrsked/mrsk/pull/311

PR #323, opened 2023-05-31
Prefix container hostname with the underlying one

djmb:
To make it easier to identity where a docker container is running, prefix its hostname with the underlying one from the host.

Docker chooses a 12 character random hex string by default, so we'll keep that as the suffix.


defp:
How to turn off this rule

PR #348, opened 2023-06-22
Allow to publish multiple ports for accessories

bohdan-shulha:
Closes https://github.com/mrsked/mrsk/issues/347

I'm not a ruby expert, but, luckily, there are 'host' and 'hosts' directives already, so I just followed the approach.

### DIFF
--- a/lib/mrsk/configuration/accessory.rb
+++ b/lib/mrsk/configuration/accessory.rb
@@ -179,13 +179,13 @@ class Mrsk::Configuration::Accessory
     end
 
     def ports_from_ports
-      if specifics.key?("ports")
-        ports = specifics["ports"]
-        if ports.is_a?(Array)
-          ports.map do |port| format_port(port) end
-        else
-          raise ArgumentError, "Ports should be an Array for accessory `#{name}`"
-        end
+      return nil unless specifics.key?("ports")
+      ports = specifics["ports"]
+
+      unless ports.is_a?(Array)
+        raise ArgumentError, "Ports should be an Array for accessory `#{name}`"
       end
+
+      ports.map { |port| format_port(port) }
     end
 end
--- a/lib/mrsk/configuration/accessory.rb
+++ b/lib/mrsk/configuration/accessory.rb
@@ -23,14 +23,17 @@ class Mrsk::Configuration::Accessory
     hosts_from_host || hosts_from_hosts || hosts_from_roles
   end
 
-  def port
-    if port = specifics["port"]&.to_s
-      port.include?(":") ? port : "#{port}:#{port}"
+  def ports
+    if specifics["port"] && specifics["ports"]
+      raise ArgumentError, "Specify one of `port` or `ports` for accessory `#{name}`"
     end
+
+    return ports_from_port if specifics["port"]
+    return ports_from_ports
   end
 
   def publish_args
-    argumentize "--publish", port if port
+    argumentize "--publish", ports if ports
   end
 
   def labels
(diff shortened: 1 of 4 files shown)


lib/mrsk/configuration/accessory.rb
```diff
@@ -23,14 +23,20 @@ def hosts
...
+  def ports
+    if (specifics.keys & ["port", "ports"]).size > 1
+      raise ArgumentError, "Specify one of `port` or `ports` for accessory `#{name}`"
+    end
+
+    (ports_from_port || ports_from_ports)&.map do |p|
+      port = p.to_s
+
       port.include?(":") ? port : "#{port}:#{port}"
     end
```
Zeko369, line 35: 1st, nice, first time seeing the `array & array` syntax and I most admit I love it

But, I think that in this case `specifics['port'] && specifics['ports']` would be easier to read (just IMO?), same goes for the map leter?

```suggestion
    if specifics["port"] && specifics["ports"]
      raise ArgumentError, "Specify one of `port` or `ports` for accessory `#{name}`"
    end
    
    return ports_from_port if specifics["port"]
    return ports_from_ports if specifics["ports"]
```

And then later do conversion to `num:num`
Maybe even make the second return implicit and remove the if

Zeko369 (approved):
- Left a stylist change comment
- What do you think about adding another test to test if `num:num` is also mapped correctly (I know it works but for the sake of thoroughness)?

PS: Replace `Covers` with `Closes` in the PR description and github will autolink (and then close) this PR to the issue

lib/mrsk/configuration/accessory.rb
bohdan-shulha: Thanks for review.

Didn't get this part though:

> And then later do conversion to num:num


By saying "later" you mean the callsite of `#ports`?

bohdan-shulha: > What do you think about adding another test to test if num:num is also mapped correctly (I know it works but for the sake of thoroughness)?

Not sure whether it is needed, this case is implicitly covered multiple times. The only thing that I didn't see - tests for the format correctness (i.e. to prevent typos in config like `8080: 8080` - with space or other letter).

Zeko369: > By saying "later" you mean the callsite of #ports?

Yeah I meant to instead of doing ports_from_port and ports_from_ports and then formatting here, to do the formatting inside of those methods (or wrap here) but not to the map since it looks like more things are happening

> Multiple tests

Yeah true, my thoughts were more we test for `port: 1234:4321 / 4123` but only `ports: 1234`, but it's fine since we cover it by the other one

bohdan-shulha: Hi @Zeko369 ,

I've applied your change and proposal. Could you please take a look?

lib/mrsk/configuration/accessory.rb
```diff
@@ -166,4 +169,23 @@ def hosts_from_roles
...
+
+    def ports_from_ports
+      if specifics.key?("ports")
+        ports = specifics["ports"]
+        if ports.is_a?(Array)
+          ports.map do |port| format_port(port) end
+        else
+          raise ArgumentError, "Ports should be an Array for accessory `#{name}`"
+        end
+      end
```
Zeko369, line 189:
```suggestion
      return nil unless specifics.key?("ports")
      ports = specifics["ports"]

      unless ports.is_a?(Array)
        raise ArgumentError, "Ports should be an Array for accessory `#{name}`"
      end

      ports.map { |port| format_port(port) }
```

Zeko369 (approved):
Just a control flow nitpick refactor idea, but 🥳

Zeko369 (approved):

lewispb:
Thanks for your contribution @bohdan-shulha - it'll be useful to publish multiple ports for an accessory.

I disagree with the style changes proposed by @Zeko369 - I think it would be better to maintian the existing code style, and keep the implementation in the same vein as the existing #hosts_from_host and #hosts_from_hosts.

Otherwise, this looks good to me.

bohdan-shulha:
Hi @lewispb ,

Thanks for review. ;)

Would it be enough to revert to this commit then?

https://github.com/mrsked/mrsk/pull/348/commits/f6f34dea7914bde7e1ee4633656ba06cf839eea5

PR #352, opened 2023-06-22
feat: Add template argument to envify

Zeko369:
Closes #351

- Adds `-t` argument to `envify`
- injects `MRSK_DESTINATION` so a single template can be used to generate env files for multiple destinations
- injects `MRSK_DESTINATION` when loading `deploy.yml` files

### TODO
- [ ] Docs

PR #353, opened 2023-06-23
feat: Add --env-path specify env path

Zeko369:
Continuation of #352 with by making it so you can also load/output .env files for a different path (i.e. in out case we'd end up with 35+ top level .env files)

### TODO
- [ ] Docs

### DIFF
--- a/lib/mrsk/cli/base.rb
+++ b/lib/mrsk/cli/base.rb
@@ -31,7 +31,7 @@ module Mrsk::Cli
 
     private
       def load_envs
-        if destination = options[:destination]
+        if (destination = options[:destination])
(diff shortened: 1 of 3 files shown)


snopoke:
+1 this would great and also solves the case where the project already uses a `.env` file for local dev.

jdelStrother:
Anything we can do that would help get this merged? I see it has conflicts - if I fixed them would there be any interest in merging it?

igor-alexandrov:
@djmb this is not relevant anymore with Kamal 2.

djmb:
Secrets have moved to `.kamal/secrets` in Kamal 2

PR #360, opened 2023-06-27
feat: Support optional env secrets

### LINKED ISSUE #359: Support optional secrets
(Continuation of #346)

My idea was to somehow override them in the destination files but since they're merged there needs to be a "way" of allowing optional ENVs (without having to do `[empty]` and then check in the app)

### My idea of how this should work

```yml
# deploy.yml
env:
  secret:
    - APP_NAME
    - RETURN_URL?
```

### Cases
1. `APP_NAME` missing -> errors
2. `RETURN_URL` present -> loaded
3. `RETURN_URL` missing -> ignored (so using `ENV['RETURN_URL']` in container would return `nil`)

Zeko369:
Closes #359

This PR allows specifying optional env secrets with `- FOO?` syntax

### Example
In the following example the app won't build without `RAILS_MASTER_KEY` env variable being provided but will work both with and without FOO?

```yml
env:
  secrets:
    - RAILS_MASTER_KEY
    - FOO?
```

### Todos
- [ ] Think about overriding (i.e. if we have `FOO` in `deploy.yml` and `FOO?` in `deploy.stage.yml`, what's the output) [comment here](https://github.com/mrsked/mrsk/issues/359#issuecomment-1609386638)
- [ ] Finish specs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 coverage/*
 .DS_Store
 gemfiles/*.lock
+.vscode/
+.idea/
--- a/lib/mrsk/configuration/role.rb
+++ b/lib/mrsk/configuration/role.rb
@@ -139,14 +139,25 @@ class Mrsk::Configuration::Role
     # Secrets are stored in an array, which won't merge by default, so have to do it by hand.
     def merged_env_with_secrets
       merged_env.tap do |new_env|
-        new_env["secret"] = Array(config.env["secret"]) + Array(specialized_env["secret"])
-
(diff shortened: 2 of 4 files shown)


xiaohui-zhangxh:
Good idea, now I have to write them like this:

```yaml
env:
  secrets:
    - RAILS_MASTER_KEY
  <% if ENV['FOO'].to_s != '' %>
    - FOO
  <% end %>
```

djmb:
Going to close this one as its very out of date now. Not sure we'd want to go down this route anyway

PR #364, opened 2023-06-28
Update README.md

nickhammond:
Add a note about utilizing biometrics with the envify example.


Zeko369 (approved):
Damn, had the same problem the other day 🚀

PR #368, opened 2023-06-29
Allow Traefik to run without publishing port

tsvallender:
Adds the `publish` option which, if set to false, does not pass `--publish` to `docker run` when starting Traefik. This is useful when running Traefik behind a reverse proxy, for example.

Example deploy.yml:
```
traefik:
  publish: false
```

If not specified or set to true, the default or specified port will be used as per current behaviour.

### DIFF
--- a/lib/kamal/commands/traefik.rb
+++ b/lib/kamal/commands/traefik.rb
@@ -11,7 +11,7 @@ class Kamal::Commands::Traefik < Kamal::Commands::Base
     docker :run, "--name traefik",
       "--detach",
       "--restart", "unless-stopped",
-      "--publish", port,
+      *publish_args,
       "--volume", "/var/run/docker.sock:/var/run/docker.sock",
       *env_args,
       *config.logging_args,
@@ -64,6 +64,10 @@ class Kamal::Commands::Traefik < Kamal::Commands::Base
   end
 
   private
+    def publish_args
+      argumentize "--publish", port unless config.traefik["publish"] == false
+    end
+
     def label_args
       argumentize "--label", labels
     end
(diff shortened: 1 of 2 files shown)


Zeko369 (commented):
Could you add a few lines in the docs about?

1: that this exists now
2: how to use it with an additional traefik/nginx rev proxy (or just a link to a guide) and a usecase (i.e. "if you want to host 2 apps on a same machine use this + a separate traefik/nginx")

tsvallender:
> Could you add a few lines in the docs about?
> 
> 1: that this exists now 2: how to use it with an additional traefik/nginx rev proxy (or just a link to a guide) and a usecase (i.e. "if you want to host 2 apps on a same machine use this + a separate traefik/nginx")

Sure, in this case are the docs the README? Happy to add it to the site repo too but I assumed that would happen after this was merged (or at least approved)?

lib/mrsk/commands/traefik.rb
```diff
@@ -60,6 +60,12 @@ def port
   end
 
   private
+    def publish_args
+      return if config.traefik["publish"] == false
```
dhh, line 64: Let's just turn this into a proper conditional:

```ruby
argumentize("--publish", port) unless config.traefik["publish"] == false
```

tsvallender: Sure, that's definitely cleaner :+1:

defsdoor (approved):

dhh:
Can you rebase? Looks good to me.

dhh:
Doc change needs to happen against basecamp/kamal-site repo.

tsvallender:
I've rebased and added documentation to the site repo https://github.com/basecamp/kamal-site/pull/17

defsdoor (approved):
Good work and a nice option - we run multiple application containers on different docker networks behind a broader proxy and have no need to expose to the host - this avoids port shuffling.

PR #380, opened 2023-07-09
Cleanup envify output

Zeko369:
Many editors (i.e. RubyMine) format erb with indents, causing `envify` to not output the nicest env files (output below, even though it technically works)

This PR resolves that by removing all leading whitespace + reming multiline newlines (if you have more than 1 blank line it's reduced to 1)

```erb
<% if 1 > 2 %>
  HELLO_WORLD=<%= op read ... <%>
<% end %>
```

outputs
```dotenv
  HELLO_WORLD=foobar
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 coverage/*
 .DS_Store
 gemfiles/*.lock
+.idea/
--- a/lib/mrsk/cli/main.rb
+++ b/lib/mrsk/cli/main.rb
@@ -174,7 +174,10 @@ class Mrsk::Cli::Main < Mrsk::Cli::Base
       env_path          = ".env"
     end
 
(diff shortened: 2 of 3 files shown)


djmb:
Multiline env variables are valid, so this would be a problem if they had indentation that was to be retained.

e.g.

```
INDENTED_MESSAGE="  Hello
  This is an indented message"
```

One option though would be to use set the [trim_mode](https://ruby-doc.org/3.2.2/stdlibs/erb/ERB.html#method-c-new) on ERB.new to "-", so you can use something like:

```
<% if true -%>
  ENV=123
<% end -%>
```

That will trim newlines from `<% if true -%>` and `<% end -%>` lines but not fix the indentation.

PR #389, opened 2023-07-15
Include role options when executing commands

### LINKED ISSUE #388: Executing commands on a new container does not start the container with the options defined in the configuration
I have a configuration like this in my web role to access to the database server, which is not running on the same network as docker, but on the same host.

```yml
servers:
  web:
    hosts:
      - 1.1.1.1
    options:
      add-host: host.docker.internal:host-gateway
```

So, the host becomes host.docker.internal.

When trying to run commands with `mrsk app exec`, the new container that starts does not include these options, so the host is invalid and the application cannot connect to the database.

brunoprietog:
Fix #388

### DIFF
--- a/lib/mrsk/commands/app.rb
+++ b/lib/mrsk/commands/app.rb
@@ -76,11 +76,14 @@ class Mrsk::Commands::App < Mrsk::Commands::Base
   end
 
   def execute_in_new_container(*command, interactive: false)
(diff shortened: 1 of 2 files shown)


dhh:
Can you add a test for this?

brunoprietog:
Done

PR #391, opened 2023-07-17
Automatically detect yml.erb

Zeko369:
Although you could handle this already by renaming the `.yml.erb` file to `.erb.yml` and then expecting mrsk to replace `.yml` with `.[destination].yml` for editor compatibility (and keeping up with how rails expects files) files should end in .erb for them to be detected as such

This PR makes it so that `.yml` and `.yml.erb` and supported as extensions and replaced correctly


Zeko369:
Hey @djmb, I know this didn't get merged since it's on top of the old MRSK codebase... but would you be down for merging this?

We have it our fork but we'd like to simplify stuff and upstream if possible?

PR #393, opened 2023-07-19
Support a --rolling option for traefik reboots

lewispb:
Rolling restarts allow Traefik configuration changes without downtime with:

`mrsk traefik reboot --rolling`

### DIFF
--- a/lib/mrsk/cli/traefik.rb
+++ b/lib/mrsk/cli/traefik.rb
@@ -9,12 +9,17 @@ class Mrsk::Cli::Traefik < Mrsk::Cli::Base
     end
   end
 
+  method_option :rolling, type: :boolean, default: false
   desc "reboot", "Reboot Traefik on servers (stop container, remove container, start new container)"
   def reboot
     mutating do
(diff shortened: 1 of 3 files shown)


lib/mrsk/cli/traefik.rb
```diff
@@ -1,24 +1,25 @@
...
     mutating do
       on(MRSK.traefik_hosts) do
-        execute *MRSK.registry.login if login
+        execute *MRSK.registry.login
         execute *MRSK.traefik.run, raise_on_non_zero_exit: false
       end
     end
   end
 
+  method_option :rolling, type: :boolean, default: false
```
djmb, line 12: We use `option` which is just an alias for `method_option` everywhere else - could you update for consistency?

djmb (changes requested):
Cheers @lewispb! Could you also update the readme?

lewispb:
Thanks for the review @djmb. Readme updated, and option documented.

djmb (approved):

PR #394, opened 2023-07-20
WIP: Support for depot.dev builders

navinpeiris:
This PR adds support for [Depot] (https://depot.dev/) builders.

This is a WIP PR to get feedback if the approach makes sense, and will refine this PR accordingly. This works 

## Examples:

1. Basic configuration. Builds for the same architecture as the local machine:

    ```yaml
    builder:
      depot: true
    ```

2. Specifying one or more platforms to build for:

    ```yaml
    builder:
      depot:
        platforms:
          - linux/amd64
          - linux/arm64
    ```

### DIFF
--- a/lib/mrsk/commands/builder.rb
+++ b/lib/mrsk/commands/builder.rb
@@ -7,6 +7,8 @@ class Mrsk::Commands::Builder < Mrsk::Commands::Base
 
   def target
     case
+    when config.builder.depot?
+      depot
     when !config.builder.multiarch? && !config.builder.cached?
       native
     when !config.builder.multiarch? && config.builder.cached?
@@ -18,6 +20,10 @@ class Mrsk::Commands::Builder < Mrsk::Commands::Base
     else
       multiarch
     end
+  end
+
+  def depot
+    @depot ||= Mrsk::Commands::Builder::Depot.new(config)
   end
 
   def native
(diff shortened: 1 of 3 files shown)


dhh:
Don't see us adding a specific integration for a commercial service straight into MRSK. But maybe we can explore a plugin setup where something like this could be added separately.

navinpeiris:
Thanks @dhh, makes perfect sense! Won't attempt to explore a plugin system just yet, but will revisit this once one is in place or I've got more familiarity with mrsk. Closing this PR until such a time.

In the meantime for anyone wanting to use depot builder, will keep the `main` branch of https://github.com/navinpeiris/mrsk-depot in sync with this repository and with instructions on how you can use that.

jacobwgillespie:
Hey, I'm one of the Depot cofounders 👋 A plugin setup sounds cool, for a more native integration. That said, we've added the ability to use Depot as a Docker CLI plugin, so that it handles `docker build` and `docker buildx build` without calling the `depot` CLI directly. It's still early, but I think that might just work today given how MRSK is calling the Docker CLI.

I still need to add some docs, but if you want to try it, it's:

```
depot configure-docker --shim-buildx
```

That installs the plugin and sets it as the default Docker builder for `docker build`. `--shim-buildx` is optional, to make it also handle `docker buildx build`, which I believe MRSK is using when building for `multiarch`.

PR #399, opened 2023-07-25
Manage SSH connection starts

djmb:
Creating SSH connections concurrently can be an issue when deploying to many servers.

To mitigate this:
- limit concurrent connection starts to 30 at a time.
- set a long idle timeout of 900 seconds on connections to prevent re-connection storms after a long idle period, like building an image or waiting for CI

Both settings are configurable:

```yaml
sshkit:
  max_concurrent_starts: 10
  pool_idle_timeout: 300
```

### DIFF
--- a/lib/mrsk/configuration.rb
+++ b/lib/mrsk/configuration.rb
@@ -153,13 +153,22 @@ class Mrsk::Configuration
   end
 
   def ssh_options
-    { user: ssh_user, proxy: ssh_proxy, auth_methods: [ "publickey" ] }.compact
+    { user: ssh_user, proxy: ssh_proxy, auth_methods: [ "publickey" ], keepalive: true, keepalive_interval: 30 }.compact
   end
 
 
   def sshkit_max_concurrent_starts
     raw_config.sshkit["max_concurrent_starts"] if raw_config.sshkit.present?
   end
+
+  def sshkit_pool_idle_timeout
+    if raw_config.sshkit.present?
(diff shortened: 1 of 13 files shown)


lib/mrsk/configuration.rb
```diff
@@ -153,7 +153,20 @@ def ssh_proxy
...
+  def sshkit_max_concurrent_starts
+    raw_config.sshkit["max_concurrent_starts"] if raw_config.sshkit.present?
+  end
+
+  def sshkit_pool_idle_timeout
+    if raw_config.sshkit.present?
+      raw_config.sshkit["pool_idle_timeout"] || 900
+    else
+      900
+    end
```
igor-alexandrov, line 169: Since there are already a bunch of different SSH configuration methods, I think it would be better to extract them to a separate class similar to `Mrsk::Configuration::Accessory`, `Mrsk::Configuration::Builder` and others.

igor-alexandrov (commented):
I never used MRSK with more than 3 SSH connections, but this looks really reasonable.

lib/mrsk/configuration.rb
djmb: I've extracted ssh and sshkit configuration classes here - https://github.com/mrsked/mrsk/pull/399/commits/94d6a763a8895c3d5250470e92d4223d7fecab0e

igor-alexandrov: Looks much cleaner now.

PR #409, opened 2023-08-03
Kamal::Git.version now uses git describe

hey-leon:
## Why change Kamal::Git.version?
Currently container names can get very long with - service, role, destination, sha and even possible uncommited changes. This can make them hard to read or compare.

## Looking for feedback before finalising this PR
The tests are not passing but I am wondering if there is interest in this change or maybe an alternative one that would simply shorten the sha - e.g. `git rev-parse --short HEAD`

## Description
Previously Kamal::Git.version used rev-parse to retrieve the current commits full sha. While this works it currently creates a very long version name. This PR changes it to use `git describe`, which will use a tag if it exists, suffix with an abberviated commit sha if there are new commits, and lastly suffix it the same as before if their are uncommited changes. Some e.g.

- No tags, no uncommited changes: '12er34d5'
- No tags, uncommited-changes: '12er34d5_uncommited_23494aa9'
- Tags, no new commits, no uncommited changes: '1.2.3'
- Tags, new commits, no uncommited changes: '1.2.3-12er34d5'
- Tags, new commits, uncommited changes: '1.2.3-12er34d5_uncommited_23494aa9'

Further reading: https://git-scm.com/docs/git-describe

### DIFF
--- a/lib/kamal/git.rb
+++ b/lib/kamal/git.rb
@@ -10,7 +10,7 @@ module Kamal::Git
   end
 
   def revision
-    `git rev-parse HEAD`.strip
+    `git describe --always`.strip
   end
 
   def uncommitted_changes


lib/mrsk/configuration.rb
```diff
@@ -252,10 +252,8 @@ def role_names
 
     def git_version
       @git_version ||=
-        if system("git rev-parse")
-          uncommitted_suffix = Mrsk::Utils.uncommitted_changes.present? ? "_uncommitted_#{SecureRandom.hex(8)}" : ""
-
-          "#{`git rev-parse HEAD`.strip}#{uncommitted_suffix}"
+        if system("git describe")
+          `git describe --always --dirty='-dirty-#{SecureRandom.hex(8)}`
```
djmb, line 256: `git describe` doesn't treat untracked files as dirty, could we check for them and append the `dirty-` suffix manually?

lib/mrsk/configuration.rb
```diff
@@ -252,10 +252,11 @@ def role_names
 
     def git_version
       @git_version ||=
-        if system("git rev-parse")
-          uncommitted_suffix = Mrsk::Utils.uncommitted_changes.present? ? "_uncommitted_#{SecureRandom.hex(8)}" : ""
+        if system("git describe")
+          git_reference = `git describe --always`
+          dirty_suffix  = "dirty-#{SecureRandom.hex(8)}" if Mrsk::Utils.uncommitted_changes.present?
```
hey-leon, line 257: I suspect the idea of including the "_uncommited_" or "-dirty-" part in the version was originally intended to be a hint to the developer that their was uncommited changes in their build.

I am wondering if it would be better to just print a warning to the console and keep the git_version a bit more succinct? We would still need the random chars to avoid image cache clashes though.

hey-leon:
This has been revised to be up to date with the new `Kamal::Git` module

xiaohui-zhangxh:
need this to be merged, then we can ping the correct container's hostname

hey-leon:
I am going to close this as stale at this point. If anyone still needs this and would like to take over feel free to. It is no longer relevant for me. Thanks!

PR #412, opened 2023-08-08
Shorten git version to generate correct container DNS hostname

xiaohui-zhangxh:
Here I got a problem on resolving app IP address by container name, like this **'myapp-web-staging-599126a8f6c31f286636a8675226d4ab6585287c_uncommitted_94602f20085489c4'** has **87** chars, according to https://en.wikipedia.org/wiki/Subdomain#Overview , **"Each label may contain from 0 to 63 [octets]"**, this is invalid subdomain, so I can't container IP with:

```bash
/srv/www # ping myapp-web-staging-599126a8f6c31f286636a8675226d4ab6585287c_uncommitted_94602f20085489c4
ping: bad address 'myapp-web-staging-599126a8f6c31f286636a8675226d4ab6585287c_uncommitted_94602f20085489c4'
```

We can get short git version instead of long one, and make a version as short as possible.

### DIFF
--- a/lib/mrsk/configuration.rb
+++ b/lib/mrsk/configuration.rb
@@ -250,12 +250,14 @@ class Mrsk::Configuration
       raw_config.servers.is_a?(Array) ? [ "web" ] : raw_config.servers.keys.sort
     end
 
+    # git version will be part of container hostname, DNS has a limit of 63 characters
+    # so, make this a bit shorter, see: https://en.wikipedia.org/wiki/Subdomain#Overview
     def git_version
       @git_version ||=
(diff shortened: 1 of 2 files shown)


hey-leon:
I would also be interested in this @xiaohui-zhangxh. I posted a PR recently. Different solution to yours https://github.com/mrsked/mrsk/pull/409. I think I would be happy with either but I agree I think shortening the container name would be good.

xiaohui-zhangxh:
@hey-leon didn't notice you have created one. Closing due to duplicate

PR #417, opened 2023-08-13
Fix builder registry cache when using default registry

manastyretskyi:
When using the following configuration with default registry, e.g. not specifying `registry.server`, it will fail since `ref` that is starting with `/` is not valid. I updated code so that leading `/` will be omitted if `registry.server` is blank.
```yaml
builder:
  cache:
    type: registry
  remote:
    arch: arm64
    host: ssh://root@ip
```

Error that I was getting:
```bash
docker stderr: ERROR: failed to solve: failed to configure registry cache exporter: invalid reference format
```

### DIFF
--- a/lib/mrsk/configuration/builder.rb
+++ b/lib/mrsk/configuration/builder.rb
@@ -97,11 +97,7 @@ class Mrsk::Configuration::Builder
     end
 
     def cache_image_ref
-      if @server.present?
-        "#{@server}/#{cache_image}"
-      else
-        cache_image
-      end
+      [ @server, cache_image ].compact.join("/")
     end
 
     def cache_from_config_for_gha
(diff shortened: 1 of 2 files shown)


lib/mrsk/configuration/builder.rb
```diff
@@ -96,19 +96,27 @@ def cache_image
       @options["cache"]&.fetch("image", nil) || "#{@image}-build-cache"
     end
 
+    def cache_image_ref
+      if @server.present?
+        "#{@server}/#{cache_image}"
+      else
+        cache_image
+      end
```
djmb, line 104: [ @server, cache_image ].compact.join("/")

manastyretskyi: Done ✅

PR #423, opened 2023-08-22
Rename project to Kamal

dhh:
> The kamal is one of the earliest navigational tools that used measuring altitude to determine latitude. The word kamal means “guide” in Arabic.1 Introduced to the Europeans by the Arab people, the kamal was useful in determining altitude, and lead to the development of newer tools such as the cross-staff. Famed explorer Vasco da Gama’s experienced navigator used the kamal to help them navigate around the tip of Africa as they sailed to Asia.

https://exploration.marinersmuseum.org/object/kamal/


harisvsulaiman:
Why though?

nickhammond:
Is MRSK a copyright issue?

dhh:
A certain shipping company rattled their saber.

dhh:
Will sort CI issues in a follow-up.

unixmonkey:
Just as an FYI, [Kemal](https://kemalcr.com) is a longtime and fairly popular Sinatra-like web framework for Crystal.

It also seems like they [might be the same word in Turkish and Arabic when used as a person's name](https://en.wikipedia.org/wiki/Kemal_(name)).

Just pointing this out since it might cause some confusion in both communities.

brandondrew:
> Just as an FYI, [Kemal](https://kemalcr.com) is a longtime and fairly popular Sinatra-like web framework for Crystal.
> 
> It also seems like they [might be the same word in Turkish and Arabic when used as a person's name](https://en.wikipedia.org/wiki/Kemal_(name)).
> 
> Just pointing this out since it might cause some confusion in both communities.

It's not an issue in print, since the spelling is different.  I'm pretty sure that's were it will be used 99% of the time.  And although there is definitely overlap between the Ruby and Crystal communities, the context should easily disambiguate in any spoken conversations.

milushov:
Okay, it's definitely easier to type

image

PR #434, opened 2023-08-26
Supports Passing SSH Agent Socket to Build Options

rience:
This PR adds support for passing ssh agent socket as documented in here: https://docs.docker.com/engine/reference/commandline/buildx_build/#ssh

This allows accessing private repositories without passing GITHUB_TOKEN.

If that's accepted - I'll follow up with PR to Kamal site with documentation.

P.S. Only after I opened this one I noticed #235 doing similar thing. This one though does not pass any SSH arguments if not configured - so it doesn't affect default behaviour. I'm happy to contribute to the other one if said so to get this merged.


rience:
Removing because just noticed #235 doing the same.

PR #438, opened 2023-08-30
Copy env files to remote hosts

djmb:
Setting env variables in the docker arguments requires having them on the deploy host.

Instead we'll add two new commands `kamal env push` and `kamal env delete` which will manage copying the environment as .env files to the remote host.

Docker will pick up the file with `--env-file <path-to-file>`. Env files will be stored under `<kamal run directory>/env`.

Running `kamal env push` will create env files for each role and accessory, and traefik if required.

`kamal envify` has been updated to also push the env files.

By avoiding using `kamal envify` and creating the local and remote secrets manually, you can now avoid accessing secrets needed for the docker runtime environment locally. You will still need build secrets.

One thing to note - the Docker doesn't parse the environment variables in the env file, one result of this is that you can't specify multi-line values - see https://github.com/moby/moby/issues/12997.

We maybe need to look docker config or docker secrets longer term to get around this.

Hat-tip to @kevinmcconnell - this was his idea.

Docs PR: https://github.com/basecamp/kamal-site/pull/22

### DIFF
--- a/lib/kamal/cli/env.rb
+++ b/lib/kamal/cli/env.rb
@@ -1,0 +1,52 @@
+require "tempfile"
+
+class Kamal::Cli::Env < Kamal::Cli::Base
+  desc "push", "Push the env file to the remote hosts"
+  def push
+    mutating do
+      on(KAMAL.hosts) do
+        KAMAL.roles_on(host).each do |role|
+          role_config = KAMAL.config.role(role)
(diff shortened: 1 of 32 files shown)


snopoke:
FYI this broke our deploy since the env files didn't exist and aren't created by `deploy`. There is also no reference to this change in the docs.

PR #443, opened 2023-09-05
Configurable Healthcheck Expose Port

rience:
Healthcheck port (exposed one) is hard coded to 3999. This is a problem when doing parallel deployments and might result in this error

```
ERROR (SSHKit::Command::Failed): Exception while executing on host plus-6fc20fca3f82fa64049fec08bac4b533.svc: docker exit status: 125
--
  | docker stdout: fec19f2cf5310d4900e21ed48681ce9e896901185f656cfd1797d109af9bf251
  | docker stderr: docker: Error response from daemon: driver failed programming external connectivity on endpoint healthcheck-xxx (92d063a6607cbf7525970c123a75598dcde5dae34862ad9e436c05e9bcb82406): Bind for 0.0.0.0:3999 failed: port is already allocated.
```

This PR adds an option to `healthcheck` configuration and allow to pass `exposed_port` via configuration. Default value stays the same and is still 3999.

### DIFF
--- a/lib/kamal/commands/healthcheck.rb
+++ b/lib/kamal/commands/healthcheck.rb
@@ -1,5 +1,4 @@
 class Kamal::Commands::Healthcheck < Kamal::Commands::Base
-  EXPOSED_PORT = 3999
 
   def run
     web = config.role(:web)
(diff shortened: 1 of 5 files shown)


brunoprietog:
Can you document this in [kamal-site](https://github.com/basecamp/kamal-site)? Otherwise I will try to do it when I have some time

PR #444, opened 2023-09-05
Configurable Number of Lines in Healthcheck Log Output

rience:
While I was configuring Kamal for our deployments I was limited by number of lines that (by default) Kamal is printing out when running healthcheck. It is hardcoded to "50" and quite often I was only looking at the bottom of stack trace and could not figure out what has gone wrong.

Container is removed after failed deployment and I couldn't find a way to get these logs from anywhere.

So this PR is adding configurable number of lines for healthcheck.

### DIFF
--- a/lib/kamal/commands/healthcheck.rb
+++ b/lib/kamal/commands/healthcheck.rb
@@ -26,7 +26,7 @@ class Kamal::Commands::Healthcheck < Kamal::Commands::Base
   end
 
   def logs
-    pipe container_id, xargs(docker(:logs, "--tail", 50, "2>&1"))
+    pipe container_id, xargs(docker(:logs, "--tail", lines, "2>&1"))
   end
 
   def stop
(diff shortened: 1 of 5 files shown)


djmb:
Hi @rience, thank you for the PR!

There are some conflicts, if you could resolve them when you have time?

Also what do you think about changing the setting name to `log_lines`?

rience:
Hi @djmb,

That's rebased and name of setting has changed to `log_lines`. There is one test that is failing - I cannot see how this is connected to my test though.

djmb:
@rience - The tests passed on a re-run. The integration tests are running with docker inside docker inside a GitHub action so we do get the odd random failure. Thanks for the PR!

rience:
Sure - one more thing @djmb - I've also added this short description to `kamal-site`: https://github.com/basecamp/kamal-site/pull/28

PR #463, opened 2023-09-13
Load Deploy Config with *.erb Extension if Exists

rience:
At the moment configuration is by default loaded from `deploy.yml` file. It's the evaluated using ERB. But that means that IDE won't pick up ERB dialect. It'd be great if you could create `deploy.yml.erb` file and do not have to pass path to this file using `-c` flag.

So this PR is adding two things

1. If configuration file with "erb" extension exists - it'll be loaded instead
    * For example if `deploy.yml.erb` exists - it'll be loaded instead of `deploy.yml`
    * If custom config `custom_deploy.yml.erb` exists - it'll also be loaded instead of `custom_deploy.yml` (Note: this will also happens when you specify `-c custom_deploy.yml`)
2. It's also changing one of fixtures from `deploy.erb.yml` to `depoy.yml.erb` since I think this should be default naming for template file (just like in Rails you have `*.html.erb`)

### DIFF
--- a/lib/kamal/cli/base.rb
+++ b/lib/kamal/cli/base.rb
@@ -41,6 +41,13 @@ module Kamal::Cli
         options.merge(@_initializer.last[:class_options] || {})
       end
 
+    def config_file_path
+      path = File.expand_path(options[:config_file])
+      return "#{path}.erb" if File.exist?("#{path}.erb")
+
+      path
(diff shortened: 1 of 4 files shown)


rience:
Closing this as #391 is I think better solution than this since it honours "destination" setting.

Issue #465, opened 2023-09-14
sprockets-manifest missing from public


npezza93:
I think this is related to https://github.com/basecamp/kamal/pull/449. It seems like it is missing copying over the .sprockets-manifest.json. If i launch base using an existing container the manifest is missing but if i dont reuse it is present. So the file is a part of the image but it is missing for running containers.

PR #466, opened 2023-09-14
Include hidden files when copying assets

npezza93:
By default it appears /* doesnt include hidden files. When using sprockets with rails, a .sprockets-manifest.json file is generated which needs to be carried over. Without the manifest file rails apps complain about missing assets. 

Fixes https://github.com/basecamp/kamal/issues/465


djmb:
Thanks for raising this @npezza93! I'm going to close if as there were some failing tests and it has been resolved in #467.

npezza93:
Awesome, thanks @djmb!

PR #467, opened 2023-09-15
Copy all files into asset volume

djmb:
Adding -T to the copy command ensures that the files are copied at the same level into the target directory whether it exists or not.

That allows us to drop the `/*` which was not picking up hidden files.

Fixes: https://github.com/basecamp/kamal/issues/465

### DIFF
--- a/lib/kamal/commands/app.rb
+++ b/lib/kamal/commands/app.rb
@@ -234,6 +234,6 @@ class Kamal::Commands::App < Kamal::Commands::Base
     end
 
     def copy_contents(source, destination, continue_on_error: false)
-      [ :cp, "-rn", "#{source}/*", destination, *("|| true" if continue_on_error)]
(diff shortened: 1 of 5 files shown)


lib/kamal/commands/app.rb
```diff
@@ -234,6 +234,6 @@ def find_and_remove_older_siblings(path)
     end
 
     def copy_contents(source, destination, continue_on_error: false)
-      [ :cp, "-rn", "#{source}/*", destination, *("|| true" if continue_on_error)]
+      [ :cp, "-rnT", "#{source}", destination, *("|| true" if continue_on_error)]
```
jeremy, line 237: Looks like `-T` is GNU-only. Copying "#{source}/.", "#{destination}/." could be a workaround, avoiding the differing trailing-slash behavior in BSD vs GNU.

PR #482, opened 2023-09-20
feat: Use git email as performer when available

pagbrl:
Getting performer identifier from git configuration when it's available makes sense to me for a couple reasons :
- Value should be more unique and create less conflicts
- Git config can be overriden on a per-repo basis, which means a user could appear in the audit logs with their preferred git identity.

I think I made it so we'll still fall back to the sane default `whoami` that ensures a value is set.


igor-alexandrov:
@djmb, please take a look when you have time. Using email when possible makes total sense to me.

PR #487, opened 2023-09-21
Add option to skip .env check

snopoke:
This would allow deploying without having the `.env` file generated locally. This is especially useful for automated deploys which may not have access to the secrets e.g. 1Password.

Ideally this option should only be for deploy but the env check is part of the base command.

Side note: I'm new to the repo and Ruby so please let me know if there are test that should be added or doc updates etc.


djmb:
Thanks for pointing this out @snopoke! I've raised https://github.com/basecamp/kamal/pull/499 to fix this. We can remove the check entirely as it's generally not required anymore now that the environment is pushed to remote servers in a separate step.

Issue #492, opened 2023-09-23
`kamal app exec` without `--interactive` or `--reuse` does not allow to load env file


calmyournerves:
When upgrading to Kamal 1.0.0, we noticed our `kamal app exec 'bin/rails db:migrate'` command we run during deployment stopped working, because the environment variables are not available.

Running `kamal app exec --version=xyz --destination=staging --interactive 'bin/rails console'` works just fine. Without the `--interactive` option, it crashes with an error, saying it couldn't find an environment variable (`REDIS_URL` in our case).

Looking at the commands that are being executed, it seems that with `--interactive` the `--env-file` option is passed to the docker command, whereas without the `--interactive` option, the `--env-file` option is not passed. When using the interactive option, the command is executed with 

```
KAMAL.app(role: KAMAL.primary_role).execute_in_new_container_over_ssh(cmd, host: KAMAL.primary_host)
```

(note the `role: KAMAL.primary_role`) and without the flag, it is executed with 
```
KAMAL.app.execute_in_new_container(cmd)
```

It seems there is no way to pass/use a role when not using `--interactive` or `--reuse`.

I'm happy to open a PR, but I'm not entirely sure what the intentions behind `app exec` without these flags are. There is a global `--roles` option which could be used here, but the description for it is "Run commands on these roles instead of all (separate by comma)", so not sure if this would be correct.

PR #498, opened 2023-09-25
App exec with env file

djmb:
When calling `kamal app exec` for new non interactive containers, run the command per role on each server and include the role config including the environment.

Fixes: https://github.com/basecamp/kamal/issues/492


jmarchello:
@djmb Thanks! This was a major blocker for my deployment setup. 1.1.0 was a very well one surprise!

PR #499, opened 2023-09-25
Remove the env check

djmb:
The env check is not needed anymore as all the commands rely on the env files having already been created remotely.

The only place the env is needed is when running `kamal env push` and that will still raise an appropriate error.


snopoke:
Would be great to get this merged and released!

djmb:
@snopoke - that's released now in Kamal v1.1.0 👍

jmarchello:
@djmb i just want to thank you for this change. This was a pain point for me.

PR #500, opened 2023-09-25
Allow custom name for traefik container

n-studio:
Following this discussion [Question: Multiple apps on one host sharing the same traefik instance #257](https://github.com/basecamp/kamal/discussions/257)
it would be interesting to be able to host multiple apps on the same host.

For example to host demo apps for clients, for portfolios, for QA, etc. at minimum cost.
In my case it would enable me to deploy a staging app for each of PRs so the QA can validate them independently.

It is now difficult because each projects would create a container with the same name `traefik`. The deploy would fail because a container with the name `traefik` already exists.

Instead, this PR allows us to give a unique name to the traefik containers.

Performance wise, it's terrible to have multiple instances of traefik, but for these specific use cases performance doesn't matter.
The change is minimal and has no impact on maintenance.

```yml
# USAGE:
traefik:
  name: traefik-my-app-pr-500
  args:
    entrypoints.my-app-pr-500-web.address: ':80'

labels:
  traefik.http.routers.my-app-pr-500-web.entrypoints: my-app-pr-500-web
```

### DIFF
--- a/lib/kamal/commands/traefik.rb
+++ b/lib/kamal/commands/traefik.rb
@@ -8,7 +8,7 @@ class Kamal::Commands::Traefik < Kamal::Commands::Base
   }
 
   def run
-    docker :run, "--name traefik",
+    docker :run, "--name #{container_name}",
       "--detach",
       "--restart", "unless-stopped",
       *publish_args,
@@ -23,11 +23,11 @@ class Kamal::Commands::Traefik < Kamal::Commands::Base
   end
 
   def start
-    docker :container, :start, "traefik"
+    docker :container, :start, container_name
   end
 
   def stop
-    docker :container, :stop, "traefik"
+    docker :container, :stop, container_name
   end
 
   def start_or_run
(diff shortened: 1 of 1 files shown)


igor-alexandrov:
@n-studio great job, please update tests accordingly.

n-studio:
It seems @dhh wants to get rid of traefik for Kamal 2. (https://twitter.com/dhh/status/1766151697817313628)

Maybe this PR isn't relevant anymore?

djmb:
Hey @igor-alexandrov - yes we are looking at getting rid of Traefik for Kamal v2.0!

It's worked OK for us, but there's are some issues:
1. We need the cord file workaround for zero downtime deployments - https://github.com/basecamp/kamal/pull/439
2. The difficulties in running multiple apps on a single host
3. Traefik is fairly complicated - there's a lot to learn and it does a lot more than we need from it.

So we are building a new minimal proxy to replace Traefik designed just for Kamal (not Thruster). You'll switch between hosts by sending a command to it, and as much of the config as possible will be specified in that command (timeouts, max request size etc). It will also have SSL support like Thruster does, but beyond that we'll try and keep it as minimal as possible.

Because you can do so much with Traefik, there will be features that people are using with Kamal. We'll not try to replicate everything, but we may judiciously add some things that are commonly needed. But otherwise those may need to be implemented elsewhere, whether in the deployed containers or in a load balancer external to the deployed host.

igor-alexandrov:
Thank for the clarification @djmb. If you need any help with this new proxy or Kamal 2.0 in general – please let me know.

scart88:
@djmb will be something written from scratch, or you will leverage something like https://github.com/cloudflare/pingora

Here is their example for the proxy
https://github.com/cloudflare/pingora/blob/main/pingora-proxy/examples/load_balancer.rs

Thank you!

djmb:
Closing this as we'll remove Traefik in v2.0

dhh:
Lots of other options to pursue if Kamal 2 doesn't work for you. Including staying with Kamal 1. I think the only absurd thing here is to complain how your particular situation doesn't work with a piece of free software you got as a gift.

dhh:
Apology accepted. We will continue sharing our open source gifts! And if Kamal 1, with its Traefik integration, is working well for you, no need to look longingly for Kamal 2, which will run its own proxy.

dszymon:
@dhh do you plan to use an existing toolkit (e.g., Microsoft YARP)? Do you already know the list of supported protocols in the initial version?

dhh:
No. We're writing our own proxy that's tuned just for this. Will share more when it's cooked. Shouldn't be that much longer.

PR #501, opened 2023-09-26
Optionally Skip Push for "envify"

rience:
Pushing ".env" to remote host is not always required. If you just want to manage containers, traefik or accessories and initial push has already been done, that's adding additional step. Additionally - it reduces a risk of pushing incorrect ".env" to remote server.

This PR is adding "—skip_push" flag to "envify" command so it's possible to only generate it locally.


rience:
You mean - default will be not doing push and then adding `-p` would do push?

jmarchello:
I think the flag should be '--skip-push' (kabob case) to match the same flag from 'kamal deploy'.

djmb:
We push the env file by default because:

1. That's one less step to get started
2. Combining the two if safer because you are less likely to push a stale file if you've just generated it

But it makes sense to have this flag in cases where you don't want to push immediately.

Issue #303, opened 2023-05-20
Bad error when accessory cli is entered wrong


vertis:
When trying to stop/remove an accessory I incorrectly edited the command and it gave a poor error message.

What I meant to type was `mrsk accessory stop newrelic -d production` and what I actually ran was `mrsk accessory details stop -d production`:

Which is a nonsense command, but it gives:

```bash
$ mrsk accessory details stop -d production
  ERROR (Module::DelegationError): Mrsk::Commands::Accessory#hosts delegated to accessory_config.hosts, but accessory_config is nil: #<Mrsk::Commands::Accessory:0x0000ffff7bd31708 @config=#<Mrsk::Configuration:0x0000ffff7bd3c838 @raw_config=#<ActiveSupport::InheritableOptions {}>, @destination="production", @declared_version=nil, @roles=[#<Mrsk::Configuration::Role:0x0000ffff7bd3c158 @name="web", @config=#<Mrsk::Configuration:0x0000ffff7bd3c838 ...>, @hosts=["170.64.164.114", "170.64.172.8"]>], @accessories=[#<Mrsk::Configuration::Accessory:0x0000ffff7bd314d8 @name="newrelic", @config=#<Mrsk::Configuration:0x0000ffff7bd3c838 ...>, @specifics={"image"=>"newrelic/infrastructure:latest", "roles"=>["web"], "env"=>{"clear"=>{"NRIA_LICENSE_KEY"=>"NOTTHEREALKEY"}}, "options"=>{"network"=>"host", "cap-add"=>"SYS_PTRACE", "privileged"=>true, "pid"=>"host"}, "volumes"=>["/:/host:ro", "/var/run/docker.sock:/var/run/docker.sock"]}>]>, @accessory_config=nil>
```

This is a confusing message because the problem isn't that there isn't a hosts config option, it's that I've used `stop` where the `accessory_name` (in my case `newrelic`) should be.

It would be helpful if it told me that it can't find my accessory `stop`.

dhh:
Please do explore a PR for this 👍

PR #502, opened 2023-09-26
Fix accessory name check passing to `kamal accessory`

latyshev:
This PR closes an [issue](https://github.com/basecamp/kamal/issues/303 ) when wrong accessory name is passing to the `kamal accessory`.


lib/kamal/cli/accessory.rb
```diff
@@ -226,19 +226,17 @@ def remove_service_directory(name)
 
   private
     def with_accessory(name)
-      if accessory = KAMAL.accessory(name)
-        yield accessory, accessory_hosts(accessory)
+      if KAMAL.config.accessory(name)
+        yield KAMAL.accessory(name), accessory_hosts(accessory)
       else
-        error_on_missing_accessory(name)
+        raise ArgumentError, error_on_missing_accessory(name)
```
igor-alexandrov, line 232: Why did you decide to raise ArgumentError instead of using `Thor#error` method?

latyshev: Reverted it back.

igor-alexandrov:
@djmb I think, this can be merged.

PR #506, opened 2023-09-30
Allow for Custom Accessory Service Name

rience:
In my company we deploy several applications to single vm. They all share accessories - Postgres, Redis. Unfortunately at the moment - service name for accessory is created by concatenating deployment app name with name of accessory. So if I deploy `app1` and then `app2` - I would end up with two Postgres instances - `app1-postgres` and `app2-postgres`. Because of that we have separate YAML file - only for managing accessories.

This PR adds `service` setting to accessory - and that allows for custom service name. Of course previous behaviour stays - this is optional.

EDIT: I believe it could also help with this discussion: https://github.com/basecamp/kamal/discussions/257

### DIFF
--- a/lib/kamal/configuration/accessory.rb
+++ b/lib/kamal/configuration/accessory.rb
@@ -8,7 +8,7 @@ class Kamal::Configuration::Accessory
   end
 
   def service_name
-    "#{config.service}-#{name}"
+    specifics["service"] || "#{config.service}-#{name}"
   end
 
   def image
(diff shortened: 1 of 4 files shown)


igor-alexandrov:
Looks good to me. Thanks for this!

lib/kamal/configuration/accessory.rb
```diff
@@ -8,7 +8,7 @@ def initialize(name, config:)
   end
 
   def service_name
-    "#{config.service}-#{name}"
+    specifics['service'] || "#{config.service}-#{name}"
```
igor-alexandrov, line 11: Please use double quotes for strings.

igor-alexandrov (commented):
This looks good to me. Thank you.

igor-alexandrov:
@rience can you please check locally, what test does not pass?

igor-alexandrov (approved):

PR #507, opened 2023-10-01
Add option to set an absolute directory path

### LINKED ISSUE #453: Accessory Data: Directories vs Volumes
I would like to store my accessories data in a dedicated `/var/` directory.

The are two options to mount accessories data: [Volumes](https://kamal-deploy.org/docs/configuration#using-volumes) and [Directories](https://kamal-deploy.org/docs/configuration#using-directories). The main difference is that `directories` creates the directory on the host before mounting the volume.     
`directories` seems to be the right choice for me since I don't want to create the directories manually.

The issue I'm facing is that `directories` are scoped to the user's home directory. When I set `directories` to `/var/lib/redis:/data`, the directory `$PWD/my-appname-redis//var/lib/redis` is created on the host.

I could create the directories manually and use the `volumes` key, but I would like to avoid that.

I'm not sure if this behavior is intended or not. If my deploy user has the permission to create directories in the `/var/lib` directory, I see no reason why Kamal should not allow me to do so.

---

Additional Information:

Volumes approach:
```yml
redis:
  volumes:
    - /var/lib/redis:/data
```

Output
```bash
# kamal accessory boot redis
Running docker run [...] --volume /var/lib/redis:/data
```

---

Directories approach:
```yml
redis:
  directories:
    - /var/lib/redis:/data
```

Output
```bash
# kamal accessory boot redis
Running docker run [...] --volume $PWD/my-appname-redis//var/lib/redis:/data
```

If there are options to change this behavior, I'll be happy to explore them and contribute to resolving this issue.

leonvogt:
This PR will introduce the option to set an absolute directory path.   

**Before** - For example, a Redis accessory directory like:
- `data:/data` would create the directory `$PWD/my-appname-redis/data`
- `/var/lib/redis:/data` would create the directory `$PWD/my-appname-redis//var/lib/redis`

**After** - With this change, a Redis accessory directory like:
- `data:/data` would create the directory `$PWD/my-appname-redis/data`
- `/var/lib/redis:/data` would create the directory `/var/lib/redis`

The decision on whether Kamal will create an absolute or relative directory is based on the presence of a leading `/` in the path.

Please note that this change may introduce a breaking change for some users...

resolves #453

### DIFF
--- a/lib/kamal/configuration/accessory.rb
+++ b/lib/kamal/configuration/accessory.rb
@@ -70,8 +70,8 @@ class Kamal::Configuration::Accessory
 
   def directories
     specifics["directories"]&.to_h do |host_to_container_mapping|
-      host_relative_path, container_path = host_to_container_mapping.split(":")
-      [ expand_host_path(host_relative_path), container_path ]
+      host_path, container_path = host_to_container_mapping.split(":")
+      [ expand_host_path(host_path), container_path ]
     end || {}
   end
 
@@ -138,13 +138,17 @@ class Kamal::Configuration::Accessory
 
     def remote_directories_as_volumes
       specifics["directories"]&.collect do |host_to_container_mapping|
-        host_relative_path, container_path = host_to_container_mapping.split(":")
-        [ expand_host_path(host_relative_path), container_path ].join(":")
+        host_path, container_path = host_to_container_mapping.split(":")
+        [ expand_host_path(host_path), container_path ].join(":")
       end || []
     end
 
-    def expand_host_path(host_relative_path)
-      "#{service_data_directory}/#{host_relative_path}"
+    def expand_host_path(host_path)
+      absolute_path?(host_path) ? host_path : "#{service_data_directory}/#{host_path}"
+    end
+
+    def absolute_path?(path)
+      Pathname.new(path).absolute?
     end
 
     def service_data_directory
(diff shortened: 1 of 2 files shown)


lib/kamal/configuration/accessory.rb
```diff
@@ -138,13 +138,17 @@ def remote_files_as_volumes
...
-        [ expand_host_path(host_relative_path), container_path ].join(":")
+        host_path, container_path = host_to_container_mapping.split(":")
+        [ expand_host_path(host_path), container_path ].join(":")
       end || []
     end
 
-    def expand_host_path(host_relative_path)
-      "#{service_data_directory}/#{host_relative_path}"
+    def expand_host_path(host_path)
+      absoulte_path?(host_path) ? host_path : "#{service_data_directory}/#{host_path}"
```
strzibny, line 147: There is a typo

leonvogt: Thanks for catching that!

lib/kamal/configuration/accessory.rb
```diff
@@ -138,13 +138,17 @@ def remote_files_as_volumes
...
     end
 
-    def expand_host_path(host_relative_path)
-      "#{service_data_directory}/#{host_relative_path}"
+    def expand_host_path(host_path)
+      absolute_path?(host_path) ? host_path : "#{service_data_directory}/#{host_path}"
+    end
+
+    def absolute_path?(path)
+      path.slice(0) == "/"
```
djmb, line 151:
```suggestion
      Pathname.new(path).absolute?
```

djmb (changes requested):

lib/kamal/configuration/accessory.rb
leonvogt: I didn't know something like that existed, thanks!

PR #513, opened 2023-10-03
Surround docker login password with quotes

kazarin:
Resolve https://github.com/basecamp/kamal/issues/514

### DIFF
--- a/lib/kamal/commands/registry.rb
+++ b/lib/kamal/commands/registry.rb
@@ -2,7 +2,7 @@ class Kamal::Commands::Registry < Kamal::Commands::Base
   delegate :registry, to: :config
 
   def login
-    docker :login, registry["server"], "-u", sensitive(lookup("username")), "-p", sensitive(lookup("password"))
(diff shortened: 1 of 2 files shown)


lib/kamal/commands/registry.rb
```diff
@@ -2,7 +2,7 @@ class Kamal::Commands::Registry < Kamal::Commands::Base
   delegate :registry, to: :config
 
   def login
-    docker :login, registry["server"], "-u", sensitive(lookup("username")), "-p", sensitive(lookup("password"))
+    docker :login, registry["server"], "-u", sensitive(lookup("username")), "-p", sensitive("'#{lookup("password")}'")
```
djmb, line 5: This wouldn't work if the password contained a `'`.

We could use [String.shellescape](https://docs.ruby-lang.org/en/3.2/Shellwords.html#method-i-shellescape) instead. Even though it's unlikely to contain anything that need it, we should also escape the username.

djmb:
Closing in favour of https://github.com/basecamp/kamal/pull/704

PR #531, opened 2023-10-16
optionize reject nil value

xiaohui-zhangxh:
I think `optionize` shouldn't output the key which value is nil, we should treate nil as removing the key. If we need a key-only argument, the value should be `true` instead of `nil`.

eg.

production environment, we use http provider for traerik

```yaml
# config/deploy.yaml
traefik:
  args:
    providers.http.endpoint: "http://app:3000/v1/traefik/routes"
```

development environment, we are changing http provider to file provier

```yaml
# config/deploy.dev.yaml
traefik:
  args:
    providers.http.endpoint: null
    providers.file.directory: "/etc/traefik/enabled-sites"
    providers.file.watch: true
```


xiaohui-zhangxh:
duplicate of https://github.com/basecamp/kamal/pull/774

Issue #537, opened 2023-10-18
Accessory reboot all


dmitrytrager:
According to command list, this should work without additional changes:
`kamal accessory reboot all` should work and reboot every accessory.

Result:
`ERROR (Module::DelegationError): Kamal::Commands::Accessory#hosts delegated to accessory_config.hosts, but accessory_config is nil: #<Kamal::Commands::Accessory:0x0000000104591f78 ...`

Version 1.0.0

jmarchello:
@dmitrytrager Could you provide a snippet showing the accessories sections of your deploy.yml?

dmitrytrager:
@jmarchello 
```
accessories:
  db:
    image: postgres:14.8
    host: <ip>
    port: 5432
    env:
      clear:
        POSTGRES_USER: <user>
        POSTGRES_DB: <db_name>
      secret:
        - POSTGRES_PASSWORD
    directories:
      - data:/var/lib/postgresql/data
  redis:
    image: redis:7.0
    host: <ip>
    port: 6379
    directories:
      - data:/data
```
By the way, when I reboot them one by one, it works smoothly

jmarchello:
@dmitrytrager 

Looking at the code it appears that `reboot` does not currently support the `all` argument, while `boot` and a few others do. Here is what I see when I run the help command: 
```
# kamal accessory help
Commands:
  kamal accessory boot [NAME]        # Boot new accessory service on host (use NAME=all to boot all accessories)
  kamal accessory details [NAME]     # Show details about accessory on host (use NAME=all to show all accessories)
  kamal accessory exec [NAME] [CMD]  # Execute a custom command on servers (use --help to show options)
  kamal accessory help [COMMAND]     # Describe subcommands or one specific subcommand
  kamal accessory logs [NAME]        # Show log lines from accessory on host (use --help to show options)
  kamal accessory reboot [NAME]      # Reboot existing accessory on host (stop container, remove container, start new container)
  kamal accessory remove [NAME]      # Remove accessory container, image and data directory from host (use NAME=all to remove all accessories)
  kamal accessory restart [NAME]     # Restart existing accessory container on host
  kamal accessory start [NAME]       # Start existing accessory container on host
  kamal accessory stop [NAME]        # Stop existing accessory container on host
 ```
 I can put up a PR to add this feature in a bit, unless you'd like to.

dmitrytrager:
Oh, I see.
Thanks, let me try then)

PR #550, opened 2023-10-25
feature: add NAME=all option for accessory reboot

dmitrytrager:
This missing cli option was discusses within [the issue](https://github.com/basecamp/kamal/issues/537)

### DIFF
--- a/lib/kamal/cli/accessory.rb
+++ b/lib/kamal/cli/accessory.rb
@@ -49,17 +49,21 @@ class Kamal::Cli::Accessory < Kamal::Cli::Base
     end
   end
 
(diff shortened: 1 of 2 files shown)


djmb:
Thanks for the PR @dmitrytrager! 

Could you add a test for this? There's a similar one in [test/cli/accessory_test.rb for `boot all`](https://github.com/basecamp/kamal/blob/b732b2dd5599037f2a4719b6e2ce3cd62f5dd627/test/cli/accessory_test.rb#L14-L28).

dmitrytrager:
Sure, will do

dmitrytrager:
@djmb updated

PR #558, opened 2023-10-30
Add --skip_push option to setup

GeNiuS69:
Add possibility to skip build and push image on setup if you already have builded image in registry, e.g. after GitLab CI/CD.

### DIFF
--- a/lib/kamal/cli/main.rb
+++ b/lib/kamal/cli/main.rb
@@ -4,15 +4,13 @@ class Kamal::Cli::Main < Kamal::Cli::Base
   def setup
     print_runtime do
       mutating do
-        invoke_options = deploy_options
-
         say "Ensure Docker is installed...", :magenta
-        invoke "kamal:cli:server:bootstrap", [], invoke_options
+        invoke "kamal:cli:server:bootstrap"
 
         say "Push env files...", :magenta
-        invoke "kamal:cli:env:push", [], invoke_options
+        invoke "kamal:cli:env:push"
 
-        invoke "kamal:cli:accessory:boot", [ "all" ], invoke_options
+        invoke "kamal:cli:accessory:boot", [ "all" ]
         deploy
       end
     end
--- a/lib/kamal/cli/main.rb
+++ b/lib/kamal/cli/main.rb
@@ -1,15 +1,18 @@
 class Kamal::Cli::Main < Kamal::Cli::Base
   desc "setup", "Setup all accessories, push the env, and deploy app to servers"
+  option :skip_push, aliases: "-P", type: :boolean, default: false, desc: "Skip image build and push"
   def setup
     print_runtime do
       mutating do
+        invoke_options = deploy_options
+
         say "Ensure Docker is installed...", :magenta
-        invoke "kamal:cli:server:bootstrap"
+        invoke "kamal:cli:server:bootstrap", [], invoke_options
 
         say "Push env files...", :magenta
-        invoke "kamal:cli:env:push"
+        invoke "kamal:cli:env:push", [], invoke_options
 
-        invoke "kamal:cli:accessory:boot", [ "all" ]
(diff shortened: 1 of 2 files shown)


igor-alexandrov:
@GeNiuS69 you passed for no reason `invoke_options` to `kamal:cli:server:bootstrap`, `kamal:cli:env:push` and to `kamal:cli:accessory:boot`. However the only change that is really need to make this work is the change you made on line 3: https://github.com/basecamp/kamal/pull/558/files#diff-cd863d3bebefa5ecfbacd577abdff981aa7695ff9c0a39a634ac1baaee3901b2R3.

Please update the PR accordingly.

GeNiuS69:
@igor-alexandrov done

test/cli/main_test.rb
```diff
@@ -2,12 +2,17 @@
 
 class CliMainTest < CliTestCase
   test "setup" do
+    invoke_options = { "config_file" => "test/fixtures/deploy_simple.yml", "version" => "999", "skip_hooks" => false }
+
```
Sija, line 6: That seems unneeded as well:

```suggestion
```

igor-alexandrov: Yes, but at the same time I don't see tests, which cover `skip_push` option for the `setup` task.

GeNiuS69: Yes, sure, my bad. Removed

lib/kamal/cli/main.rb
```diff
@@ -1,5 +1,6 @@
 class Kamal::Cli::Main < Kamal::Cli::Base
   desc "setup", "Setup all accessories, push the env, and deploy app to servers"
+  option :skip_push, aliases: "-P", type: :boolean, default: false, desc: "Skip image build and push"
```
igor-alexandrov, line 3: Please add tests for the `skip_push` option similar to https://github.com/basecamp/kamal/blob/main/test/cli/main_test.rb#L40

igor-alexandrov (changes requested):

lib/kamal/cli/main.rb
kalashnikovisme: Looks like tests were added :slightly_smiling_face: 

Is there something that prevents us from having this wonderful feature in the release? :slightly_smiling_face:

kalashnikovisme:
Tried this version of gem locally.

```
~: kamal setup --skip-push
  INFO [e011c4ce] Running /usr/bin/env mkdir -p .kamal on 165.227.149.148
  INFO [e011c4ce] Finished in 1.337 seconds with exit status 0 (successful).
Acquiring the deploy lock...
Ensure Docker is installed...
Releasing the deploy lock...
  Finished all in 1.9 seconds
ERROR: "kamal server bootstrap" was called with arguments ["--skip-push"]
Usage: "kamal server bootstrap"
```

Trying to figure out what is going on :slightly_smiling_face:

GeNiuS69:
@kalashnikovisme hmm, it was my initial solution, but I couldn't defend it

igor-alexandrov:
@GeNiuS69 sorry, it was my fault, now I understood why you used `deploy_options` to pass params to internal tasks.
@kalashnikovisme in your solution assignment of `invoke_options = deploy_options` seems to be unnecessary, `deploy_options` can be used directly.

kalashnikovisme:
@igor-alexandrov yeah! You're right :slightly_smiling_face: I don't like this assignment `invoke_options = deploy_options` either. I've added because the same was in the next function 

```ruby
def deploy
    runtime = print_runtime do
      mutating do
        invoke_options = deploy_options # <-----------------------
        # ...
      end
    end
end
```

djmb:
Thanks for the PR @GeNiuS69!

zmoon460:
Very important and useful feature, thank you

PR #559, opened 2023-10-30
Supporting multi-dc deploys

mdkent:
This is an early sketch for discussion.

This branch: 

- Enables aliases+anchors in yaml.
  - This is huge for DRY within the layered config.yaml and config.staging.yaml to make controlling the large number of roles to sustain a multi-site deployment realistic.
- Drops the requirement to provide a :web role. Let's just use whatever role is running traefik and call it good enough. 
  - This lets us name all of the roles after a datacenter - rather than carrying one awkward :web role.
- Drops the requirement for roles to have hosts. The fact they're empty is harmless in the end if the user is informed. 
  - This let us toggle roles on/off for partial deployments. Very useful in outages.

### DIFF
--- a/lib/kamal/commands/healthcheck.rb
+++ b/lib/kamal/commands/healthcheck.rb
@@ -1,7 +1,7 @@
 class Kamal::Commands::Healthcheck < Kamal::Commands::Base
 
   def run
-    web = config.role(:web)
+    web = config.role(:web) || config.roles.select(&:running_traefik?).first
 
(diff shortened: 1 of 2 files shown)


mdkent:
I've broken this out into

https://github.com/basecamp/kamal/pull/577
https://github.com/basecamp/kamal/pull/578
https://github.com/basecamp/kamal/pull/582
https://github.com/basecamp/kamal/pull/583

Issue #338, opened 2023-06-19
`docker: command not found` error on host server


jeromedalbert:
I was watching the [Introducing MRSK](https://www.youtube.com/watch?v=LL1cV2FXZ5I) Youtube screencast, and one of the steps early on is to run `mrsk deploy`. But I was getting the following error:

```
$ mrsk deploy
Acquiring the deploy lock...
Log into image registry...
  INFO [347dbde0] Running docker login -u [REDACTED] -p [REDACTED] as jerome@localhost
  INFO [347dbde0] Finished in 1.225 seconds with exit status 0 (successful).
  INFO [63af6e8d] Running docker login -u [REDACTED] -p [REDACTED] on [REDACTED]
Releasing the deploy lock...
  Finished all in 2.5 seconds
  ERROR (SSHKit::Command::Failed): Exception while executing on host [REDACTED]: docker exit status: 127
docker stdout: Nothing written
docker stderr: bash: line 1: docker: command not found
```

It turned out that the video is out of sync with the current way of doing things. I should have followed the README instead, which mentions that you need to run the following command to install Docker on the server:

```
mrsk setup
```

This fixed the issue. Hopefully this writeup helps anyone who might have blindly followed the video like me.

jeromedalbert:
Related issue: https://github.com/mrsked/mrsk/issues/301

Danyaga:
Hello. I have to this issue even though I'm running `kamal setup`.

Basically, I added the second server and it seems that it isn't detecting that docker is installer in the first server, but not in the second.

PR #567, opened 2023-11-06
Added logic to detect if curl is installed on servers and, when the OS are debian-like, installing it using apt before installing docker

Danyaga:
Hello everyone.

This PR is a follow-up from my comment [here](https://github.com/basecamp/kamal/issues/338#issuecomment-1793721838) where basically I understood that the server should have curl already installed, [but documentation says that kamal setup should actually install it when not installed on a server.](https://github.com/basecamp/kamal-site/blob/main/_docs/installation.md#L48)

This PR is intended to: 
- Define a Kamal::Commands for curl that is able to check wherever curl is installed using `--version` and that curl is installed with apt on Debian-like OSs or return a understandable error when the installation was not possible (E. G. unrecognized OS or not being a superuser);
- Upon `kamal setup`, check that curl is installed and try the installation; 
- Define unit tests to check that the methods I defined are returning the right results.

I hope the code is actually in continuity with the style of the repo as my Ruby knowledge is pretty basic. I'm available further follow-up and any improvements to what I have done to meet the requirements of the project.

### DIFF
--- a/lib/kamal/cli/server.rb
+++ b/lib/kamal/cli/server.rb
@@ -1,22 +1,37 @@
 class Kamal::Cli::Server < Kamal::Cli::Base
   desc "bootstrap", "Set up Docker to run Kamal apps"
   def bootstrap
-    missing = []
+    missing_curl = []
+    missing_docker = []
 
     on(KAMAL.hosts | KAMAL.accessory_hosts) do |host|
-      unless execute(*KAMAL.docker.installed?, raise_on_non_zero_exit: false)
+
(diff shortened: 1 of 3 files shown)


djmb:
Thanks for the PR and the pointer about the incorrect documentation @Danyaga!

I'm going to close this in favour of https://github.com/basecamp/kamal/pull/710. We'll try to install with curl and wget and ensure that we stop in case of a failure. I'll update the documentation as well once this is released 👍

PR #573, opened 2023-11-08
Pre and post Traefik reboot hooks

djmb:
Provide pre and post reboot hooks for Traefik, that can be used to remove/add to an external load balancer to prevent requests from being sent during the reboot.

Works best with the --rolling setting, where each hook is called once per host.


queries
Sija, line 1: Looks like this file was committed by mistake?

djmb: Oh thanks, I'll remove!

PR #579, opened 2023-11-12
Add ability to stop containers asynchronously

yoelcabo:
## Motivation

We've used Kamal to move some **long-running jobs** (ffmpeg transcoding) from AWS to Hetzner. The savings in $$-to-big-tech are amazing, thanks a lot for this project 🙌

The only problem we have is that we cannot find a way to prevent our jobs from being stopped midway while also having quick deployments as part of our Monolith's CI/CD system. 

The simplest solution I could think of is:
- Kamal fires a `docker stop` command with a generous grace period and forgets about my workers.
- The worker stops picking up new jobs upon receiving the SIGTERM, and exits after finishing the current job.
- Just in case something goes wrong, Kamal persists the containers that should be dying and makes sure they are dead once the grace period is over.

I gave a bit more context on this discussion: https://github.com/basecamp/kamal/discussions/491

### `stop_asynchronously`
New configuration flag that can be applied per role: `stop_asynchronously`. This is more or less how we are using it
```
  flong_running_workers:
    hosts: 
      - MY IPS
    cmd: long_running_job_worker --do-not-pick-jobs-after-sigterm --grace-period-until-sigkill 600
    stop_asynchronously: true

stop_wait_time: 601
```

### Running docker stop in the background

The command that is run is something like:
```
nohup sh -c 'echo "CONTAINER_IDS" | xargs docker stop -t 3600' > /dev/null 2>&1 & disown
```

### Keeping track of stopped containers

We persist the records in a simple plain text file that looks like this:
```
# CONTAINER_ID,STOP_TIME
root@web-staging:~# cat .kamal/happyscribe-media-worker-staging-async_stop_records 
4bf1c771bcec,2023-11-12 08:18:20 UTC
```

If the container is still up in a subsequent deploy that happens after the recorded stop time, then Kamal stops it synchronously.

## Closing thoughts

I understand long-running jobs may not be the focus of this tool, and therefore the extra complexity in this PR might not be justified. The reason I built this is that I'd rather maintain a fork of Kamal than deploy our own Kubernetes cluster. Guidance on reducing this complexity would be greatly appreciated if you think this can be a good fit for Kamal.

Also: I am not very sure the way I encapsulated this behavior is in line with Kamal's architecture. I am more than open to feedback and happy to do any modification that would align better with the project.

### DIFF
--- a/lib/kamal/cli/app.rb
+++ b/lib/kamal/cli/app.rb
@@ -55,7 +55,12 @@ class Kamal::Cli::App < Kamal::Cli::Base
                   end
                 end
 
-                execute *app.stop(version: old_version), raise_on_non_zero_exit: false
+                if role_config.stop_async?
+                  ssh_context = self
+                  Kamal::Cli::Async::Stopper.new(app, ssh_context: ssh_context, version: old_version).stop
+                else
+                  execute *app.stop(version: old_version), raise_on_non_zero_exit: false
+                end
 
                 execute *app.clean_up_assets if role_config.assets?
               end
@@ -89,6 +94,22 @@ class Kamal::Cli::App < Kamal::Cli::Base
         roles.each do |role|
           execute *KAMAL.auditor.record("Stopped app", role: role), verbosity: :debug
           execute *KAMAL.app(role: role).stop, raise_on_non_zero_exit: false
+        end
+      end
+    end
+  end
+
+  desc "stop_async", "Stop app container on servers asynchronously"
+  def stop_async
+    using_version(version_or_latest) do |version|
+      mutating do
+        on(KAMAL.hosts) do |host|
+          ssh_context = self
+          roles = KAMAL.roles_on(host)
+          roles.each do |role|
+            app = KAMAL.app(role: role)
+            Kamal::Cli::Async::Stopper.new(app, ssh_context: ssh_context, version: version).stop
+          end
         end
       end
     end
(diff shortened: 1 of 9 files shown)


djmb:
Hi @yoelcabo,

I realize this has been sitting open for a long time! We don't want to add this feature to Kamal - its much simpler to always shut the old containers down. But I didn't want to close this without providing a recipe for doing what you want to do.

I haven't tested this but I think it should work. It involves post deploy hooks and `kamal app exec`. You can use them to create your own containers that kamal itself will not try to manage.

Firstly you will want a role and some dummy containers that do nothing in `config/deploy.yml`:

```yaml
servers:
  web:
    hosts:
      - web1.example.com
      - web2.example.com

  worker:
    hosts:
      - worker1.example.com
      - worker2.example.com
    cmd: sleep infinity
```

Then in a post deploy hook, you can manually boot a container and send the shutdown signal to old ones:
```sh
#!/bin/sh
set -e

GRACE_PERIOD=3600

# Signal existing worker exec containers to stop (they'll finish current jobs)
# Container name pattern: {service}-{role}-exec-{version}-{random}
echo "Signaling old workers to stop accepting jobs..."
kamal server exec --roles=worker "docker ps -q --filter 'name=${KAMAL_SERVICE}-worker-exec-'  --filter 'status=running' | xargs -r docker stop -t ${GRACE_PERIOD}"

# Don't wait for old containers to stop - start new workers immediately
echo "Starting new workers..."
kamal app exec --roles=worker --detach job_command

echo "Worker deployment complete - old workers will finish current jobs and exit"
```

yoelcabo:
completely understand @djmb !

Thanks so much for taking the time to provide this workaround. We'll try it out! This way we can easily update to the last version of Kamal :ok_hand:

djmb:
Cheers, @yoelcabo.

Please let me know how you get on!

mitdonga:
Thanks @djmb 

Hi @yoelcabo 

Is this solution working for you? I am also facing this same challenge. Thanks!

PR #577, opened 2023-11-11
Support customizing the primary_web_role

mdkent:
This adds support for controlling the `primary_web_role` instead of always defaulting to `web`.

This allow for more flexible role naming in configs, eg: `web_chicago`.

Example config:
```
servers:
  web_chicago:
    traefik: true
    hosts:
      - 192.168.0.1
      - 192.168.0.2
  web_tokyo:
    traefik: true
    hosts:
      - 192.168.0.3
      - 192.168.0.4
```

PR #580, opened 2023-11-12
Allow Kamal to run without traefik

yoelcabo:
Related discussion: https://github.com/basecamp/kamal/discussions/345

I know Kamal is designed for deploying web applications, as DHH was mentioning here: https://github.com/basecamp/kamal/issues/191#issuecomment-1501728562

That said, the code change is small, and I think there is a valid use case for not having a web role:

### Our use case
We are slowly migrating out of Heroku/AWS. Our first step was migrating some background workers. 

If a worker is unreliable, jobs will be picked by "someone else", but if a web container is unreliable, it can cause serious downtime. That's why we are keeping the web in Heroku for now, but the workers are slowly being moved to Hetzner with Kamal.

A workaround for us was running a web container that was never accessed. However, this makes the configuration confusing, creates unnecessary connections to the database and makes the deployment slower.

### DIFF
--- a/lib/kamal/cli/healthcheck.rb
+++ b/lib/kamal/cli/healthcheck.rb
@@ -3,6 +3,7 @@ class Kamal::Cli::Healthcheck < Kamal::Cli::Base
 
   desc "perform", "Health check current app version"
   def perform
+    return unless KAMAL.primary_role.running_traefik?
     on(KAMAL.primary_host) do
       begin
         execute *KAMAL.healthcheck.run
--- a/lib/kamal/cli/healthcheck.rb
+++ b/lib/kamal/cli/healthcheck.rb
@@ -3,7 +3,7 @@ class Kamal::Cli::Healthcheck < Kamal::Cli::Base
 
   desc "perform", "Health check current app version"
   def perform
-    raise "The primary host is not configured to run Traefik" unless KAMAL.config.primary_role.running_traefik?
+    raise "The primary host is not configured to run Traefik" unless KAMAL.config.role(KAMAL.config.primary_role).running_traefik?
     on(KAMAL.primary_host) do
       begin
         execute *KAMAL.healthcheck.run
--- a/lib/kamal/cli/healthcheck.rb
+++ b/lib/kamal/cli/healthcheck.rb
@@ -3,7 +3,7 @@ class Kamal::Cli::Healthcheck < Kamal::Cli::Base
 
   desc "perform", "Health check current app version"
   def perform
-    return unless KAMAL.primary_role.running_traefik?
+    raise "The primary host is not configured to run Traefik" unless KAMAL.config.primary_role.running_traefik?
     on(KAMAL.primary_host) do
       begin
         execute *KAMAL.healthcheck.run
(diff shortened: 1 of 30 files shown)


lib/kamal/cli/healthcheck.rb
```diff
@@ -3,6 +3,7 @@ class Kamal::Cli::Healthcheck < Kamal::Cli::Base
 
   desc "perform", "Health check current app version"
   def perform
+    return unless KAMAL.primary_role.running_traefik?
```
djmb, line 6: Could we raise an exception here instead and protect the callers with if statements? If someone managed to call this without a web role, I think it would be better to fail then to silently pass the check.

lib/kamal/commands/healthcheck.rb
```diff
@@ -2,6 +2,7 @@ class Kamal::Commands::Healthcheck < Kamal::Commands::Base
 
   def run
     web = config.role(:web)
+    return unless web.present?
```
djmb, line 5: Again could we raise here instead of returning?

djmb (changes requested):
Hi @yoelcabo!

I agree with your reasoning for this change. If asked for a couple of changes. I think you'll also need to make a few changes to get this working again after https://github.com/basecamp/kamal/pull/577 was merged.

yoelcabo:
Thanks a lot for the feedback :pray: @djmb 

I've addressed your comments and resolved the conflicts.

Due to https://github.com/basecamp/kamal/pull/577 there are some important differences in my PR:
- Rename `primary_web_role` configuration variable to `primary_role`. This is a breaking change for anyone who adopted https://github.com/basecamp/kamal/pull/577
- `primary_role` enables traefik by default. https://github.com/basecamp/kamal/pull/577 was raising if it was not enabled. Since my PR makes it possible to run without traefik, I think it makes sense that enabling it for the primary is the default.

cc @mdkent 

Regarding your comment on the `running_traefik?` check. This is my change:
- The check happens now in the deploy command instead. If traefik is disabled in the primary, the healthcheck is not performed.
- I added a raise for whenever someone invokes `kamal healthcheck perform` manually.
- I removed the check for the command seems this doesn't seem a possible code path.

lib/kamal/cli/healthcheck.rb
```diff
@@ -3,6 +3,7 @@ class Kamal::Cli::Healthcheck < Kamal::Cli::Base
 
   desc "perform", "Health check current app version"
   def perform
+    raise "The primary host is not configured to run Traefik" unless KAMAL.config.role(KAMAL.config.primary_role).running_traefik?
```
yoelcabo, line 6: `KAMAL.config.role(KAMAL.config.primary_role)` might be too verbose.

In my original changes `KAMAL.config.primary_role` was an instance of `Role`.

However, with #577, `KAMAL.config.primary_web_role` was introduced as a String.

I didn't want to change that to make the review simpler. I can submit a PR afterwards with a proposed cleanup if you think it's interesting

djmb: Hi @yoelcabo - yes if you could raise another PR to convert `KAMAL.config.primary_role` to a `Role` that would be great. I'll merge this PR now 🙏

PR #583, opened 2023-11-13
Add wildcards to roles and hosts filters

mdkent:
Add support for wildcard matches with `*` on roles and hosts.

 eg:
```
      --roles=*_chicago,*_tokyo
      --hosts=app-*
```

Useful for targeted deploys to groups of roles and hosts.

### DIFF
--- a/lib/kamal/commander.rb
+++ b/lib/kamal/commander.rb
@@ -28,11 +28,11 @@ class Kamal::Commander
   end
 
   def specific_roles=(role_names)
-    @specific_roles = config.roles.select { |r| role_names.include?(r.name) } if role_names.present?
+    @specific_roles = Kamal::Utils.filter_specific_items(role_names, config.roles) if role_names.present?
   end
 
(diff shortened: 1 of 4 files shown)


lib/kamal/commander.rb
```diff
@@ -28,11 +28,34 @@ def specific_primary!
...
   def specific_roles=(role_names)
-    @specific_roles = config.roles.select { |r| role_names.include?(r.name) } if role_names.present?
+    return unless role_names.present?
+
+    @specific_roles = []
+
+    Array(role_names).select do |sp|
+      @specific_roles += config.roles.select do |r|
+        # Only allow * for a wildcard
+        pattern = Regexp.escape(sp).gsub('\*', '.*')
```
djmb, line 38: Could you extract this to a method in `Kamal::Utils`?

mdkent: Done! Looks a bit cleaner.

PR #585, opened 2023-11-14
Add docker-setup hook

tsvallender:
This allows the user to make any necessary configuration changes to Docker before setting up any containers, allowing those configuration changes to take effect from the outset.

My primary motivation for adding this was wanting all the containers to be on the same Docker network. This change means the `docker-setup` hook script can ensure the network exists before deployment rather than needing to do so manually. This should be more broadly useful for any general Docker configuration that may be necessary by the user, however. The new hook point is necessary so it can be done _after_ we ensure Docker is installed but _before_ deploying any containers.

I’ve tested this manually but don’t fully understand the way hooks are being tested so would appreciate some assistance getting the new test to pass.

### DIFF
--- a/test/cli/server_test.rb
+++ b/test/cli/server_test.rb
@@ -23,10 +23,13 @@ class CliServerTest < CliTestCase
     SSHKit::Backend::Abstract.any_instance.expects(:execute).with('[ "${EUID:-$(id -u)}" -eq 0 ] || command -v sudo >/dev/null || command -v su >/dev/null', raise_on_non_zero_exit: false).returns(true).at_least_once
     SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:curl, "-fsSL", "https://get.docker.com", "|", :sh).at_least_once
     SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:mkdir, "-p", ".kamal").returns("").at_least_once
+    Kamal::Commands::Hook.any_instance.stubs(:hook_exists?).returns(true)
+    SSHKit::Backend::Abstract.any_instance.expects(:execute).with(".kamal/hooks/docker-setup", anything).at_least_once
 
     run_command("bootstrap").tap do |output|
       ("1.1.1.1".."1.1.1.4").map do |host|
         assert_match "Missing Docker on #{host}. Installing…", output
+        assert_match "Running the docker-setup hook", output
       end
     end
   end
--- a/lib/kamal/cli/server.rb
+++ b/lib/kamal/cli/server.rb
@@ -19,5 +19,7 @@ class Kamal::Cli::Server < Kamal::Cli::Base
     if missing.any?
       raise "Docker is not installed on #{missing.join(", ")} and can't be automatically installed without having root access and the `curl` command available. Install Docker manually: https://docs.docker.com/engine/install/"
     end
+
+    run_hook "docker-setup"
   end
 end
--- a/lib/kamal/cli/templates/sample_hooks/docker-setup.sample
+++ b/lib/kamal/cli/templates/sample_hooks/docker-setup.sample
@@ -1,0 +1,7 @@
+#!/usr/bin/env ruby
+
+# A sample docker-setup hook
+#
+# Sets up a Docker network which can then be used by the application’s containers
+
+ssh user@example.com docker network create kamal
--- a/test/integration/docker/deployer/app/.kamal/hooks/docker-setup
+++ b/test/integration/docker/deployer/app/.kamal/hooks/docker-setup
@@ -1,0 +1,3 @@
+#!/bin/sh
+echo "Docker set up!"
+mkdir -p /tmp/${TEST_ID} && touch /tmp/${TEST_ID}/docker-setup


lib/kamal/cli/main.rb
```diff
@@ -6,6 +6,8 @@ def setup
         say "Ensure Docker is installed...", :magenta
         invoke "kamal:cli:server:bootstrap"
 
+        run_hook "docker-setup"
```
djmb, line 9: Do we want to always run this, or only run it when docker is actually installed?

Either we should move calling the hook inside `kamal:cli:server:bootstrap` so that it is also called when that command is invoked directly.

djmb:
@tsvallender - I'm not sure why the test is failing, at first glance it looks like it should work. 

If you move the calling the hook into `kamal:cli:server:bootstrap` then we'll need to move the test into `test/cli/server_test.rb` anyway. Let's do that first and then we can debug it from there.

lib/kamal/cli/main.rb
tsvallender: Yes, good shout, that does make more sense :+1:

tsvallender:
> @tsvallender - I'm not sure why the test is failing, at first glance it looks like it should work.
> 
> If you move the calling the hook into `kamal:cli:server:bootstrap` then we'll need to move the test into `test/cli/server_test.rb` anyway. Let's do that first and then we can debug it from there.

Thanks for the feedback. I’ve made the changes you suggested and done another manual test. Still getting the same error in the test suite though so any thoughts on that appreciated! I can see the `Running the docker-setup hook...` text included in the output when running things, so I presume it must be something around the test setup but I don’t know what :thinking:

tsvallender:
Just done a bit more investigation and it seems like in the test environment the hook isn’t running because `Running the docker-setup hook...KAMAL.hook.hook_exists?(hook)` is returning `false` :thinking:

tsvallender:
Hm, made a bit more progress. Realised I needed to stub `hook_exists?`, which gets us past that error. The error I have now seems to suggest to me my `hook_variables` values are wrong, but I can’t see how. Maybe I’m barking up the wrong tree.

```
Failure:
CliServerTest#test_bootstrap_install_as_root_user [/home/tsv/code/kamal/lib/kamal/cli/base.rb:149]:
unexpected invocation: #<AnyInstance:SSHKit::Backend::Abstract>.execute(".kamal/hooks/docker-setup", {:env => {"KAMAL_RECORDED_AT" => "2023-11-19T11:00:39Z", "KAMAL_PERFORMER" => "tsv", "KAMAL
_VERSION" => "999", "KAMAL_SERVICE_VERSION" => "app@999", "KAMAL_HOSTS" => "1.1.1.1,1.1.1.2,1.1.1.3,1.1.1.4", "KAMAL_COMMAND" => "server", "KAMAL_SUBCOMMAND" => "bootstrap"}})
satisfied expectations:
- expected at least once, invoked 4 times: #<AnyInstance:SSHKit::Backend::Abstract>.execute(:mkdir, "-p", ".kamal")
- expected at least once, invoked 4 times: #<AnyInstance:SSHKit::Backend::Abstract>.execute(:curl, "-fsSL", "https://get.docker.com", "|", :sh)
- expected at least once, invoked 4 times: #<AnyInstance:SSHKit::Backend::Abstract>.execute("[ \"${EUID:-$(id -u)}\" -eq 0 ] || command -v sudo >/dev/null || command -v su >/dev/null", :raise_
on_non_zero_exit => false)
- expected at least once, invoked 4 times: #<AnyInstance:SSHKit::Backend::Abstract>.execute(:docker, "-v", :raise_on_non_zero_exit => false)
- allowed any number of times, invoked once: #<AnyInstance:Kamal::Commands::Hook>.hook_exists?(any_parameters)
```

99linesofcode:
I'm a fan. If this is accepted please make sure to update the documention on hooks accordingly.

tsvallender:
Just came back to this and managed to get the tests passing :raised_hands: 

Hopefully it can progress now—if there’s anything I can do to make that happen let me know.

test/cli/server_test.rb
```diff
@@ -23,10 +23,14 @@ class CliServerTest < CliTestCase
     SSHKit::Backend::Abstract.any_instance.expects(:execute).with('[ "${EUID:-$(id -u)}" -eq 0 ] || command -v sudo >/dev/null || command -v su >/dev/null', raise_on_non_zero_exit: false).returns(true).at_least_once
     SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:curl, "-fsSL", "https://get.docker.com", "|", :sh).at_least_once
     SSHKit::Backend::Abstract.any_instance.expects(:execute).with(:mkdir, "-p", ".kamal").returns("").at_least_once
+    Kamal::Commands::Hook.any_instance.stubs(:hook_exists?).returns(true)
+    hook_variables = { version: 999, service_version: "app@999", hosts: "1.1.1.1,1.1.1.2,1.1.1.3,1.1.1.4", command: "server", subcommand: "bootstrap" }
```
Sija, line 27: This variable seems unused?

tsvallender: Good catch thanks, was left over from a previous attempt at getting tests passing. I’ve removed it.

djmb (approved):

fredrikburmester:
The kamal remove command does not remove the network created with this command. Is this the expected behaviour?

tsvallender:
> The kamal remove command does not remove the network created with this command. Is this the expected behaviour? 

Good question. Debatable, I suppose. I guess adding _another_ hook at the point of removal might be the best way to do this. I can take a look at adding this in soon.

PR #590, opened 2023-11-19
refactor: make config.primary_role a Role instead of a string

yoelcabo:
Small refactor discussed here: https://github.com/basecamp/kamal/pull/580#discussion_r1395344471


yoelcabo:
It seems this is a bit more involved than I was expecting. Moving back to draft and I'll take a look once I get a chance :relaxed:

djmb:
I've implemented this here - https://github.com/basecamp/kamal/pull/715

PR #593, opened 2023-11-22
Role specific logging configuration

nicklozon:
# Problem
Logging configuration is shared across all containers which is problematic if containers require specific configuration. Example: Sending logs to specific log groups when using `awslogs` driver.

# Solution
Add `logging` to the role configuration.

# Examples
## Basic
```yml
servers:
  web:
    hosts: ['1.1.1.1']
    logging:
      driver: local
      options:
         max-size: 100M
  job:
    hosts: ['1.1.1.2']
    logging:
      driver: local
      options:
         max-size: 10M
```
## Deep Merge
I used `deep_merge` so the configuration would be inherited from the root configuration.
```yml
servers:
  web:
    hosts: ['1.1.1.1']
    logging:
      options:
         max-size: 100M

logging:
  driver: local
```

### DIFF
--- a/lib/kamal/configuration/role.rb
+++ b/lib/kamal/configuration/role.rb
@@ -38,7 +38,7 @@ class Kamal::Configuration::Role
 
   def logging_args
     args = config.logging || {}
-    args.deep_merge!(specializations['logging']) if specializations['logging'].present?
+    args.deep_merge!(specializations["logging"]) if specializations["logging"].present?
 
     if args.any?
       optionize({ "log-driver" => args["driver"] }.compact) +
--- a/lib/kamal/configuration/role.rb
+++ b/lib/kamal/configuration/role.rb
@@ -36,6 +36,18 @@ class Kamal::Configuration::Role
     argumentize "--label", labels
   end
 
+  def logging_args
+    args = config.logging || {}
+    args.deep_merge!(specializations['logging']) if specializations['logging'].present?
+
+    if args.any?
+      optionize({ "log-driver" => args["driver"] }.compact) +
+        argumentize("--log-opt", args["options"])
+    else
+      config.logging_args
+    end
+  end
+
 
   def env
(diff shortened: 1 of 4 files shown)


nicklozon:
Please let me know if there is an existing way to accomplish this, and I'll close the PR.

nicklozon:
If this PR is desirable, it then begs some questions. Should this also be done on accessories? Should this also be done for volumes?

If so, it'd stand to reason to abstract out container based configuration. I started a branch with the ability to configure volumes and logging at the role and accessory level: https://github.com/CleverFew/kamal/compare/role_logging_config...CleverFew:kamal:container_abstract_config?expand=1#diff-8c8703b4a72e6267819d3a1ecf329703f898343085f98e48402ab3f465af5a52

igor-alexandrov:
Great job!

lib/kamal/configuration/role.rb
```diff
@@ -36,6 +36,18 @@ def label_args
     argumentize "--label", labels
   end
 
+  def logging_args
+    args = config.logging || {}
+    args.deep_merge!(specializations['logging']) if specializations['logging'].present?
```
djmb, line 41: Could you use double quotes here please?

djmb (changes requested):

nicklozon:
Rebased and feedback addressed.

In the spirit of this PR I started a branch that abstracts out the configuration inheritance into a generic `Kamal::Configuration::ContainerBase` class. I never opened a PR because I wanted to garner feedback first - if you have any spare capacity to [review it quicky](https://github.com/CleverFew/kamal/compare/role_logging_config...CleverFew:kamal:container_abstract_config#diff-8c8703b4a72e6267819d3a1ecf329703f898343085f98e48402ab3f465af5a52) and provide a yay or nay on the concept, that'd be awesome!

djmb (approved):

PR #608, opened 2023-12-12
Accessory CLI respects `--hosts`

nicklozon:
# Problem
Accessory CLI does not respect the `--hosts` parameter - all hosts are pulled from the accessory configuration.

I.e.
- `kamal accessory logs redis --hosts 1.1.1.2`
- `kamal accessory boot all --hosts 1.1.1.3` - this is my use case when scaling up our app

# Solution
This PR intersects the accessory hosts configuration with the specified `--hosts`.


djmb:
Confirmed the tests pass locally 👍 (Github removes pending workflows after 30 days)

PR #609, opened 2023-12-13
Allow env vars to be passed in from the host.

trobrock:
We are creating our VMs with our DATABASE_URL, etc already configured as an environment variable on the VM.

This change allows us to set:
```yaml
env:
  host:
    - DATABASE_URL
```
And when the container loads the environment file it will load the DATABASE_URL from the host machine.

### DIFF
--- a/lib/kamal/env_file.rb
+++ b/lib/kamal/env_file.rb
@@ -3,7 +3,7 @@ class Kamal::EnvFile
   def initialize(env)
     @env = env
   end
-  
+
   def to_s
     env_file = StringIO.new.tap do |contents|
       if (secrets = @env["secret"]).present?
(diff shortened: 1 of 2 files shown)


djmb:
Thanks for the PR @trobrock. I'm going to close though as you can do this now with:

```
env:
  DATABASE_URL: "${DATABASE_URL}"
```

See https://kamal-deploy.org/docs/configuration/environment-variables/#injecting-env-variables-from-the-host

trobrock:
@djmb I don't think I was clear enough on this, host is not a clear enough word. These env vars are not from the machine I am deploying from it is from the machine that is running docker.

djmb:
Hey @trobrock - sorry maybe the docs aren't clear enough either! That's also what `DATABASE_URL: "${DATABASE_URL}"` will do 👍

trobrock:
Alright, I just tested this:

Here is my role config:
```
  worker:
    hosts:
      - 0.0.0.0
    cmd: ./bin/rake solid_queue:start
    env:
      QUEUES: default
      DOCKER_HOSTNAME: "${HOSTNAME}"
```

On the server:
```
user@comfortly-worker-1:~# echo $HOSTNAME
comfortly-worker-1
```

In the docker container:
```
inn-management(prod)> ENV['DOCKER_HOSTNAME']
=> "${HOSTNAME}"
```

PR #614, opened 2023-12-17
Trim long hostnames and remove leading hyphens

kryachkov:
Encountered this issue on Ubuntu 22.04.3 and docker version:

```
docker version
Client: Docker Engine - Community
 Version:           24.0.7
 API version:       1.43
 Go version:        go1.20.10
 Git commit:        afdd53b
 Built:             Thu Oct 26 09:07:41 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          24.0.7
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.10
  Git commit:       311b9ff
  Built:            Thu Oct 26 09:07:41 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.26
  GitCommit:        3dd1e886e55dd695541fdcd67420c2888645a495
 runc:
  Version:          1.1.10
  GitCommit:        v1.1.10-0-g18a0cb0
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

docker returns the following error when `--hostname` is longer than 64 characters:

```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: sethostname: invalid argument: unknown.
```

according to https://man7.org/linux/man-pages/man7/hostname.7.html it should be 63 characters max, but seems like docker allows 64. Removing leading hyphens is extra as hostname can't start with it

### DIFF
--- a/test/cli/cli_test_case.rb
+++ b/test/cli/cli_test_case.rb
@@ -38,7 +38,7 @@ class CliTestCase < ActiveSupport::TestCase
         .with { |arg1, arg2| arg1 == :rm && arg2 == ".kamal/locks/app/details" }
     end
 
-    def assert_hook_ran(hook, output, version:, service_version:, hosts:, command:, subcommand: nil, runtime: nil)
+    def assert_hook_ran(hook, output, version:, service_version:, hosts:, command:, subcommand: nil, runtime: false)
       performer = `whoami`.strip
 
       assert_match "Running the #{hook} hook...\n", output
@@ -52,7 +52,7 @@ class CliTestCase < ActiveSupport::TestCase
         KAMAL_HOSTS=\"#{hosts}\"\s
         KAMAL_COMMAND=\"#{command}\"\s
         #{"KAMAL_SUBCOMMAND=\\\"#{subcommand}\\\"\\s" if subcommand}
-        #{"KAMAL_RUNTIME=\\\"#{runtime}\\\"\\s" if runtime}
+        #{"KAMAL_RUNTIME=\\\"\\d+\\\"\\s" if runtime}
         ;\s/usr/bin/env\s\.kamal/hooks/#{hook} }x
 
       assert_match expected, output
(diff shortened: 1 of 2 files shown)


lib/kamal/cli/app.rb
```diff
@@ -42,7 +42,9 @@ def boot
 
               execute *auditor.record("Booted app version #{version}"), verbosity: :debug
 
-              execute *app.run(hostname: "#{host}-#{SecureRandom.hex(6)}")
+              hostname = "#{host}-#{SecureRandom.hex(6)}"
+
+              execute *app.run(hostname: (hostname[-64..-1] || hostname).gsub(/^-+/, ''))
```
djmb, line 47: Can we truncate from the end of the hostname instead? That means that we no longer need to look for `-`s as they are allowed on the end of a hostname. We'll want to strip any trailing `.`s though.

Also I think we should extract a method here to convert `host` to ` hostname.

kryachkov: We can truncate from the end of the string, but then we might be truncating the random part that is added in `SecureRandom.hex(6)`, unless you suggest to truncate the host part and keep 12 random characters appended. WDYT?

Sija: Random part seems less relevant, since you'd mostly identify service by the name, I think.

djmb: @kryachkov - yes sorry I meant truncate first, then append the random string!

bvrxs2yyw9fq49n5: > @kryachkov - yes sorry I meant truncate first, then append the random string!

Working on a fix! Thanks

kryachkov: I accidentally messed up this PR after rebasing my fork my source, so I had to open a new PR - https://github.com/basecamp/kamal/pull/762

PR #615, opened 2023-12-19
Remove accessory env file from host, when accessory is removed

leonvogt:
I was surprised that `kamal accessory remove [name]` does not delete the .env file in the `~/.kamal/env/accessories` directory.      
This PR would add that functionality.


leonvogt:
Closing since its no longer accurate with Kamal 2. Kamal 2 seems to no longer have a concept of `remove_env_file` since https://github.com/basecamp/kamal/commit/56754fe40cff73aa62ce3c6d21b7b55ddd037986

PR #619, opened 2023-12-26
Fix redeploy command missing registry authentication

tonysm:
### Fixed

- Fixes the redeploy command missing registry authentication

### DIFF
--- a/lib/kamal/cli/main.rb
+++ b/lib/kamal/cli/main.rb
@@ -63,6 +63,9 @@ class Kamal::Cli::Main < Kamal::Cli::Base
       mutating do
         invoke_options = deploy_options
 
+        say "Log into image registry...", :magenta
+        invoke "kamal:cli:registry:login", [], invoke_options
+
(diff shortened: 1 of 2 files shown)


dhh:
This is by design. Point of redeploy over deploy is to do fewer steps that aren't necessary when repeatedly deploying.

tonysm:
@dhh hmm, this is failing for me because it's trying to build and push the image to a private registry. It only works if I use the `--skip-push` flag. Should the redeploy command default to only pulling then?

dhh:
If you’ve deployed once on that machine, you should already be logged into the registry. Then you can do redeploy from there. If that’s not working for some reason, you can also just use deploy every time.

tonysm:
Yeah, I think the issue is because I'm using the alias install, so it runs in a container. The docker client there isn't authenticated. But that's okay, I'll use the deploy cmd or redeploy with the `--skip-push` flag if I need a config change.

PR #620, opened 2023-12-26
Allow custom user and port for builder host

mlitwiniuk:
When `ssh` options are set, they overwrite custom username and password passed via ssh builder uri. Passing part of uri for sshkit is fine, as it then properly extracts username and password and forwards it as `host.ssh_options` (in which case it's no longer empty)

### DIFF
--- a/lib/kamal/cli/build.rb
+++ b/lib/kamal/cli/build.rb
@@ -114,8 +114,11 @@ class Kamal::Cli::Build < Kamal::Cli::Base
     def connect_to_remote_host(remote_host)
       remote_uri = URI.parse(remote_host)
       if remote_uri.scheme == "ssh"
-        options = { user: remote_uri.user, port: remote_uri.port }.compact
-        on(remote_uri.host, options) do
+        host = SSHKit::Host.new(
+          hostname: remote_uri.host,
+          ssh_options: { user: remote_uri.user, port: remote_uri.port }.compact
+        )
+        on(host, options) do
           execute "true"
         end
       end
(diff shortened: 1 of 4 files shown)


lib/kamal/cli/build.rb
```diff
@@ -115,7 +115,7 @@ def connect_to_remote_host(remote_host)
       remote_uri = URI.parse(remote_host)
       if remote_uri.scheme == "ssh"
         options = { user: remote_uri.user, port: remote_uri.port }.compact
-        on(remote_uri.host, options) do
+        on(remote_host.gsub('ssh://', ''), options) do
```
djmb, line 118: I think we could do:

```ruby
host = SSHKit::Host.new(
  hostname: remote_uri.host,
  ssh_options: { user: remote_uri.user, port: remote_uri.port }
)
on(host) do
  execute "true"
end
```

which would avoid needing to alter the options in sshkit_with_ext.rb

mlitwiniuk: Good catch, I've updated PR accordingly

mlitwiniuk:
Yep, you're right. Closing

djmb:
@mlitwiniuk - I was just referring to this line when I said we don't need it anymore. I think the rest of the PR is still valid!

```
host.ssh_options = (host.ssh_options || {}).merge({ port: host.port, user: host.user }.compact)
```

djmb:
Going to re-open as I think we want this change!

laptopmutia:
is it possible to add custom IdentityFile too ?

Issue #626, opened 2024-01-04
Sprockets choose wrong manifest with asset_path option used


aishek:
When using the `asset_path` option, Kamal copies all files from the assets folder to the brand-new deployed container. If the app inside the container uses Sprockets gem with default settings, then in the assets folder it has two manifest files. The app chooses [random one](https://github.com/rails/sprockets/blob/main/lib/sprockets/manifest_utils.rb#L44), unless it is [specified](https://github.com/rails/sprockets-rails/blob/master/lib/sprockets/railtie.rb#L215) explicitly.

From the end-user perspective it looks like no new assets were deployed.

The fix is simple:

```ruby
# config/environments/production.rb

config.assets.manifest = Rails.root.join('config', 'manifest.json')
```

I am not sure if the issue is worth mentioning in `config/deploy.yml` comments. At least I've posted it here.

milk1000cc:
I encountered the same problem. Thanks for the solution.

PR #627, opened 2024-01-04
Mention Sprockets config in deploy template

aishek:
This is the fix for https://github.com/basecamp/kamal/issues/626. I've added the hint to the deploy.yml template to save time for people who using Sprockets with Kamal.

I was also considering adding a message on gem post-install. But as for me, a comment is more noticeable.

### DIFF
--- a/lib/kamal/cli/templates/deploy.yml
+++ b/lib/kamal/cli/templates/deploy.yml
@@ -77,6 +77,10 @@ registry:
 # Bridge fingerprinted assets, like JS and CSS, between versions to avoid
 # hitting 404 on in-flight requests. Combines all files from new and old
 # version inside the asset_path.
+#
+# If your app is using Sprockets gem, ensure it has `config.assets.manifest`
+# set. See https://github.com/basecamp/kamal/issues/626 for details
+#
 # asset_path: /rails/public/assets
 
 # Configure rolling deploys by setting a wait time between batches of restarts.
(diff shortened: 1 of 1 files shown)


lib/kamal/cli/templates/deploy.yml
```diff
@@ -77,6 +77,10 @@ registry:
 # Bridge fingerprinted assets, like JS and CSS, between versions to avoid
 # hitting 404 on in-flight requests. Combines all files from new and old
 # version inside the asset_path.
+#
+# If your app is using Sprockets gem, ensure it has `config.assets.manifest`
+# set. See https://github.com/basecamp/kamal/issues/626 for details
```
djmb, line 82:
```suggestion
# If your app is using the Sprockets gem, ensure it sets `config.assets.manifest`.
# See https://github.com/basecamp/kamal/issues/626 for details
```

dferrazm:
@dhh I recently stumbled on the issue related to this PR and found the solution here. I noticed though that you [removed the comment](https://github.com/basecamp/kamal/commit/b6cd4f8070dfc50ec8e252c4bb9225cf838a09fe) that this PR added to the `deploy.yml` file hinting people when still using Sprockets.

Was that intentional? I think that comment would still be useful.

dhh:
Sprockets is legacy for Rails. Kamal's default should target what's current (i.e. Propshaft). AI is going to find the issue when you ask anyway 😄

dferrazm:
> AI is going to find the issue when you ask anyway

I tried with Cursor and it didn't find anything though. Just vague replies. 😬 😅  I had to dig it through.

aishek:
You have to give more context to AI, so this is fine. At least we have the thread, let's hope AI assistance will pay more attention to it 🤞

PR #656, opened 2024-01-22
Informative message on lock error

DanielJackson-Oslo:
_Edit: I originally proposed linking to the documentation on the website, but @adrienpoly  and @acidtib suggested referring to `kamal lock help` instead. I've updated the description to match._ 

Recently I aborted a kamal deploy with `cmd-c` and upon trying to deploy again, I got the lock error message:

`Deploy lock found.` 

Unfortunately it took some googling to find out what this meant, and that I had to release the lock manually to proceed. 

To mitigate others having the same experience, this PR does two things: 

1) Adds "Run 'kamal lock help' for more information" to the error message, giving the user some pointers on what to do about this situation.
2) Adds a header in red over the lock status message that is already printed, to make it clearer that this information is relevant to the error message you just received. (I didn't understand that at first glance, so thinking others might not either).

Put together this is how it looks in use:

Current, without this PR:
image

Suggested, with this PR:
image

Do you think this could work and be helpful? If not, there might be other ways to make the experience of locking oneself out more intuitive.

### DIFF
--- a/lib/kamal/cli/base.rb
+++ b/lib/kamal/cli/base.rb
@@ -123,6 +123,7 @@ module Kamal::Cli
         yield
       rescue SSHKit::Runner::ExecuteError => e
         if e.message =~ /cannot create directory/
+          say "Deploy lock already in place!", :red
           on(KAMAL.primary_host) { puts capture_with_debug(*KAMAL.lock.status) }
           raise LockError, "Deploy lock found. Run 'kamal lock help' for more information"
         else
--- a/lib/kamal/cli/base.rb
+++ b/lib/kamal/cli/base.rb
@@ -124,7 +124,7 @@ module Kamal::Cli
       rescue SSHKit::Runner::ExecuteError => e
         if e.message =~ /cannot create directory/
           on(KAMAL.primary_host) { puts capture_with_debug(*KAMAL.lock.status) }
-          raise LockError, "Deploy lock found. See https://kamal-deploy.org/docs/commands#checking-and-setting-the-lock"
+          raise LockError, "Deploy lock found. Run 'kamal lock help' for more information"
         else
           raise e
         end
--- a/lib/kamal/cli/base.rb
+++ b/lib/kamal/cli/base.rb
@@ -124,7 +124,7 @@ module Kamal::Cli
       rescue SSHKit::Runner::ExecuteError => e
         if e.message =~ /cannot create directory/
           on(KAMAL.primary_host) { puts capture_with_debug(*KAMAL.lock.status) }
-          raise LockError, "Deploy lock found"
+          raise LockError, "Deploy lock found. See https://kamal-deploy.org/docs/commands#checking-and-setting-the-lock"
         else
           raise e
         end
--- a/test/integration/lock_test.rb
+++ b/test/integration/lock_test.rb
@@ -10,7 +10,7 @@ class LockTest < IntegrationTest
     assert_match /Locked by: Deployer at .*\nVersion: #{latest_app_version}\nMessage: Integration Tests/m, status
 
     error = kamal :deploy, capture: true, raise_on_error: false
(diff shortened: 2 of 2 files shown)


adrienpoly:
What about suggesting the cli help command 

```
kamal lock help
```

it will display 
```
Commands:
  kamal lock acquire -m, --message=MESSAGE  # Acquire the deploy lock
  kamal lock help [COMMAND]                 # Describe subcommands or one specific subcommand
  kamal lock release                        # Release the deploy lock
  kamal lock status                         # Report lock status

Options:
  -v, [--verbose], [--no-verbose]        # Detailed logging
  -q, [--quiet], [--no-quiet]            # Minimal logging
      [--version=VERSION]                # Run commands against a specific app version
  -p, [--primary], [--no-primary]        # Run commands only on primary host instead of all
  -h, [--hosts=HOSTS]                    # Run commands on these hosts instead of all (separate by comma, supports wildcards with *)
  -r, [--roles=ROLES]                    # Run commands on these roles instead of all (separate by comma, supports wildcards with *)
  -c, [--config-file=CONFIG_FILE]        # Path to config file
                                         # Default: config/deploy.yml
  -d, [--destination=DESTINATION]        # Specify destination to be used for config file (staging -> deploy.staging.yml)
  -H, [--skip-hooks], [--no-skip-hooks]  # Don't run hooks
                                         # Default: false
```

I see 2 advantages:
1. we remain in the command line
2. this message is not linked to an url that could change in the future

DanielJackson-Oslo:
If you think that would work, I can make changes here.

I also think we can make the current running of `kamal lock status` just before the error message clearer, by moving the error message text we are discussing here up above the kamal lock status text as well, as a common header for both the status and error: 

Here is the complete error today:

```
Acquiring the deploy lock...
Locked by: Daniel Jackson at 2024-01-22T09:46:34Z
Version: b22c9c1d8ba883a6c34f835c9dbd2802e06244ae
Message: Automatic deploy lock
  Finished all in 3.0 seconds
  ERROR (Kamal::Cli::LockError): Deploy lock found
```

The new error could read something like this:

```
Acquiring the deploy lock...

ERROR: Deploy lock found. Cannot deploy. Run 'kamal lock help'  for more information.

Locked by: Daniel Jackson at 2024-01-22T09:46:34Z
Version: b22c9c1d8ba883a6c34f835c9dbd2802e06244ae
Message: Automatic deploy lock

  Finished all in 3.0 seconds
  ERROR (Kamal::Cli::LockError): Deploy lock found. 
```

acidtib:
I agree with @adrienpoly If there's a need for modification, let's confine it within the CLI for a more stable experience. Redirecting users to a URL runs the risk of becoming obsolete with subsequent version releases. 

The current error messages are quite lucid, and there's comprehensive documentation available on [kamal-deploy.org](https://kamal-deploy.org/), GitHub issues, and Discord server. 

Maintaining clarity within the CLI would ensure consistency and avoid potential pitfalls associated with external URLs that may change over time.

DanielJackson-Oslo:
Ok, I made two new commits. The first just replaces the URL with the CLI help command, as suggested. 

The second also adds a little helpful header in red over the lock status message, to make it clearer that this information is relevant to the error message you just received. (I didn't understand that at first glance, so thinking others might not either).

If you don't want the second commit with the header in red, then I'll just revert that one. 

Put together this is how it looks in use:

Current:
image

Suggested:
image

PS: I didn't find a way to add a description of locking on top of the help message like I wanted to here https://github.com/basecamp/kamal/pull/656#issuecomment-1903627648 - but I've never used Thor before and might have missed something.

DanielJackson-Oslo:
I've updated the description to reflect the state of the code. Hope this PR is helpful to the project! Let me know if there's any further changes you would like done.

djmb:
Thanks for the PR @DanielJackson-Oslo!

PR #666, opened 2024-02-01
Set KAMAL_DESTINATION for envify command

nickhammond:
Since destinations usually map to an environment such as a Rails environment it'd be helpful to load the current destination so that you can more easily use a destination template file. I made the assumption that we load the clear variables from Kamal but right now envify just loads the template and writes the output destination file, no clear/secret loading from Kamal. 

For example:

.env.template.erb

```
# Generated by kamal envify, to regenerate run `kamal envify -d <%= ENV['KAMAL_DESTINATION'] %>`

RAILS_ENV=<%= ENV['KAMAL_DESTINATION'] %>

API_KEY_EXAMPLE=<%= `op read "op://hey-#{ENV['KAMAL_DESTINATION']}/API_KEY_EXAMPLE/credential" -n`.strip %>
```

.env.staging.erb and .env.production.erb
```
<%= 
  ERB.new(File.read(".env.template.erb"), trim_mode: "-").result(binding)
%>
```

Which would generate something like this when running `kamal envify -d staging`:

```
# Generated by kamal envify, to regenerate run `kamal envify -d staging`

RAILS_ENV=staging

API_KEY_EXAMPLE=secret
```

#352 is related from the perspective of having a template for your destination files.

### DIFF
--- a/lib/kamal/cli/main.rb
+++ b/lib/kamal/cli/main.rb
@@ -175,6 +175,7 @@ class Kamal::Cli::Main < Kamal::Cli::Base
   option :skip_push, aliases: "-P", type: :boolean, default: false, desc: "Skip .env file push"
   def envify
     if destination = options[:destination]
+      ENV['KAMAL_DESTINATION'] = destination
       env_template_path = ".env.#{destination}.erb"
       env_path          = ".env.#{destination}"
     else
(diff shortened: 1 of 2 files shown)


nickhammond:
@igor-alexandrov `ENV['KAMAL_DESTINATION']` is only populated in the hooks at the moment I believe. But yes, I'm doing something similar to your .env.staging.erb example.

jeromedalbert:
Wondering if it would make sense to make `ENV['KAMAL_DESTINATION']` available more globally, not just for `envify`. I have found the need to use `ENV['KAMAL_DESTINATION']` in my `deploy.yml` so I don't have to hardcode the word "production" in it.

PR #692, opened 2024-02-22
Add a simple validation to the service name to prevent setup issues

### LINKED ISSUE #671: Getting Bad Gateway on a fresh setup. When using dot in my service name.
I was trying to set up a couple of instances with Kamal and was constantly getting Bad Gateway on all of them. I tried redeploying and spun up fresh ones as well.

With health checks passing, the `setup` and `deploy` commands were successful. 

I did some digging: 
- both containers are up
- docker inspect looked fine(to me).
- I even SSH'd into the `app` and `traefik` containers on the instance to make sure they could communicate with each other.
- when I tried `curl` from the `traefik` container, it worked without any issues.

**Traefik logs:** 
```bash
2024-02-07T17:09:17.248587866Z time="2024-02-07T17:09:17Z" level=debug msg="'502 Bad Gateway' caused by: dial tcp 172.17.0.2:0: connect: connection refused"
2024-02-07T17:12:43.194810142Z time="2024-02-07T17:12:43Z" level=debug msg="'502 Bad Gateway' caused by: dial tcp 172.17.0.2:0: connect: connection refused"
```

**~Solution**
I used a dot in the service name, like `app7.1`, which seemed harmless at first glance.
To fix it, I simply removed the dot from the service name, making it `app71`, and everything started working smoothly.
Just to be clear it's `config/deploy.yml:`
```yml
# Name of your application. Used to uniquely configure containers.
service: app71

```
**My theory** is that, although I'm not an expert in Docker routing and Traefik, the issue might be related to how the config files are named after the service name and Traefik/Docker labels. It seems like they could be causing some unintended issues.

Can someone confirm/deny it? If this is the case we can add it to documentation.

nickhammond:
Including dots in your service name is going to cause issues with traefik and initial Kamal setup. 

This PR adds a simple regex to validate that the service name is alphanumeric and can include hyphens `/^[a-z0-9-]+$/`.

Feels a bit like overkill but just trying to ease everyone's first impression of Kamal to make it easier to get up and running. 

Closes #671.

### DIFF
--- a/lib/kamal/configuration.rb
+++ b/lib/kamal/configuration.rb
@@ -284,7 +284,7 @@ class Kamal::Configuration
     end
 
     def ensure_valid_service_name
-      raise ArgumentError, "Service name can only include alphanumeric characters and hyphens" unless raw_config[:service] =~ /^[a-z0-9-]+$/
+      raise ArgumentError, "Service name can only include alphanumeric characters, hyphens, and underscores" unless raw_config[:service] =~ /^[a-z0-9-_]+$/
 
       true
     end
--- a/lib/kamal/configuration.rb
+++ b/lib/kamal/configuration.rb
@@ -218,7 +218,7 @@ class Kamal::Configuration
 
 
   def valid?
-    ensure_destination_if_required && ensure_required_keys_present && ensure_valid_kamal_version
+    ensure_destination_if_required && ensure_required_keys_present && ensure_valid_kamal_version && ensure_valid_service_name
   end
 
   def to_h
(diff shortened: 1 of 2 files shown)


lib/kamal/configuration.rb
```diff
@@ -283,6 +283,12 @@ def ensure_required_keys_present
       true
     end
 
+    def ensure_valid_service_name
+      raise ArgumentError, "Service name can only include alphanumeric characters and hyphens" unless raw_config[:service] =~ /^[a-z0-9-]+$/
```
djmb, line 287: I think we should also allow `_`.

nickhammond: Good call, added.

djmb:
@nickhammond - I'm seeing this warning now, I guess we want to reverse the order of the `-` and `_`?

```
.../kamal/lib/kamal/configuration.rb:291: warning: character class has '-' without escape: /^[a-z0-9-_]+$/
```

nickhammond:
@djmb Fixed in #708, sorry about that.

PR #697, opened 2024-03-01
configure volumes, directories and files for any role

aka47:
For the app, the traefic role, and any accessory you can configure files, volumes, and directories from your project to be accessible inside the docker container.

The files and directories are uploaded to the host and mounted via docker volumes to the docker container.
This worked previously only for accessories. So I moved code related to this from accessory configuration to a new class to make it useable for the traefik proxy and the main app as well.

With this change we can add in the config for traefik volumes, files and folders

```
traefik:
  files:
    - config/traefik.yaml:/etc/traefik/traefik.yaml
  directories:
    - config/traefik.d:/etc/traefik/traefik.d
  volumes:
    - /mnt/volume-data/letsencrypt/acme.json:/letsencrypt/acme.json
```

For the main app config it looks like this:

```
volumes:
  - "/mnt/volume-data/wordpress-data:/var/www/html"
files:
  - apache.conf:/etc/apache2/sites-enabled/wp3.conf
directories:
  - config/apache/:/etc/conf-avail
```

And it works as it currently does for accessories.
```
accessories:
  redis:
    files:
      - config/redis.conf:/etc/redis/redis.conf
    directories:
      - config/conf.d:/etc/conf/conf.d
    volumes:
      - /mnt/volume-data/conf/conf.json:/conf/conf.json
```

### DIFF
--- a/lib/kamal/cli/accessory.rb
+++ b/lib/kamal/cli/accessory.rb
@@ -7,29 +7,12 @@ class Kamal::Cli::Accessory < Kamal::Cli::Base
       else
         with_accessory(name) do |accessory|
           directories(name)
-          upload(name)
+          upload(accessory, accessory.hosts)
 
           on(accessory.hosts) do
             execute *KAMAL.registry.login if login
             execute *KAMAL.auditor.record("Booted #{name} accessory"), verbosity: :debug
             execute *accessory.run
-          end
-        end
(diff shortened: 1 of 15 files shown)


djmb:
Closing as we no longer use Traefik in Kamal 2

Issue #524, opened 2023-10-09
Ensure cleanup file system before `docker build`. Opinion needed.


igor-alexandrov:
I opened PR that adds `vendor/bundle` to `.dockerignore` file template in Rails repo: https://github.com/rails/rails/pull/48540.

However, @byroot correctly mentioned that `vendor/bundle` can be used by some applications, and maybe it is not a good idea to add it to `.dockerignore` and `.gitignore`. Instead, he proposed that Kamal ensure a clean file system before `docker build`.

I want to hear your opinion on this case and find the best solution.

byroot:
To clarify, I think beyond this specific `vendor/bundle` issue that is happening to you, building from a directory that may contain local modifications that weren't committed is risky, as you may accidentally include things you didn't intent do.

It may make sense for kamal to build from a temporary directory created with `git archive` or similar.

I'm really unfamiliar with kamal though so not too sure.

PR #700, opened 2024-03-05
Build from a git archive

djmb:
Building directly from a checkout will pull in uncommitted files, or more sneakily, files that are git ignored, but not docker ignored.

To avoid this, we'll add an option to build from a git archive of HEAD instead. Docker doesn't provide a way to build directly from a git repo, so instead we create a tarball of the current HEAD with git archive and pipe it into the build command.

When building from a git archive, we'll still display the warning about uncommitted changes, but we won't add the `_uncommitted_...` suffix to the container name as they won't be included in the build.

~Perhaps this should be the default, but we'll leave that decision for now.~

Actually after some thought, I think we should make this the default if no build context is set, then we don't need a new option, and if someone wants to retain the existing behaviour they can set the context to `.`.

Fixes: https://github.com/basecamp/kamal/issues/524 (and H/T @byroot for the suggestion).

### DIFF
--- a/lib/kamal/cli/templates/deploy.yml
+++ b/lib/kamal/cli/templates/deploy.yml
@@ -31,6 +31,7 @@ registry:
 #   user: app
 
 # Configure builder setup.
+# Set git_archive: true to build from a git archive of HEAD
 # builder:
 #   args:
 #     RUBY_VERSION: 3.2.0
@@ -39,6 +40,7 @@ registry:
 #   remote:
 #     arch: amd64
 #     host: ssh://app@192.168.0.1
+#   git_archive: false
 
 # Use accessory services (secrets come from .env).
 # accessories:
--- a/lib/kamal/cli/templates/deploy.yml
+++ b/lib/kamal/cli/templates/deploy.yml
@@ -31,7 +31,6 @@ registry:
 #   user: app
 
 # Configure builder setup.
-# Set git_archive: true to build from a git archive of HEAD
 # builder:
 #   args:
 #     RUBY_VERSION: 3.2.0
@@ -40,7 +39,6 @@ registry:
 #   remote:
 #     arch: amd64
 #     host: ssh://app@192.168.0.1
-#   git_archive: false
 
 # Use accessory services (secrets come from .env).
 # accessories:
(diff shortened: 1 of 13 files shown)


99linesofcode:
This is a bit implicit and could be documented more clearly somehow.

voberoi:
This also broke my build. Would be great to clearly document this.

In my case I rely on a script to generate a couple of files for me pre-build, and they need to be in the builder context.

Adding this to config did what I needed:

```
builder:
  context: "."
```

jpdombrowski:
Are you aware that this leads to `.dockerignore` not being applied?

djmb:
Ah I wasn't aware no - @jpdombrowski, but you are right - see https://github.com/docker/buildx/issues/2353. Not an issue if `.dockerignore` is a subset of `.gitignore` but its probably not wise to assume that. May need to rethink this change 😬

byroot:
I don't follow. if `.dockerignore` is committed, it should end up in the archive, no?

jpdombrowski:
It ends ends up in the archive, but the build context piped in as a git archive tarball is not processed the same way as when building from local files using `docker build .`.

I just noticed my image size growing and investigated.

Personally would revert this for now, because its highly unexpected behaviour that could easily go unnoticed. That `.dockerignore` is a subset of `.gitignore` is likely rare (`.gitignore` is often included in `.dockerignore`)

byroot:
> Personally would revert this for now, because its highly unexpected behaviour that could easily go unnoticed.

I'm sure there must be a way to make the dockerignore file work. Reverting means going back to potentially building with git ignored (potentially secrets) or uncommitted changes. Which is just as bad if not worse.

djmb:
Ok, going to try https://github.com/basecamp/kamal/pull/798 instead and build from a local clone.

A bit annoying that it will leave some state around, even if it's in a tmpdir, but docker does seem to be a bit faster about generating the build context on subsequent builds from a directory so we get something out of it.

@99linesofcode, @voberoi - it will now log about using the clone directory and if there are uncommitted changes it will warn you about whether they are being included or excluded. Hopefully that's a bit better than before.

PR #702, opened 2024-03-05
Relax `dotenv` version

adam12:
Looking at the [breaking changes listed](https://github.com/bkeepers/dotenv/blob/c9ecfa4e5598593dbadcf6c610cf5d8a9ed0bec1/Changelog.md) for `dotenv`, it seems to only affect Rails applications that are directly using `dotenv` and are upgrading between versions of `dotenv`.

Presumably, if they were using `dotenv` directly, they would have the version of `dotenv` in their `Gemfile` that works for them, and we're safe to allow a newer version.


igor-alexandrov:
@adam12 `kamal` is not required to be in the application Gemfile. There is no reason to relax its dependencies.

adam12:
DHH seems to suggest it should be used globally, so I'll close this.

PR #703, opened 2024-03-06
Auto push of environment files when deploying

djmb:
#### Split env into separate secrets/clear envs
Split each env file in two on the deploy hosts, one for secrets and one for clear values. This will allow us to update them independently.

#### Add push_env setting
This setting allows you to automatically push env files when deploying.
The default is not to push any files, but you can set it to `all`, `clear` or `secret` to push the relevant files.

The most useful setting is `clear` which will push the clear env files every time you deploy.

In addition you can choose the env_type to push when calling
`kamal env push` directly:

```
kamal env push --env-type clear
kamal env push --env-type secret
kamal env push --env-type all # same as kamal env push
```

This change will require running `kamal envify` or `kamal env push` once to create the new clear and secret env files.


nickhammond:
@djmb What's the thinking around breaking env into two files or when would you want to push them independently? How does that work with `.env.erb`?

djmb:
@nickhammond - the clear variables are specified in the config, so we'd want them to generally be updated whenever we deploy. 

The secrets though are external so automatically updating them on a deployment if not safe since they may not be up to date where you are deploying from.

I think we'll probably not merge this PR as is though with the `push_env` setting and instead just have the clear env update on deployment and the secrets only on env push.

In that case we maybe don't need the clear env file at all and can specify those env variables directly in the `docker run` command like we used to.

morgoth:
@djmb That would be useful indeed, ie when deploying from CI where I don't want to put credentials like master key, etc.

djmb:
Closed in favour of https://github.com/basecamp/kamal/pull/732

PR #708, opened 2024-03-06
Remove warning for valid service name

nickhammond:
Fixes the warning from #692 that @djmb mentioned.


djmb:
Thanks @nickhammond!

PR #716, opened 2024-03-08
Support init command with destination

morgoth:


morgoth:
this gonna change in https://github.com/basecamp/kamal/pull/924/

PR #722, opened 2024-03-13
Pushing Build Should not Require Lock Acquire nor Server Setup

rience:
When we deploy feature branches - we dynamically set up new servers based on branch name. However, when one only wants to run CI build and make sure test passes - we only build Docker image with `kamal build push`. At that point servers are not set up. Of course they can be - but that'd be waste of resources since not all feature branches are deployed.

I'm not sure if there's a reason why would `kamal build push` require servers to be set up. So this PR is removing call to `mutating`.

If this is a "no go" - would you consider PR where one could be passing "—skip-server-setup" or similar param?

### DIFF
--- a/lib/kamal/cli/build.rb
+++ b/lib/kamal/cli/build.rb
@@ -13,31 +13,29 @@ class Kamal::Cli::Build < Kamal::Cli::Base
 
   desc "push", "Build and push app image to registry"
   def push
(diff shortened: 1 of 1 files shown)


djmb:
Thanks @rience! I'm closing this as it was resolved by #755

PR #732, opened 2024-03-21
Always send the clear env to the container

djmb:
Secret and clear env variables have different lifecycles. The clear ones are part of the repo, so it makes sense to always deploy them with the rest of the repo.

The secret ones are external so we can't be sure that they are up to date, therefore they require an explicit push via `envify` or `env push`.

We'll keep the env file, but now it just contains secrets. The clear values are passed directly to `docker run`.


nickhammond:
@djmb I just realized how helpful this is going to be for a few things like deploy tracking, thanks for adding this! Excited for this to get cut into a release.

djmb:
@nickhammond - yes it will be useful I think. It also means that we can pull environment variables in from the host like we were before the `env push` command was introduced

PR #744, opened 2024-03-27
Per destination latest image tag

djmb:
We tag the deployed image with the latest image tag when deploying. But if we are deploying multiple destinations to the same host they'll conflict with each other.


djmb:
Superseded by https://github.com/basecamp/kamal/pull/748

PR #770, opened 2024-04-15
Allow capital letters to match valid service name, such as in MyApp

### LINKED ISSUE #769: Unexpected: ERROR (ArgumentError): Service name can only include alphanumeric characters, hyphens, and underscores
The most recent version of Kamal includes a regex check to ensure a successful service name. I saw there was a bit of back and forth about this regex in other commits, but the current version does not allow for capital letters.

```
From: /Users/ttilberg/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/kamal-1.4.0/lib/kamal/configuration.rb @ line 295 :

    290:
    291:       true
    292:     end
    293:
    294:     def ensure_valid_service_name
 => 295:       require 'irb'; binding.irb
    296:       raise ArgumentError, "Service name can only include alphanumeric characters, hyphens, and underscores" unless raw_config[:service] =~ /^[a-z0-9_-]+$/
    297:
    298:       true
    299:     end
    300:

irb(#<Kamal::Configuration:0x0000...):001> raw_config[:service]
=> "DemandAtlas"
irb(#<Kamal::Configuration:0x0000...):002> raw_config[:service] =~ /^[a-z0-9_-]+$/
=> nil
```

If this is expected, the error message should include that. If not, the case insensitive flag should be included.

ttilberg:
Currently if your service name has capital letters, such as in `MyApp`, it triggers the new `ensure_valid_service_name` validation error. The error message states `alphanumeric characters`, but does not allow capital letters.

This change allows capital letters in the service name, such as `MyApp`.

Fixes #769

### DIFF
--- a/lib/kamal/configuration.rb
+++ b/lib/kamal/configuration.rb
@@ -301,7 +301,7 @@ class Kamal::Configuration
     end
 
     def ensure_valid_service_name
-      raise ArgumentError, "Service name can only include alphanumeric characters, hyphens, and underscores" unless raw_config[:service] =~ /^[a-z0-9_-]+$/
+      raise ArgumentError, "Service name can only include alphanumeric characters, hyphens, and underscores" unless raw_config[:service] =~ /^[a-z0-9_-]+$/i
 
       true
     end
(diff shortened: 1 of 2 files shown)


ttilberg:
Alternative to this, if you did in fact intend to disallow capital letters, the error message should be improved. However, I've had an app deployed for some time now called `DemandAtlas` that has not previously had issues with Traefik.

PR #772, opened 2024-04-17
Take accessory hosts into account for --hosts

aishek:
```yml
servers:
  web:
    host: 1.2.3.5

accessories:
  service:
    host: 1.2.3.4
```

```bash
$ kamal accessory details service -h 1.2.3.4
  ERROR (ArgumentError): No --hosts match for 1.2.3.4
```

This happens because an accessory host is not included in all hosts. The PR fixes the issue.

### DIFF
--- a/lib/kamal/configuration.rb
+++ b/lib/kamal/configuration.rb
@@ -88,7 +88,7 @@ class Kamal::Configuration
 
 
   def all_hosts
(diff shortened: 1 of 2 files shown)


aishek:
See also https://github.com/basecamp/kamal/pull/608

Issue #775, opened 2024-04-18
uncommitted tag messes with remote builds


swombat:
It's not at all obvious that if you have uncommitted changes while waiting several minutes for the build to finish (which is pretty likely), then Kamal will tag the build with a random string in `configuration.rb`:

```
          [ Kamal::Git.revision, Kamal::Git.uncommitted_changes.present? ? "_uncommitted_#{SecureRandom.hex(8)}" : "" ].join
```

I have no idea what this achieves - perhaps something to help with CI - but it makes deployment more painful because it means I either have to dev in a separate repo on my own machine, or down tools while Kamal does its thing. Otherwise, the delivered build has a tag that `kamal deploy -P` can't find (it generates another new unique tag).

I would suggest making this a configurable option, starting with defaulting to the current behaviour (so as not to break people's builds) with clear documentation about turning this off unless you need it, and eventually making the default what will be helpful for most people who are setting up Kamal (assuming most people don't need these tags).

swombat:
Added a pull request... not sure if this should have been done here instead. Apologies if so.

https://github.com/basecamp/kamal/pull/776

If you want to use that functionality in the meantime you can use the fork at https://github.com/swombat/kamal . See the pull request for simple example of how to use this feature.

morgoth:
@swombat please take a look at https://github.com/basecamp/kamal/pull/700 if this fixes your issue

swombat:
@morgoth Yes, actually that would fix my issue! The point is to make it not default to creating remote builds that can't be accessed, and your solution achieves that.

morgoth:
@swombat can you close the issue then?

PR #776, opened 2024-04-18
Add "tag_uncommitted" option to "builder"

swombat:
Discussed in issue https://github.com/basecamp/kamal/issues/775

This adds an option to builder so it doesn't mess up remote building when there are uncommitted local changes. Defaults to the same behaviour as now, so should not disrupt anyone's workflow, assuming there is anyone who actually wants that behaviour of making random build tags that can't be deployed later.

Apologies for the "Bump version" commits... this is my very first pull request..! I guess I shouldn't have run `bin/release`...

Usage is really simple:

```
builder:
  multiarch: false
  tag_uncommitted: false
  remote:
    arch: amd64
    host: ssh://[redacted]@[redacted]
```

With that line, now the random tag won't be added when there are uncommitted changes.

TODO: mention this in the docs?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    kamal (1.4.0)
+    kamal (0)
       activesupport (>= 7.0)
       base64 (~> 0.2)
       bcrypt_pbkdf (~> 1.0)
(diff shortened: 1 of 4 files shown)


morgoth:
This can be closed as #775 is resolved

PR #786, opened 2024-04-27
Add --target option to Builder to support multi-stage Docker builds

hundredwatt:
The builder config now supports specifying a target, e.g.:

config/deploy.yml
```
builder:
  target: prod
```

for a Dockerfile that's structured like:
```
ARG RUBY_VERSION=3.2.2
FROM registry.docker.com/library/ruby:$RUBY_VERSION-slim as base

... etc ...

FROM base as build

... etc ...

FROM base as app

ENTRYPOINT ["/rails/bin/docker-entrypoint"]

# Start the server by default, this can be overwritten at runtime
EXPOSE 3000
CMD ["./bin/rails", "server"]

FROM app as prod

ENV RAILS_ENV="production" 

FROM app as staging

ENV RAILS_ENV="staging" 
```

### DIFF
--- a/lib/kamal/commands/builder/base.rb
+++ b/lib/kamal/commands/builder/base.rb
@@ -3,7 +3,7 @@ class Kamal::Commands::Builder::Base < Kamal::Commands::Base
   class BuilderError < StandardError; end
 
   delegate :argumentize, to: Kamal::Utils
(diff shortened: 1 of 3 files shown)


djmb:
@hundredwatt - thanks for the PR!

If you could fix up the Rubocop error, I'll get this merged.

hundredwatt:
@djmb Fixed 👍

PR #789, opened 2024-05-02
Host specific env with tags

djmb:
Allow hosts to be tagged so we can have host specific env variables.

We might want host specific env variables for things like datacenter
specific tags or testing GC settings on a specific host.

Right now you either need to set up a separate role, or have the app
be host aware.

Now you can define tag env variables and assign those to hosts.

For example:
```
servers:
  - 1.1.1.1
  - 1.1.1.2: tag1
  - 1.1.1.2: tag2
  - 1.1.1.3: [ tag1, tag2 ]
env_tags:
  tag1:
    ENV1: value1
  tag2:
    ENV2: value2
```

The tag env supports the full env format, allowing you to set secret and
clear values.

### DIFF
--- a/lib/kamal/cli/app.rb
+++ b/lib/kamal/cli/app.rb
@@ -37,7 +37,7 @@ class Kamal::Cli::App < Kamal::Cli::Base
 
         roles.each do |role|
           execute *KAMAL.auditor.record("Started app version #{KAMAL.config.version}"), verbosity: :debug
-          execute *KAMAL.app(role: role).start, raise_on_non_zero_exit: false
+          execute *KAMAL.app(role: role, host: host).start, raise_on_non_zero_exit: false
         end
       end
     end
@@ -51,7 +51,7 @@ class Kamal::Cli::App < Kamal::Cli::Base
 
         roles.each do |role|
           execute *KAMAL.auditor.record("Stopped app", role: role), verbosity: :debug
-          execute *KAMAL.app(role: role).stop, raise_on_non_zero_exit: false
+          execute *KAMAL.app(role: role, host: host).stop, raise_on_non_zero_exit: false
         end
       end
     end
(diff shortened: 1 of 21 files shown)


dhh:
This is great, but do we have anything evns we want to apply to these tags? Otherwise it feels like we're nested one level too much. But it's one of those things that if you could imagine anything else, you'd really want the nesting. Maybe just the secret split is enough to warrant?

djmb:
I suppose you might want tag specific volumes? I'm unconvinced, but having the env key does allow us to handle anything that does show up.

Either way I think the secret/clear split does warrant the `env` key anyway for clarity. Though I also think tag specific secrets would be rare.

```
servers:
  - 1.1.1.1
  - 1.1.1.2: tag1
  - 1.1.1.2: tag2
  - 1.1.1.3: [ tag1, tag2 ]
tags:
  tag1:
    clear:
      ENV1: value1
    secret:
      - DB_PASSWORD
  tag2:
    clear:
      ENV1: value1
    secret:
      - REDIS_PASSWORD
 ```

vs

```
servers:
  - 1.1.1.1
  - 1.1.1.2: tag1
  - 1.1.1.2: tag2
  - 1.1.1.3: [ tag1, tag2 ]
tags:
  tag1:
    env: 
      clear:
        ENV1: value1
      secret:
        - DB_PASSWORD
  tag2:
    env:
      clear:
        ENV1: value1
      secret:
        - REDIS_PASSWORD
 ```

dhh:
Let's change this from:

```
env:
  clear:
    TEST: "root"
    EXPERIMENT: "disabled"
env_tags:
  site1:
    SITE: site1
  site2:
    SITE: site2
  experimental:
    env:
      EXPERIMENT: "enabled"
```

to:

```
env:
  clear:
    TEST: "root"
    EXPERIMENT: "disabled"
  tags:
    site1:
      SITE: site1
    site2:
      clear:
        SITE: site2
      secret:
        - SOMETHING_SECRET
```

Issue #503, opened 2023-09-28
"No builder" error message should be just a warning


brunto:
During the first deployment, we have this error message:

```
 DEBUG [bda1b61e]       ERROR: no builder "myapp-multiarch" found
 ERROR Missing compatible builder, so creating a new one first
```

What do you think about putting a warning message instead of an error one in a less anxiety-provoking colour than red?

dorianmariefr:
```
 DEBUG [96824640] Command: docker buildx build --push --platform linux/amd64,linux/arm64 --builder kamal-code-multiarch -t dorianmariefr/code:6ce08eea7eaf33c40f6287427ab0df44d7845d73 -t dorianmariefr/code:latest --label service="code" --file Dockerfile .
 DEBUG [96824640] 	ERROR: no builder "kamal-code-multiarch" found
```

Same here, with kamal

SamuelAierizer:
Just had the same problem. Did someone find a solution?

brunto:
This is just a warning, not an actual error.

darkamenosa:
It would be an error if you didn't have docker running on your computer. I forgot to start docker on my laptop and it took a while to recognize it.

PR #790, opened 2024-05-02
Warn on missing builder

djmb:
We are going to try to create a builder if one is missing, so let's warn rather than report it as an error.


djmb:
Fixes https://github.com/basecamp/kamal/issues/503

PR #802, opened 2024-05-13
Envify during setup

dhh:
Default Rails 8 setup will use an .env.erb file. When you intially do kamal setup, this should be envified.

### DIFF
--- a/lib/kamal/cli/main.rb
+++ b/lib/kamal/cli/main.rb
@@ -186,11 +186,15 @@ class Kamal::Cli::Main < Kamal::Cli::Base
       env_path          = ".env"
     end
 
-    File.write(env_path, ERB.new(File.read(env_template_path), trim_mode: "-").result, perm: 0600)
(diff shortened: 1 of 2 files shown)


lib/kamal/cli/main.rb
```diff
@@ -185,11 +186,15 @@ def envify
...
 
-    unless options[:skip_push]
-      reload_envs
-      invoke "kamal:cli:env:push", options
+      unless options[:skip_push]
+        reload_envs
+        invoke "kamal:cli:env:push", options
+      end
+    else
+      puts "Skipping envify (no #{env_template_path} exist)"
```
Sija, line 197: I'd keep using `say` as in other places.

```suggestion
      say "Skipping envify (no #{env_template_path} exist)", :magenta
```

dhh: We're not in a context where we can use say.

PR #817, opened 2024-05-24
Add grep options to log commands

nickhammond:
This adds in the ability to pass additional options to grep when utilizing the various `logs` commands(app, accessory, traefik).

This allows the user to pass in various options to grep such as `-A 5` to look at the 5 lines after a match,`-B 5` to look at the 5 lines before, or maybe count the number of matching lines with `-c`.

The use case is you're trying to monitor or look at a specific endpoint and you want to see the response codes for a bit. You know the endpoint is "/users/2fa" so you run `kamal app logs -f --grep "users/2fa" --grep-options "-A 5"` to provide 5 more lines of context after your grep. Since the primary use case at the moment for grep is to search for a request ID my thought was that we could utilize context to look at an entire request in the first grep potentially instead of needing to run it twice. 

I also added in some missing tests for grep and broke up some of the tests that had multiple assertions within them to make it a bit easier to follow.

### DIFF
--- a/lib/kamal/cli/accessory.rb
+++ b/lib/kamal/cli/accessory.rb
@@ -150,22 +150,24 @@ class Kamal::Cli::Accessory < Kamal::Cli::Base
   option :lines, type: :numeric, aliases: "-n", desc: "Number of log lines to pull from each server"
   option :grep, aliases: "-g", desc: "Show lines with grep match only (use this to fetch specific requests by id)"
   option :follow, aliases: "-f", desc: "Follow logs on primary server (or specific host set by --hosts)"
+  option :context, aliases: "-C", desc: "Show number of lines leading and trailing a grep match (use with --grep)"
   def logs(name)
     with_accessory(name) do |accessory, hosts|
       grep = options[:grep]
+      context = options[:context]
 
       if options[:follow]
         run_locally do
           info "Following logs on #{hosts}..."
-          info accessory.follow_logs(grep: grep)
-          exec accessory.follow_logs(grep: grep)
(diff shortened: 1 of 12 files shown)


djmb:
`-C` is not a valid flag for POSIX grep (though I'm sure it will be available the vast majority of the time), so I'm not sure we should encode it into Kamal. 

How about adding a `--grep-options` flag and passing those verbatim? 

So fir your example it would be `kamal app logs -f --grep "users/2fa" -grep-options '-C 3'`.

nickhammond:
@djmb Ah, I didn't realize that. I like the `--grep-options` better though since I typically actually use `-B` or `-A` but didn't want to add two new options.

I'll get this PR adjusted!

lib/kamal/cli/accessory.rb
```diff
@@ -149,23 +149,25 @@ def exec(name, cmd)
   option :since, aliases: "-s", desc: "Show logs since timestamp (e.g. 2013-01-02T13:23:37Z) or relative (e.g. 42m for 42 minutes)"
   option :lines, type: :numeric, aliases: "-n", desc: "Number of log lines to pull from each server"
   option :grep, aliases: "-g", desc: "Show lines with grep match only (use this to fetch specific requests by id)"
+  option :grep_options, aliases: "-o", desc: "Additional options supplied to grep"
```
nickhammond, line 152: `o` seemed like a good alias but let me know if there's another preference.

nickhammond (commented):
Updated to utilize `--grep-options` with the same context option.

PR #818, opened 2024-05-24
Fix the header template of the docker-setup hook

fabiosammy:
I know that is a very simple PR, but I lot like 40 minutes before realized that the issue was this header from the template :sweat_smile:


plattenschieber:
Coming from python, I really thought it should be a ruby script and that it uses an internal `ssh` library 😅 Thanks for making this clear @fabiosammy  🙌

PR #819, opened 2024-05-24
Use /bin/sh in docker-setup.sample

mihkelal:
The `docker-setup.sample` hook file contains shell code, however the file's shebang is for ruby code.
Unless I'm missing something, the shebang should mark the file to be executed using shell.
This humble PR corrects this.


mihkelal:
Well, @fabiosammy just beat me to it in #818 😅 .

PR #828, opened 2024-06-04
Configuration validation

djmb:
Validate the Kamal configuration giving useful warning on errors. Each section of the configuration has its own config class and a YAML file containing documented example configuration.

You can run `kamal docs` to see the example configuration, and `kamal docs <section>` to see the example configuration for a specific section.

The validation matches the configuration to the example configuration checking that there are no unknown keys and that the values are of matching types.

Where there is more complex validation - e.g for envs and servers, we have custom validators that implement those rules.

Additonally the configuration examples are used to generate the configuration documentation in the kamal-site repo.

You generate them by running:

```
bundle exec bin/docs <kamal-site-checkout>
```


mhenrixon:
So I am guessing this means there won't be any LSP support @djmb? I appreciate these efforts but would have rather had feedback as I type stuff (LSP-yaml).

tangit86:
@mhenrixon 

https://github.com/tangit86/kamal-json-schema

a quick and dirty solution, might come handy, it did for me.

lenart:
There's also a version for Kamal 2.x - https://github.com/RockieStar-Inc/kamal-json-schema

PR #830, opened 2024-06-05
Fix the multiarch disabled + remote set

fabiosammy:
When we call the kamal with remote only set(no local) + multiarch disabled, it calls the native builder instead of remote builder


djmb:
This is no longer needed with the builder config changes in Kamal 2.0

Issue #588, opened 2023-11-17
ipv6 Only Servers


eric1234:
I'm testing out deploying and running an app on an ipv6 only server with that deployment happening via Kamal. I see on https://github.com/basecamp/kamal/issues/87 that ipv6 should work but I've encountered a few issues.

I have some thoughts about fixes but wasn't sure of them so I figured I would start with opening an issue. If I can get a bit of feedback on it I would be glad to do the legwork to help resolve these issues. So far there are three issues.

Mismatch Between Host and Configuration String
-------------------------------------------------------------------------

This prevents the deploy from working as the env files won't push up (may affect other commands but since it fails so early not sure what else). The issue is when I specify my ipv6 address I must include the brackets around the address (similar to what web browsers require). Example:

```yaml
servers:
  - "[2a01:4f7:c014:160::1]"
```

You can see these brackets being required in SSHKit [here](https://github.com/capistrano/sshkit/blob/3555a98d5c61e26472e988b62601fd323b374176/lib/sshkit/host.rb#L154)

The problem is both this line:

https://github.com/basecamp/kamal/blob/635876bdb96eb00ed85c91fe7231e05a25fd16ad/lib/kamal/commander.rb#L59

and this line:

https://github.com/basecamp/kamal/blob/635876bdb96eb00ed85c91fe7231e05a25fd16ad/lib/kamal/commander.rb#L75

assume the configuration and hostname are the same. But this is really only true for the [SimpleHostParser](https://github.com/capistrano/sshkit/blob/3555a98d5c61e26472e988b62601fd323b374176/lib/sshkit/host.rb#L108-L110)

[All other host parsers](https://github.com/capistrano/sshkit/blob/3555a98d5c61e26472e988b62601fd323b374176/lib/sshkit/host.rb#L188-L191) do something with that initialization string. Therefore when the two problem lines call `to_s` they are only getting the hostname (which does not include the brackets) while the values it's comparing against from the config file do include the brackets.

The additional challenge here is that `host` isn't always a `SSHKit::Host` object. When doing an interactive execution it just becomes a string derived from `Kamal.primary_host` at:

https://github.com/basecamp/kamal/blob/635876bdb96eb00ed85c91fe7231e05a25fd16ad/lib/kamal/cli/app.rb#L126

[...]

PR #862, opened 2024-06-24
Bump sshkit to support unbracketed IPv6 addresses

jeromedalbert:
Set sshkit minimum version to 1.23.0, which includes an enhancement to support unbracketed IPv6 addresses.

See https://github.com/capistrano/sshkit/pull/538
Closes https://github.com/basecamp/kamal/issues/588


djmb:
The tests are failing - I think you need to update `Gemfile.lock`

PR #865, opened 2024-06-25
Ensure envify templates aren't polluted by existing env

jeremy:
Setting `GITHUB_TOKEN` as in the docs results in reusing the existing `GITHUB_TOKEN` since `gh` returns that env var if it's set:
```bash
GITHUB_TOKEN=junk gh config get -h github.com oauth_token
junk
```

Using the original env ensures that the templates will be evaluated the same way regardless of whether envify had been previously invoked.

### DIFF
--- a/test/cli/main_test.rb
+++ b/test/cli/main_test.rb
@@ -1,6 +1,9 @@
 require_relative "cli_test_case"
 
 class CliMainTest < CliTestCase
+  setup { @original_env = ENV.to_h.dup }
+  teardown { ENV.clear; ENV.update @original_env }
+
   test "setup" do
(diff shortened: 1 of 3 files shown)


test/cli/main_test.rb
```diff
@@ -470,6 +473,13 @@ class CliMainTest < CliTestCase
     run_command("envify", "--skip-push")
   end
 
+  test "envify with clean env" do
+    with_test_dotenv(".env": "HELLO=already", ".env.erb": "HELLO=<%= ENV.fetch 'HELLO', 'never' %>") do
+      run_command("envify", "--skip-push")
+      assert_equal "HELLO=never", File.read(".env")
```
jeremy, line 479: Without the patch, `.env` contains `HELLO=already`.

PR #885, opened 2024-07-15
Dont build image if it has already been pushed

npezza93:
This is useful when you push a branch to different environments at separate times from CI. It's also useful if you have CI setup to deploy main but have a concurrency key setup. If you push quickly youll end up deploying the same commit twice and the second time the image doesnt need to be built since it already exists.

### DIFF
--- a/lib/kamal/cli/build.rb
+++ b/lib/kamal/cli/build.rb
@@ -5,7 +5,7 @@ class Kamal::Cli::Build < Kamal::Cli::Base
 
   desc "deliver", "Build app and push app image to registry then pull image on servers"
   def deliver
-    push
+    push unless already_built?
     pull
   end
 
@@ -157,5 +157,11 @@ class Kamal::Cli::Build < Kamal::Cli::Base
         execute *KAMAL.builder.pull
         execute *KAMAL.builder.validate_image
       end
+    end
+
+    def already_built?
+      run_locally { execute *KAMAL.docker.manifest(KAMAL.builder.absolute_image) }
+    rescue SSHKit::Command::Failed
+      false
     end
 end
--- a/lib/kamal/commands/builder.rb
+++ b/lib/kamal/commands/builder.rb
@@ -3,6 +3,7 @@ require "active_support/core_ext/string/filters"
 class Kamal::Commands::Builder < Kamal::Commands::Base
   delegate :create, :remove, :push, :clean, :pull, :info, :context_hosts, :config_context_hosts, :validate_image,
            :first_mirror, to: :target
+  delegate :absolute_image, to: :config
 
   include Clone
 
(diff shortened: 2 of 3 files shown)


npezza93:
@djmb How does the registry work in tests?

djmb:
@npezza93 - the integration tests use docker compose to set up a few containers:

1. A shared container that contains files needed across other containers (ssh keys)
2. A deployment container we will run kamal from
3. Some "host" containers to deploy to
4. A container running an nginx load balancer pointing to the host containers
5. A docker registry

This creates a mini environment we can deploy to and access the hosts from.

The docker registry is running in insecure mode. Any images that we are going to pull from docker hub (e.g. the nginx one) are copied across once and retagged to be pulled from the local registry. This is because otherwise you can hit docker hub rate limits pretty quickly if you are running the tests a lot.

djmb:
The test failure looks like some transient Docker issue.

@npezza93 - this change makes sense, but the one issue with it is that the pre-build hook will be skipped if you don't have to build the image. While that's probably the right thing to do, there may be people who are doing things they want to happen on ever deployment in that hook.

As such I think we'll wait to add this to the 2.0 release (which should be happening fairly soon) and include it in the upgrade notes.

npezza93:
@djmb Sounds good! Thanks for the info. Decided to just stub the call to manifest to force it to always be false which looks to have resolved the test failures.

test/integration/main_test.rb
```diff
@@ -10,7 +10,13 @@ class MainTest < IntegrationTest
 
     assert_app_is_down
 
-    kamal :deploy
+    mock = Minitest::Mock.new
```
djmb, line 13: I don't think we should be mocking things in the integration tests. We should just test the results of the commands.

In any case `kamal :deploy` actually runs `docker compose exec deployer kamal deploy`, so I'm not sure the mocking is doing anything.

We could add a second redeploy of the second version later on and add new `assert_hooks_did_not_run` assertion after it to confirm the pre-build hook was not called, which would be enough to infer that there was no rebuild.

test/integration/main_test.rb
```diff
@@ -19,7 +25,7 @@ class MainTest < IntegrationTest
 
     kamal :redeploy
     assert_app_is_up version: second_version
-    assert_hooks_ran "pre-connect", "pre-build", "pre-deploy", "post-deploy"
+    assert_hooks_ran "pre-connect", "pre-deploy", "post-deploy"
```
djmb, line 28: Does this not call the `pre-build` hook? I would expect it to because we are deploying a new version here.

npezza93:
Good call @djmb. Should be fixed now!

iloveitaly:
An interesting use case here could be building the image outside kamal, but then using kamal to deploy that image.

djmb:
This never made it into Kamal 2. Think it needs to wait for Kamal 3 now unfortunately

PR #900, opened 2024-07-31
Replace .env* with .kamal/env*

djmb:
By default look for the env file in .kamal/env to avoid clashes with other tools using .env.

For now we'll still load .env and issue a deprecation warning, but in future we'll stop reading those.

See https://github.com/rails/rails/issues/51829


djmb:
Superseded by https://github.com/basecamp/kamal/pull/924

Issue #592, opened 2023-11-21
Duplicate buildkit daemon processes with multi-app deployments


jefvlamings:
When building with Docker (via Kamal) a daemon process `buildkitd` is spawned. buildkitd is the daemon that runs in the background to handle the build processes. It is responsible for executing build tasks, managing caching, and optimizing the build process. E.g.:

```
docker buildx build --push --platform linux/arm64 --builder kamal-service-native-remote
```

The builder tag `kamal-service-native-remote` is constructed with the name of the service that you've defined in your deploy.yml.

### Problem
When you're building multiple different apps on the same build server, you will get a buildkitd for each application. The more apps you have running, the more redundant buildkitd process that keep running. Keep in mind that a buildkitd process takes up a 100's of megabytes. 

<img width="539" alt="Screenshot 2023-11-21 at 21 59 33" src="https://github.com/basecamp/kamal/assets/1597286/22acc806-cc38-4898-998d-e744f08aac7c">

As far is I understand we could reuse the same daemon process for different application builds. Remotely or locally, this will save up precious RAM.

### Solution
Do we need the service name to be part of the builder name in `--builder`? If so, would it be an option to override it in deploy.yml like:

```yaml
builder:
  remote:
    arch: arm64
    host: xxx.xxx.xxx.xxx
    name: default
```

Which would result in

```
docker buildx build --push --platform linux/arm64 --builder kamal-default-native-remote
```

PR #905, opened 2024-08-01
Simplify builders config

djmb:
- Just three builders now, local, remote and hybrid
- Default to the docker container driver but allow it to be set. No need to infer when we need a "native" builder anymore
- Specify the arches you want, but not where to build them. If you have a remote builder, Kamal will use it for non-local arches
- Require that an arch is set, and default to `amd64` in the templates
- Remote is just the connection string now
- If you have a remote string, but only build for the local arch we'll ignore it
- Don't include the service in the remote builder name (fixes https://github.com/basecamp/kamal/issues/592)
- Include the remote connection string in the remote builder name, so you don't get clashes when changing remote builders

**Examples:**

On arm64, build for arm64 locally, amd64 remotely and
on amd64, build for amd64 locally, arm64 remotely:

```yaml
builder:
  arch:
    - amd64
    - arm64
  remote: ssh://docker@docker-builder
```

On arm64, build amd64 on remote and on amd64 build locally:

```yaml
builder:
  arch: amd64
  remote: ssh://docker@docker-builder
```

Build amd64 on local:

```yaml
builder:
  arch: amd64
```

Use docker driver, building for local arch:

```yaml
builder:
  driver: docker
  arch: <%= uname_m = `uname -m`; uname_m == "x86_64" ? "amd64" : uname_m %>
```


dhh:
Much nicer, much simpler. Need to update the Rails default config/deploy.yml when this is released.

PR #906, opened 2024-08-07
Allow running detached app commands

aliismayilov:
This is useful for long running rake tasks or [scripts](https://github.com/rails/rails/pull/52335) that can be run without having to keep an open connection to servers.

Example:
```sh
kamal app exec 'bin/rails db:seed' --detach
```

### DIFF
--- a/lib/kamal/cli/app.rb
+++ b/lib/kamal/cli/app.rb
@@ -75,8 +75,10 @@ class Kamal::Cli::App < Kamal::Cli::Base
   option :interactive, aliases: "-i", type: :boolean, default: false, desc: "Execute command over ssh for an interactive shell (use for console/bash)"
   option :reuse, type: :boolean, default: false, desc: "Reuse currently running container instead of starting a new one"
   option :env, aliases: "-e", type: :hash, desc: "Set environment variables for the command"
+  option :detach, type: :boolean, default: false, desc: "Execute command in a detached container"
   def exec(cmd)
     env = options[:env]
+    detach = options[:detach]
     case
     when options[:interactive] && options[:reuse]
       say "Get current version of running container...", :magenta unless options[:version]
@@ -118,7 +120,7 @@ class Kamal::Cli::App < Kamal::Cli::Base
 
           roles.each do |role|
             execute *KAMAL.auditor.record("Executed cmd '#{cmd}' on app version #{version}"), verbosity: :debug
-            puts_by_host host, capture_with_info(*KAMAL.app(role: role, host: host).execute_in_new_container(cmd, env: env))
+            puts_by_host host, capture_with_info(*KAMAL.app(role: role, host: host).execute_in_new_container(cmd, env: env, detach: detach))
           end
         end
       end
--- a/lib/kamal/commands/app/execution.rb
+++ b/lib/kamal/commands/app/execution.rb
@@ -7,9 +7,10 @@ module Kamal::Commands::App::Execution
       *command
   end
 
-  def execute_in_new_container(*command, interactive: false, env:)
+  def execute_in_new_container(*command, interactive: false, detach: false, env:)
     docker :run,
       ("-it" if interactive),
+      ("--detach" if detach),
       "--rm",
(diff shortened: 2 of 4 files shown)


lib/kamal/cli/app.rb
```diff
@@ -94,9 +94,11 @@ def details
   option :interactive, aliases: "-i", type: :boolean, default: false, desc: "Execute command over ssh for an interactive shell (use for console/bash)"
   option :reuse, type: :boolean, default: false, desc: "Reuse currently running container instead of starting a new one"
   option :env, aliases: "-e", type: :hash, desc: "Set environment variables for the command"
+  option :detach, type: :boolean, default: false, desc: "Execute command in a detached container"
   def exec(*cmd)
     cmd = Kamal::Utils.join_commands(cmd)
     env = options[:env]
+    detach = options[:detach]
```
djmb, line 101: `detach` is not compatible with `reuse` or `interactive` - could you raise an error if we try to combine them?

djmb (changes requested):
Hi @aliismayilov,

Sorry for the delay in getting to this! I think we just need some checks for invalid option combinations and we should be good to go!

djmb:
Also it looks like there are some test failures - `CliAppTest#test_exec_detach` and `CommandsAppTest#test_execute_in_new_detached_container`

aliismayilov:
> Hi @aliismayilov,
> Sorry for the delay in getting to this! I think we just need some checks for invalid option combinations and we should be good to go!

Thanks for looking into this! I've added an incompatibility check. See if it makes sense.


> Also it looks like there are some test failures - `CliAppTest#test_exec_detach` and `CommandsAppTest#test_execute_in_new_detached_container`

Not anymore! I must've missed them in the avalanche of log lines 😄

aliismayilov:
I just realized that there will be no output from a detached container in the logs. I created a separate issue: https://github.com/basecamp/kamal/issues/1154

aliismayilov:
@djmb I've updated the pull request with more changes to support logs inspection for detached containers. In short:
- all newly started app containers get kamal logging configuration applied
- detached containers do not have `--rm` option. This allows inspection of detached containers even after they finished running
- `kamal logs` gets a new option `container_id` to inspect logs of a specific container

I've also updated the pull request description.

lib/kamal/commands/app/logging.rb
```diff
@@ -1,15 +1,15 @@
 module Kamal::Commands::App::Logging
-  def logs(version: nil, timestamps: true, since: nil, lines: nil, grep: nil, grep_options: nil)
+  def logs(container_id: nil, timestamps: true, since: nil, lines: nil, grep: nil, grep_options: nil)
```
djmb, line 2: We need the `version` option here, it's used in `Kamal::Cli::App::Boot`

aliismayilov:
@djmb I've rebased the PR. There shouldn't be any conflicts or failures 😬

lib/kamal/commands/app/logging.rb
```diff
@@ -1,18 +1,28 @@
...
+        container_id_command(container_id),
         "xargs docker logs#{" --timestamps" if timestamps}#{" --tail #{lines}" if lines} --follow 2>&1",
         (%(grep "#{grep}"#{" #{grep_options}" if grep_options}) if grep)
       ),
       host: host
   end
+
+  private
+
+  def container_id_command(container_id)
```
aliismayilov, line 21: TBH, I'm not happy about this approach, but I couldn't come up with a better approach that would allow passing literal container ID (via kamal command) and a command that outputs container ID (see: [lib/kamal/cli/app/boot.rb](https://github.com/basecamp/kamal/pull/906/files#diff-16c00ee2b4b547d9ca9cd78761f722f4f4cc56b0f0c709c61c9beab9960f08e1R94))

aliismayilov:
OK, I fixed the failing spec and to be sure, the CI passes on my ~~machine~~ account: https://github.com/aliismayilov/kamal/actions/runs/11990309351

PR #908, opened 2024-08-13
SSH: allow setting `config: false` to ignore local user `~/.ssh/config`

jeremy:
Documents the existing `config` option so it can be set via Kamal config. Allows setting `config: false` to ignore users' `~/.ssh/config`.

### DIFF
--- a/lib/kamal/configuration/docs/ssh.yml
+++ b/lib/kamal/configuration/docs/ssh.yml
@@ -64,3 +64,10 @@ ssh:
   # An array of strings, with each element of the array being
   # a raw private key in PEM format.
   key_data: [ "-----BEGIN OPENSSH PRIVATE KEY-----" ]
+
+  # Config
+  #
+  # Set to true to load the default OpenSSH config files (~/.ssh/config,
+  # /etc/ssh_config), to false ignore config files, or to a file path
+  # (or array of paths) to load specific configuration. Defaults to true.
+  config: true


igor-alexandrov:
@jeremy are you sure that it exists? I cannot find it, to be honest.

jeremy:
You mean the sshkit config option, or ?

jeremy:
It eventually [lands in](https://github.com/net-ssh/net-ssh/blob/6fe0040eb6b4b45ff1eeca58567f7077e02d3c05/lib/net/ssh.rb#L235) `Net::SSH.start` which passes it to `Net::SSH.configuration_for` [to determine](https://github.com/net-ssh/net-ssh/blob/6fe0040eb6b4b45ff1eeca58567f7077e02d3c05/lib/net/ssh.rb#L288) which config files to parse.

Can verify by passing `config: /some/other/path.conf` and verifying the config is parsed or by enabling/disabling and verifying that default config is/isn't parsed (net-ssh [defaults to parsing](https://github.com/net-ssh/net-ssh/blob/6fe0040eb6b4b45ff1eeca58567f7077e02d3c05/lib/net/ssh/config.rb#L47) all of `%w[ ~/.ssh/config /etc/ssh_config /etc/ssh/ssh_config ]`).

igor-alexandrov:
@jeremy thanks for the explanation.

igor-alexandrov (approved):

Burgestrand:
I saw this change on the documentation site, but it appears to not work due to configuration validation.

```yaml
ssh:
  config: "config/ssh_config"
```

```bash
❯ kamal config
  ERROR (Kamal::ConfigurationError): ssh/config: should be a boolean
```

I tested this on both 2.2.2 and 74a06b0ccda616c86ebe1729d0795f39bcac9f00.

Burgestrand:
I made a PR #1129 that seems to make this config option do something.

jeremy:
@Burgestrand note this applies to `sshkit` config, not `ssh`

Burgestrand:
> @Burgestrand note this applies to `sshkit` config, not `ssh`. Definitely confusing that it doesn't apply elsewhere - probably should have made this an sshkit-specific config and plumbed it through.

I don't understand, but I'd like to.

This change was made to [lib/kamal/configuration/docs/ssh.yml](https://github.com/basecamp/kamal/blob/74a06b0ccda616c86ebe1729d0795f39bcac9f00/lib/kamal/configuration/docs/ssh.yml#L65-L70), which is _specifically_ `ssh`.

`sshkit` has its own configuration file in [lib/kamal/configuration/docs/sshkit.yml](https://github.com/basecamp/kamal/blob/74a06b0ccda616c86ebe1729d0795f39bcac9f00/lib/kamal/configuration/docs/sshkit.yml), which doesn't mention anything about this option.

jeremy:
Yep, it's confusing for sure.

SSHKit is configured with config.ssh.options here: https://github.com/basecamp/kamal/blob/74a06b0ccda616c86ebe1729d0795f39bcac9f00/lib/kamal/commander.rb#L167

From there can trace along with https://github.com/basecamp/kamal/pull/908#issuecomment-2387652198

jeremy:
Hmm, indeed! That is the limited/curated set of options. I'll take a closer look at how I was getting the raw config through to SSHKit in my testing.

PR #912, opened 2024-08-26
Add aliases to Kamal

djmb:
Aliases are defined in the configuration file under the `aliases` key.

The configuration is a map of alias name to command. When we run the command the we just do a literal replacement of the alias with the string.

So if we have:

```yaml
aliases:
  console: app exec -r console -i --reuse "rails console"
```

Then running `kamal console -r workers` will run the command

```sh
$ kamal app exec -r console -i --reuse "rails console" -r workers
```

Because of the order Thor parses the arguments, this allows us to override the role from the alias command.

There might be cases where we need to munge the command a bit more but that would involve getting into Thor command parsing internals, which are complicated and possibly subject to change.

There's a chance that your aliases could conflict with future built-in commands, but there's not likely to be many of those and if it happens you'll get a validation error when you upgrade.

Thanks to @dhnaranjo for the idea!

### DIFF
--- a/lib/kamal/configuration/validator.rb
+++ b/lib/kamal/configuration/validator.rb
@@ -13,32 +13,34 @@ class Kamal::Configuration::Validator
 
   private
     def validate_against_example!(validation_config, example)
-      validate_type! validation_config, Hash
+      validate_type! validation_config, example.class
 
-      check_unknown_keys! validation_config, example
+      if example.class == Hash
+        check_unknown_keys! validation_config, example
 
-      validation_config.each do |key, value|
-        next if extension?(key)
-        with_context(key) do
-          example_value = example[key]
+        validation_config.each do |key, value|
+          next if extension?(key)
+          with_context(key) do
+            example_value = example[key]
 
-          if example_value == "..."
-            validate_type! value, *(Array if key == :servers), Hash
-          elsif key == "hosts"
-            validate_servers! value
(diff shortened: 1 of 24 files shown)


lib/kamal/configuration/validator.rb
```diff
@@ -13,32 +13,34 @@ def validate!
...
-            validate_servers! value
-          elsif example_value.is_a?(Array)
-            validate_array_of! value, example_value.first.class
-          elsif example_value.is_a?(Hash)
-            case key.to_s
-            when "options", "args"
-              validate_type! value, Hash
-            when "labels"
-              validate_hash_of! value, example_value.first[1].class
+      validate_type! validation_config, example.class
```
djmb, line 16: Previously all second level config items were expected to be hashes, but for aliases then are strings, so we need to expand the config validation to handle that.

djmb:
I've added https://github.com/basecamp/kamal/pull/912/commits/579e169be21ca27b8d1b50d1a75574316a17edb2, so that commands can be combined for `app exec` and `server exec`.

This means we can have an alias like `rails: app exec -p rails` and then run `kamal rails db:migrate`.

dhh:
This is going to be so nice. We should define a few of the stock ones we use for Rails, like console, in the Rails' default config/deploy.yml file when this is released 👌

PR #946, opened 2024-09-18
Downgrade from Kamal 2 to 1.9

djmb:
Add a downgrade command, so you can reverse the upgrade process and go back to Kamal 1.9. This replaces kamal-proxy and reboots all the accessories.

This gives an upgrade and downgrade path:

Upgrade:
1. Upgrade config to be Kamal 2 compatible + use kamal 2.0
2. Run `kamal upgrade`

Downgrade:
1. Switch back to previous config + use kamal 1.9
2. Run `kamal downgrade`

You can set `--rolling` to downgrade one host at a time.

### DIFF
--- a/lib/kamal/cli/accessory.rb
+++ b/lib/kamal/cli/accessory.rb
@@ -222,6 +222,25 @@ class Kamal::Cli::Accessory < Kamal::Cli::Base
     end
   end
 
+  desc "downgrade", "Downgrade accessories from Kamal 2 to 1.9"
+  option :rolling, type: :boolean, default: false, desc: "Upgrade one host at a time"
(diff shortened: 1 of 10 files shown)


ghillahill:
In my project I'm using Kamal 1.9.1

but if i run `kamal version // return 2.3.0`

if i run `kamal downgrade // return ERROR (Kamal::ConfigurationError): unknown key: traefik`

can you help me understand what I am doing wrong?

djmb:
Hey @ghillahill - are you running Kamal via the gem?

If so running `gem list kamal` will show you which versions you have installed.

You could try uninstalling version 2.3.0 and any other versions that are newer than 1.9, or run the version you want with `kamal _1.9.1_`?

PR #969, opened 2024-09-24
Added support for custom SSL certificate

kpumuk:
Following the [change in kamal-proxy](https://github.com/basecamp/kamal-proxy/pull/17), this MR introduces a configuration option to load custom SSL certificate and the corresponding private key from disk:

```yaml
proxy:
  ssl: true
  ssl_certificate_path: /data/cert/foo.example.com/fullchain.pem
  ssl_private_key_path: /data/cert/foo.example.com/privkey.pem
```

Documentation preview:

image

### DIFF
--- a/lib/kamal/configuration/proxy.rb
+++ b/lib/kamal/configuration/proxy.rb
@@ -22,6 +22,10 @@ class Kamal::Configuration::Proxy
     proxy_config.fetch("ssl", false)
   end
 
+  def custom_ssl_certificate?
+    proxy_config["ssl_certificate_path"].present?
+  end
+
   def hosts
     proxy_config["hosts"] || proxy_config["host"]&.split(",") || []
   end
--- a/lib/kamal/configuration/proxy.rb
+++ b/lib/kamal/configuration/proxy.rb
@@ -30,6 +30,8 @@ class Kamal::Configuration::Proxy
     {
       host: hosts,
       tls: proxy_config["ssl"].presence,
+      "tls-certificate-path": proxy_config["ssl_certificate_path"],
+      "tls-private-key-path": proxy_config["ssl_private_key_path"],
       "deploy-timeout": seconds_duration(config.deploy_timeout),
       "drain-timeout": seconds_duration(config.drain_timeout),
       "health-check-interval": seconds_duration(proxy_config.dig("healthcheck", "interval")),
--- a/lib/kamal/configuration/docs/proxy.yml
+++ b/lib/kamal/configuration/docs/proxy.yml
@@ -49,6 +49,17 @@ proxy:
   # Defaults to `false`:
   ssl: true
 
+  # Custom SSL certificate
+  #
+  # In scenarios where Let's Encrypt is not an option, or you already have your own
+  # certificates from a different Certificate Authority, you can configure kamal-proxy
+  # to load the certificate and the corresponding private key from disk.
+  #
+  # The certificate must be in PEM format and contain the full chain. The private key
+  # must also be in PEM format.
+  ssl_certificate_path: /data/cert/foo.example.com/fullchain.pem
+  ssl_private_key_path: /data/cert/foo.example.com/privkey.pem
+
   # Response timeout
   #
   # How long to wait for requests to complete before timing out, defaults to 30 seconds:
--- a/lib/kamal/configuration/role.rb
+++ b/lib/kamal/configuration/role.rb
@@ -150,8 +150,8 @@ class Kamal::Configuration::Role
   end
 
   def ensure_one_host_for_ssl
-    if running_proxy? && proxy.ssl? && hosts.size > 1
-      raise Kamal::ConfigurationError, "SSL is only supported on a single server, found #{hosts.size} servers for role #{name}"
+    if running_proxy? && proxy.ssl? && hosts.size > 1 && !proxy.custom_ssl_certificate?
+      raise Kamal::ConfigurationError, "SSL is only supported on a single server or with custom SSL certificates, found #{hosts.size} servers for role #{name}"
     end
   end
 
--- a/lib/kamal/configuration/validator/proxy.rb
+++ b/lib/kamal/configuration/validator/proxy.rb
@@ -10,6 +10,14 @@ class Kamal::Configuration::Validator::Proxy < Kamal::Configuration::Validator
       if (config.keys & [ "host", "hosts" ]).size > 1
         error "Specify one of 'host' or 'hosts', not both"
       end
+
+      if config["ssl_certificate_path"].present? && config["ssl_private_key_path"].blank?
+        error "Must set a private key path to use a custom SSL certificate"
+      end
+
+      if config["ssl_private_key_path"].present? && config["ssl_certificate_path"].blank?
+        error "Must set a certificate path to use a custom SSL private key"
+      end
     end
   end
 end
--- a/test/commands/app_test.rb
+++ b/test/commands/app_test.rb
@@ -143,6 +143,14 @@ class CommandsAppTest < ActiveSupport::TestCase
       new_command.deploy(target: "172.1.0.2").join(" ")
   end
 
+  test "deploy with custom SSL certificate" do
+    @config[:proxy] = { "ssl" => true, "host" => "example.com", "ssl_certificate_path" => "/path/to/cert.pem", "ssl_private_key_path" => "/path/to/key.pem" }
+
+    assert_equal \
+      "docker exec kamal-proxy kamal-proxy deploy app-web --target=\"172.1.0.2:80\" --host=\"example.com\" --tls --tls-certificate-path=\"/path/to/cert.pem\" --tls-private-key-path=\"/path/to/key.pem\" --deploy-timeout=\"30s\" --drain-timeout=\"30s\" --buffer-requests --buffer-responses --log-request-header=\"Cache-Control\" --log-request-header=\"Last-Modified\" --log-request-header=\"User-Agent\"",
+      new_command.deploy(target: "172.1.0.2").join(" ")
+  end
+
   test "remove" do
     assert_equal \
       "docker exec kamal-proxy kamal-proxy remove app-web",
--- a/test/configuration/proxy_test.rb
+++ b/test/configuration/proxy_test.rb
@@ -38,6 +38,16 @@ class ConfigurationProxyTest < ActiveSupport::TestCase
     assert_not config.proxy.ssl?
   end
 
+  test "ssl with certificate path and no private key path" do
+    @deploy[:proxy] = { "ssl" => true, "ssl_certificate_path" => "/path/to/cert.pem" }
+    assert_raises(Kamal::ConfigurationError) { config.proxy.ssl? }
+  end
+
+  test "ssl with private key path and no certificate path" do
+    @deploy[:proxy] = { "ssl" => true, "ssl_private_key_path" => "/path/to/key.pem" }
+    assert_raises(Kamal::ConfigurationError) { config.proxy.ssl? }
+  end
+
   private
     def config
       Kamal::Configuration.new(@deploy)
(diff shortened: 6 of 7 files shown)


kpumuk:
> I'm a newbie to kamal (and docker) and was wondering what's the recommended way to get those custom PEM files on server.

The easiest way would be to use `pre-proxy-reboot`. For example, if you use 1Password to manage secrets:

1. Put both `cert.pem` and `key.pem` under the item you use for secrets
2. Create `.kamal/hooks/pre-proxy-reboot`:
   ```bash
   #!/bin/sh
   
   set -euo pipefail
   
   KAMAL_PROXY_TLS_CERT=$(op read "op://Private/Kamal Demo/cert.pem")
   KAMAL_PROXY_TLS_PRIVATE_KEY=$(op read "op://Private/Kamal Demo/key.pem")
   
   for ip in ${KAMAL_HOSTS//,/ }; do
     ssh -q -T -o BatchMode=yes ubuntu@"${ip}" bash --noprofile <<-EOF
       mkdir -p .kamal/apps/${KAMAL_SERVICE}/tls
       echo '${KAMAL_PROXY_TLS_CERT}' > .kamal/apps/${KAMAL_SERVICE}/tls/cert.pem
       echo "${KAMAL_PROXY_TLS_PRIVATE_KEY}" > .kamal/apps/${KAMAL_SERVICE}/tls/key.pem
   EOF
   done
   ```
3. Run `kamal proxy reboot` to deploy
4. Use `~/.kamal/apps/${KAMAL_SERVICE}/tls/*.pem` to configure Kamal

> Copy using kamal+Dockerfile or do it manually ? Maybe better alternative would be to load the values via ENV variables (given kamal-proxy could support it) ?

`kamal-proxy` does not support environment variables.

yogeshjain999:
Cool, setting it up via `pre-proxy-reboot` sounds good approach. Thanks!!

kpumuk:
@djmb to follow up on the [thread in kamal-proxy](https://github.com/basecamp/kamal-proxy/pull/17#issuecomment-2387898913), here is the documentation update + support for the recent custom TLS cert changes.

agu-z:
Is it possible to specify a client certificate too? I need this in order to enable CloudFlare's [Authenticated Origin Pulls](https://developers.cloudflare.com/ssl/origin-configuration/authenticated-origin-pull/)

mtmckenna:
With this change, would it be possible to remove the `ensure_one_host_for_ssl` requirement when providing a cert and key? 

Context: I'm looking for a way to have end-to-end in-transit encryption in a regulated environment that requires TLS between the load balancer and server node. I'd also like to be able to use multiple app servers.

By providing my own cert to kamal-proxy, I was thinking I should be able to have the load balancer terminate SSL and then re-encrypt the traffic to kamal-proxy, which would use my supplied cert. wdyt?

Thank you!

kpumuk:
> By providing my own cert to kamal-proxy, I was thinking I should be able to have the load balancer terminate SSL and then re-encrypt the traffic to kamal-proxy, which would use my supplied cert. wdyt?

Yep, that's exactly how it would work with custom certificates, and it removes the limitation of one host behind the load balancer.

mtmckenna:
> Yep, that's exactly how it would work with custom certificates, and it removes the limitation of one host behind the load balancer.

That's great! Would `ensure_one_host_for_ssl` be updated in a separate PR?

https://github.com/basecamp/kamal/blob/607368121e5bf9fbd8bacd9ae399ada650f3c40c/lib/kamal/configuration/role.rb#L152


Thank you!

kpumuk:
> That's great! Would `ensure_one_host_for_ssl` be updated in a separate PR?

Great catch, I missed this change in the upstream :-) Added a commit to this MR

qinmingyuan:
Works fine. Please merge this PR.

christo-ph:
I am looking forward to this feature. Could you kindly explain how I can test this without being merged and released?

lib/kamal/configuration/proxy.rb
```diff
@@ -30,6 +34,8 @@ def deploy_options
     {
       host: hosts,
       tls: proxy_config["ssl"].presence,
+      "tls-certificate-path": proxy_config["ssl_certificate_path"],
```
djmb, line 37: We should mount a single volume into the proxy when we boot it.

This avoids having to set application level configuration for proxy volumes, which we want to avoid because multiple applications can share the same proxy.

We can then re-use that volume for any other features that need to provide files to the proxy (such as custom error pages for maintenance mode).

Also ideally we wouldn't need to pass paths around and kamal-proxy should be able to infer the location of the certificate.

I think it would mean that kamal-proxy would need to assume that app supplied files live in a different subdirectory to `/home/kamal-proxy/.config/kamal-proxy` so we can map the volume in separately - e.g. `$(PWD)/.kamal/proxy/apps:/home/kamal-proxy/.config/kamal-proxy-apps`.

`/home/kamal-proxy/.config/kamal-proxy` is then writable for the proxy, while `/home/kamal-proxy/.config/kamal-proxy-apps` is read only.

Then we can have `/home/kamal-proxy/.config/kamal-proxy-apps/#{service}-#{role}-#{destination}/tls/` and store the certs in there. `#{service}-#{role}-#{destination}` will match the service name passed to the proxy, so kamal-proxy can infer where the certs are stored. 

What to do think @kevinmcconnell?

djmb: We'll also need to update `kamal app remove` to scrub the `/home/kamal-proxy/.config/kamal-proxy-apps/#{service}-#{role}-#{destination}` directory.

djmb (changes requested):
Hi @kpumuk!

Thanks for putting this together and sorry for the delay in responding - I've been away.

I've added some comments on this - the main things are:
- providing a single way to map files into the proxy container
- read the certs from the Kamal secrets
- get kamal-proxy to infer the cert locations 

I'll need to tackle the volume/file mounting work as part of the maintenance mode work, so it might be easier to wait until that's out and then this can piggy back on that.

Rukamakama:
Looking forward to having this functionality available, we are deploying multiple nodes behind a load balancer and would like to secure the transactions between the load balancer and application servers.
Thanks !

lib/kamal/configuration/proxy.rb
kevinmcconnell: The `kamal-proxy-apps` plan sounds good to me. The proxy won't really care how it's organized, as we can give it absolute paths for the files that it needs (certs, error pages, etc). But as a scheme for organizing the per-app resources I think it makes sense 👍

frenkel:
I'm trying to move this forward. If I understand the feedback correctly the following changes are needed:

- Modify `Kamal::Commands::Proxy#run` to:
  - mount an extra volume read-only under `/home/kamal-proxy/.config/kamal-proxy-apps`
- Modify `Kamal::Cli::App#boot` to:
  - write certificate and key from secrets into files under `/home/kamal-proxy/.config/kamal-proxy-apps/#{service}-#{role}-#{destination}/tls/` using something like `docker exec kamal-proxy -ti echo "contents" > /home/kamal-proxy/.config/kamal-proxy-apps/#{service}-#{role}-#{destination}/tls/cert.pem`
  - make sure permissions are safe for the keys
- Modify `Kamal::Cli::App#remove` to:
  - remove the certificate and key from `/home/kamal-proxy/.config/kamal-proxy-apps/#{service}-#{role}-#{destination}/tls/`

I think writing the secrets into files needs to be done from within the container, so that volume cannot be read-only, right?

Let me know if this is the way you would like to proceed and I'll give it a try.

lib/kamal/configuration/proxy.rb
kevinmcconnell: Thinking about this part a bit more:

> ideally we wouldn't need to pass paths around and kamal-proxy should be able to infer the location of the certificate ...
> Then we can have `/home/kamal-proxy/.config/kamal-proxy-apps/#{service}-#{role}-#{destination}/tls/` and store the certs in there. `#{service}-#{role}-#{destination}` will match the service name passed to the proxy, so kamal-proxy can infer where the certs are stored.

I think it may be better to _not have default paths_ for this on the proxy side, and instead always make it explicit by passing the paths as part of the deploy command. Although that makes the deploy commands longer, it has a few advantages:

- Only Kamal needs to be aware of the file structure it's using, and we're free to evolve that later on the Kamal side if we need to. Whereas if we lean on defaults, both Kamal & Kamal Proxy need to agree on what those are, so that Kamal can put the files in the places where Kamal Proxy expects to find them. I think it'll simplify things if we make it clear who owns the decision of where the files go, and who owns the files themselves.
- This is perhaps a special case of the same point, but, I'm currently looking at ways to add path-based routing to the proxy (https://github.com/basecamp/kamal-proxy/issues/48), and one of the ramifications of that is TLS settings could straddle multiple services. The TLS settings are really domain-specific, but with path-based routing the same domain can cover multiple services. I'm not sure yet exactly how that will end up looking, but it does mean that mapping cert files to the service name may no longer be reliable (especially as services are added/updated/removed over time).
- Similarly, if we have default paths for the certs, how do we specify when they should be used? If there was an old cert in that location but you wanted to switch to using automatic TLS instead, how does the proxy know what to do? I suppose we could add an additional flag like `--tls-use-custom` but at that point maybe passing the file paths is just as clear.

So my preference would be that Kamal implements a scheme like you described, but that we're explicit about paths when interacting with the proxy so it can remain agnostic about where we put things. What do you think?

jcmaciel:
Hey guys, any updates about it?

treant-prime:
@kpumuk Hi, I'm having problem, I'm installing this PR with `gem 'kamal', github: 'basecamp/kamal', ref: "refs/pull/969/head"` and directly form your repo `gem 'kamal', git: 'https://github.com/kpumuk/kamal.git', branch: 'custom-ssl'` yet I am getting `ERROR (Kamal::ConfigurationError): proxy: unknown keys: ssl_certificate_path, ssl_private_key_path` error during `kamal deploy` both times,

gemfile.lock:
```
GIT
  remote: https://github.com/basecamp/kamal.git
  revision: ff32dcb0b9e6741a9d5dfbfbff44fd833f908108
  ref: refs/pull/969/head
  specs:
    kamal (2.2.2)
      activesupport (>= 7.0)
      base64 (~> 0.2)
      bcrypt_pbkdf (~> 1.0)
      concurrent-ruby (~> 1.2)
      dotenv (~> 3.1)
      ed25519 (~> 1.2)
      net-ssh (~> 7.0)
      sshkit (>= 1.23.0, < 2.0)
      thor (~> 1.3)
      zeitwerk (~> 2.5)
 ```

**Comment by viktorianer** [comment]
Did you add those keys to the config? I think you are missing those new keys.

**Comment by KhanhVanCong** [comment]
> Why recommending no SSL when behind Cloudflare? #1039

I have the same issue with you. Do you find any ways to fix it?

**Comment by kpumuk** [comment]
I am going to work on updating this MR with the recent code and suggestions from @djmb and @kevinmcconnell this week.

**Comment by khaled-badenjki** [comment]
I need this for a project i'm working on. would appreciate any progress made to it

**Comment by letItCurl** [comment]
I give you conference ticket if it can motivate you :P

I cant really understand how this is done, but, my usecase is that my VM hostname is not available from the web.

So lets encrypt cant sign this.

**Comment by letItCurl** [comment]
From my understanding, its merged:
https://github.com/basecamp/kamal-proxy/pull/17/files

But we cant toggle the feature from kamal, right?

**Comment by mhenrixon** [comment]
Do you need any help finishing this up @kpumuk? Would love to see this merged.

**Comment by acidtib** [comment]
i took a stab at it, loading the custom certificate from secrets #1531 

```yaml
proxy:
  ssl:
    certificate_pem: CERTIFICATE_PEM
    private_key_pem: PRIVATE_KEY_PEM
```

frenkel:
As #1531 was merged, I think this PR can be closed @djmb.

PR #972, opened 2024-09-25
Add provenance option

kohkimakimoto:
## Summary

This PR adds a `provenance` option in the `builder` config. You can use `provenance: false` to prevent the showing of `unknown/unknown` architecture on the container registry.

## Description

I am using Kamal with GitHub Container Registry, and it works very well. However, when I push images to the container registry, I noticed an `unknown/unknown` architecture listed on the registry. See the screenshot below.

CleanShot 2024-09-25 at 19 04 38@2x

I believe this is an issue with the UI of the registry service, such as GitHub Container Registry. You can use `provenance: false` as a workaround to avoid the issue.

Once the PR is merged, you can use the following config to prevent pushing the `unknown/unknown` architecture.

```yaml
builder:
  arch: amd64
  provenance: false
```

## Related information

- https://github.com/orgs/community/discussions/45969
- https://github.com/docker/buildx/issues/1507

### DIFF
--- a/lib/kamal/commands/builder/base.rb
+++ b/lib/kamal/commands/builder/base.rb
@@ -6,7 +6,7 @@ class Kamal::Commands::Builder::Base < Kamal::Commands::Base
   delegate :argumentize, to: Kamal::Utils
   delegate \
     :args, :secrets, :dockerfile, :target, :arches, :local_arches, :remote_arches, :remote,
-    :cache_from, :cache_to, :ssh, :driver, :docker_driver?,
+    :cache_from, :cache_to, :ssh, :provenance, :driver, :docker_driver?,
     to: :builder_config
 
   def clean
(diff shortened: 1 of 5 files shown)


lib/kamal/commands/builder/base.rb
```diff
@@ -97,6 +97,10 @@ def build_ssh
       argumentize "--ssh", ssh if ssh.present?
     end
 
+    def builder_provenance
+      argumentize "--provenance", provenance unless provenance.nil?
```
Sija, line 101:
```suggestion
      argumentize "--provenance", provenance if provenance.present?
```

kohkimakimoto: This is intentional design, because if `provenance` is `false`, it should output the `--provenance false` option. It should only skip outputting `--provenance …` if `provenance` is nil.

kohkimakimoto:
See also https://github.com/basecamp/kamal/discussions/838
It is a related discussion.

lado936:
Any new on this, i am testing kamal for side project and this feature would be really helpful as right now in Gitlab CI, you cant check any info about image

PR #994, opened 2024-09-28
[bitwarden] ability to fetch all fields from an item

honzasterba:
Sometimes a projects has a lot of secrets (more than 10). And its cumbersome to write $(kama secrets fetch ...) with a lot of field names.

I want to be able to just fetch all the fields from a given item and then just use these with $(kamal extract NAME)

The result is that you can:
- create on item in bitwaren of type Secre note
- add custom fields to that item
- fetch all the custom fields into your .kamal/secrets like so:
```
SECRETS=$(kamal secrets fetch --adapter bitwarden SECRETS_ITEM)
SECRET_VAR=$(kamal secrets extract FIELD_NAME $SECRETS)
```

### DIFF
--- a/lib/kamal/secrets/adapters/bitwarden.rb
+++ b/lib/kamal/secrets/adapters/bitwarden.rb
@@ -25,22 +25,28 @@ class Kamal::Secrets::Adapters::Bitwarden < Kamal::Secrets::Adapters::Base
       {}.tap do |results|
         items_fields(secrets).each do |item, fields|
           item_json = run_command("get item #{item.shellescape}", session: session, raw: true)
-          raise RuntimeError, "Could not read #{secret} from Bitwarden" unless $?.success?
+          raise RuntimeError, "Could not read #{item} from Bitwarden" unless $?.success?
           item_json = JSON.parse(item_json)
-
           if fields.any?
-            fields.each do |field|
-              item_field = item_json["fields"].find { |f| f["name"] == field }
-              raise RuntimeError, "Could not find field #{field} in item #{item} in Bitwarden" unless item_field
-              value = item_field["value"]
-              results["#{item}/#{field}"] = value
-            end
+            results.merge! fetch_secrets_from_fields(fields, item, item_json)
           elsif item_json.dig("login", "password")
             results[item] = item_json.dig("login", "password")
+          elsif item_json["fields"]&.any?
+            fields = item_json["fields"].pluck("name")
+            results.merge! fetch_secrets_from_fields(fields, item, item_json)
           else
             raise RuntimeError, "Item #{item} is not a login type item and no fields were specified"
           end
         end
+      end
+    end
+
+    def fetch_secrets_from_fields(fields, item, item_json)
+      fields.to_h do |field|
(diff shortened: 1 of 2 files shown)


lib/kamal/secrets/adapters/bitwarden.rb
```diff
@@ -29,6 +29,7 @@ def fetch_secrets(secrets, account:, session:)
           item_json = JSON.parse(item_json)
 
           if fields.any?
+            fields = item_json["fields"].pluck("name") if fields == [ "__ALL__" ]
```
djmb, line 32: I think we can just allow the name to be passed without any fields and then return all the values - so something like this:

```ruby
          elsif (item_fields = item_json.dig("fields"))
            item_fields.each do |item_field|
              results["#{item}/#{item_field["name"]}"] = item_field["value"]
            end
          else
```

Then we can avoid the special `__ALL__` value and you can say:

```
SECRETS=$(kamal secrets fetch --adapter bitwarden --account email@example.com ITEM)
FIELD1=$(kamal secrets extract FIELD1)
```

honzasterba: this is in conflict with the current code which just defaults to extracting a password from the item selected
currently `secrets fetch ITEM` returns a hash {"ITEM" => "password from item"} for login item, otherwise an error
another idea I had was to add it as an arg option `kamal secrets fetch --all-fields ITEM` not sure how feasible that is as it would impact the overall adapter API
another idea is to do make `secrets fetch ITEM` return all the fields as { "FIELD_1" => val, ... } if the item is not a Login item, that would not break compatibility
WDYT?

honzasterba: Since I have received no feedback I have decided that the best way would be to do `kamal secrets fetch ITEM` in case the item is not a login item it will auto-fetch all the fields from that Item

honzasterba:
Please let me know if there is anything I can do to move this along, or if should just give up and close it.

lib/kamal/secrets/adapters/bitwarden.rb
```diff
@@ -25,25 +25,31 @@ def fetch_secrets(secrets, account:, session:)
...
+            fields = item_json["fields"].pluck("name")
+            fetch_secrets_from_fields fields, item, item_json, results
           else
             raise RuntimeError, "Item #{item} is not a login type item and no fields were specified"
           end
         end
       end
     end
 
+    def fetch_secrets_from_fields(fields, item, item_json, results)
```
djmb, line 44: Let's avoid passing the results hash into here, and return a new hash that we can merge into results. It makes it a bit clearer what's happening when calling `fetch_secrets_from_fields`

```ruby
results.merge!(fetch_secrets_from_fields(fields, item, item_json))
```

djmb (changes requested):
Sorry for the delay, @honzasterba, I've been away.

I think the current approach looks good! Just one suggestion about how `fetch_secrets_from_fields` if called and I think we are good to go.

djmb (approved):

honzasterba:
Thank you! 🎆

PR #1000, opened 2024-09-29
Allow specifying multiple hosts for kamal-proxy via an array

kpumuk:
Currently, the way to specify multiple hosts for `kamal-proxy` is via the comma-separated value passed to `host` argument:

```yaml
host: foo.example.com,bar.example.com
```

This leads to frustration when users accidentally enter hosts via more idiomatic array. This MR allows to use both `host` and `hosts`, where the latter accepts an array.

```yaml
hosts:
  - foo.example.com
  - bar.example.com
```

> [!NOTE]  
> This matches the approach taken for accessory hosts: https://kamal-deploy.org/docs/configuration/accessories/#accessory-hosts

Edit: Example of such feedback [here](https://twitter.com/julian_rubisch/status/1840103436752433317).

### DIFF
--- a/lib/kamal/configuration/proxy.rb
+++ b/lib/kamal/configuration/proxy.rb
@@ -28,7 +28,7 @@ class Kamal::Configuration::Proxy
 
   def deploy_options
     {
-      host: hosts.present? ? hosts.join(",") : nil,
+      host: hosts,
       tls: proxy_config["ssl"] ? true : nil,
       "deploy-timeout": seconds_duration(config.deploy_timeout),
       "drain-timeout": seconds_duration(config.drain_timeout),
--- a/lib/kamal/configuration/proxy.rb
+++ b/lib/kamal/configuration/proxy.rb
@@ -22,13 +22,13 @@ class Kamal::Configuration::Proxy
     proxy_config.fetch("ssl", false)
   end
 
-  def host
-    proxy_config["host"]
+  def hosts
+    proxy_config["hosts"] || proxy_config["host"]&.split(",") || []
   end
 
   def deploy_options
     {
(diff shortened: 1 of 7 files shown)


igor-alexandrov (approved):
I think it will be good to deprecate `host` and remove it in 2.1 release. No reason to support both config options.

Everything else looks great. Thank you!

lib/kamal/configuration/proxy.rb
```diff
@@ -22,13 +22,13 @@ def ssl?
...
-  def host
-    proxy_config["host"]
+  def hosts
+    proxy_config["hosts"] || proxy_config["host"]&.split(",") || []
   end
 
   def deploy_options
     {
-      host: proxy_config["host"],
+      host: hosts.present? ? hosts.join(",") : nil,
```
djmb, line 31: We can just pass `hosts` here as kamal-proxy has support for `--host a.example.com --host b.example.com`

djmb (changes requested):
Thanks for this @kpumuk!

We'd added support for multiple hosts in `kamal-proxy` before the release but hadn't had time to get them into the gem before the release of 2.0.

Just one change to pass the hosts as an array and we should be good to go 👍

lib/kamal/configuration/proxy.rb
kpumuk: Good point, thank you!

```shell
$ bin/kamal-proxy deploy nginx --target localhost:8080 --health-check-path / --host example.com --host anotherexample.com

$ bin/kamal-proxy list
Service  Host                            Target          State    TLS
nginx    example.com,anotherexample.com  localhost:8080  running  no
```

Updated the MR.

Issue #1007, opened 2024-09-29
`kamal secrets extract` JSON parse error on shell escaped secrets


mblayman:
Hi!

I'm trying to use the new secret helpers from Kamal 2 to manage secrets for my project. I tried to follow the style listed in [the secrets documentation under "Environment variables"](https://kamal-deploy.org/docs/configuration/environment-variables/#secrets), but I'm running into a problem with shell escaping.

I think what is happening is that the raw secrets JSON data is getting to the Ruby parsing code *with* the shell escape characters still intact. Parsing fails because the JSON parser doesn't accept the backslashes.

I tried to do some `puts` debugging in the `extract` method of `lib/kamal/cli/secrets.rb` to confirm what was going into the `JSON.parse` call. You can see from my output below that the shell escape characters are still present.

```bash
$ SECRETS=$(kamal secrets fetch --adapter 1password --account *** --from some-vault/some-item KAMAL_REGISTRY_PASSWORD)

$ echo $SECRETS
\{\"some-vault/some-item/KAMAL_REGISTRY_PASSWORD\":\"just_alphanumeric_and_underscores\"\}

$ kamal secrets extract KAMAL_REGISTRY_PASSWORD $SECRETS
Hello modified <== my change of `puts "Hello modified"`
\{\"some-vault/some-item/KAMAL_REGISTRY_PASSWORD\":\"just_alphanumeric_and_underscores\"\} <== my change of `puts secrets`
  ERROR (JSON::ParserError): unexpected token at '\{\"some-vault/some-item/KAMAL_REGISTRY_PASSWORD\":\"just_alphanumeric_and_underscores\"\}'
```

I attempted this on bash and zsh to try to rule out differences in shells.

Is there supposed to be a different way to invoke `kamal secrets fetch` or `kamal secrets extract`?

Thanks for the help! I hope this bug report is useful. Please let me know if y'all need any other info.

mblayman:
I'm not really a Ruby dev, so this might be an inappropriate solution, but Chat GPT helped me devise this alternative implementation of `extract`:

```ruby
  desc "extract", "Extract a single secret from the results of a fetch call"
  option :inline, type: :boolean, required: false, hidden: true
  def extract(name, secrets)
    unescaped_secrets = Shellwords.shellsplit(secrets).first
    parsed_secrets = JSON.parse(unescaped_secrets)
    ...
```
If `fetch` consistently outputs a shell escaped JSON blob in a single line, I think this might work well.

[...]

PR #1008, opened 2024-09-30
Escape secrets in inline command substitution

djmb:
Kamal "inlines" calls to `kamal secrets` in the dotenv file, but the results of the calls were not being escaped properly. To "fix" this `kamal secrets fetch` escaped the JSON string before returning it.

The two errors cancelled out, but it meant that the commands didn't work from a shell.

To fix, we'll escape the inline command results and remove the escaping from `kamal secrets fetch`.

Fixes https://github.com/basecamp/kamal/issues/1007

### DIFF
--- a/lib/kamal/cli/secrets.rb
+++ b/lib/kamal/cli/secrets.rb
@@ -7,7 +7,7 @@ class Kamal::Cli::Secrets < Kamal::Cli::Base
   def fetch(*secrets)
     results = adapter(options[:adapter]).fetch(secrets, **options.slice(:account, :from).symbolize_keys)
 
-    return_or_puts JSON.dump(results).shellescape, inline: options[:inline]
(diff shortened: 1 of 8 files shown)


djmb:
No this won't work. 

The `dotenv` gem needs the escaped output from `kamal secrets fetch` function correctly - it's not behaving exactly the same as a shell would. Escaping the output of the "inlined" kamal secrets commands kind of works around that, but it means that the inlined commands don't work the same a a non inlined version would.

So the problem here is that dotenv and the shell are out of sync. I think maybe the best thing to do is add a `kamal secrets print` command that can be used when debugging secrets so you don't need to run the commands directly from the command line.

djmb:
Closing in favour of https://github.com/basecamp/kamal/pull/1009

PR #1019, opened 2024-10-01
Set KAMAL_DESTINATION when loading config

nickhammond:


lib/kamal/configuration.rb
```diff
@@ -20,6 +20,8 @@ class Kamal::Configuration
 
   class << self
     def create_from(config_file:, destination: nil, version: nil)
+      ENV["KAMAL_DESTINATION"] = destination
```
nickhammond, line 23: Is there a preferred way to set this? I know most config is loaded via the `env` helper, just seeing `VERBOSE` set directly with `ENV`. It also needs to be set as early as possible and `create_from` is the entry point basically.

djmb: Here is fine I think 👍

jeromedalbert:
I'm glad to see that this issue is being tackled. I have needed `ENV['KAMAL_DESTINATION']` in my `config/deploy.yml` so I can reference for `Dotenv.load(".kamal/secrets.#{ENV['KAMAL_DESTINATION']}")` or likewise with `Dotenv.parse`.

This PR will remove my workaround of pre-parsing the destination in `bin/kamal` and setting the environment variable then.

Issue #1027, opened 2024-10-02
Can't override `--config-file` in `aliases`


hundredwatt:
I have a secondary service that's deployed from my app's repo via a separate configuration file. 

I can deploy the secondary service with `kamal deploy --config-file=config/other.deploy.yml` and that command succeeds.

I tried to make this command an alias in my main app's `config/deploy.yml`, but I get a registry error because the secondary service uses a different registry but kamal tries to use the main app's registry to pull the image:

```
docker stderr: Unable to find image main.myregistry.com/secondary:latest' locally
```

When in `config/other.deploy.yml`, I have the `registry.server` set to `other.myregistry.com`.

My alias configuration in `config/deploy.yml` looks like:

```yaml
aliases:
  "other-deploy": "deploy --config-file=config/other.deploy.yml"
```

If I'm using kamal in an unintended way please tell me 😅. But it would be nice if I could configure an alias to deploy my secondary service so I don't have to add `--config-file=config/other.deploy.yml` every time.

djmb:
Yes that's not something that was supported, but making it work turned out to be easy enough - https://github.com/basecamp/kamal/pull/1032

PR #1032, opened 2024-10-02
Allow destination and config-file in aliases

djmb:
We only loaded the configuration once, which meant that aliases always used the initial configuration file and destination.

We don't want to load the configuration in subcommands as it is not passed all the options we need. But just checking if we are in a subcommand is enough - the alias reloads and the subcommand does not.

One thing to note is that anything passed on the command line overrides what is in the alias, so if an alias says
`other_config: config -c config/deploy2.yml` and you run `kamal other_config -c config/deploy.yml`, it won't switch.

Fixes: https://github.com/basecamp/kamal/issues/1027

### DIFF
--- a/lib/kamal/cli/base.rb
+++ b/lib/kamal/cli/base.rb
@@ -30,7 +30,8 @@ module Kamal::Cli
       else
         super
       end
-      initialize_commander unless KAMAL.configured?
+
+      initialize_commander unless config[:invoked_via_subcommand]
     end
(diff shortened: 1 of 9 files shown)


djmb:
Should also fix https://github.com/basecamp/kamal/issues/1237, https://github.com/basecamp/kamal/issues/1300 and https://github.com/basecamp/kamal/issues/1316

PR #1038, opened 2024-10-03
Kamal::Secrets did not load environment variables in .env on my Apple M1

vladiim:
I consistently got error messages trying to login to docker on my Apple M1. The issue was .env variables were not being passed correctly to kamal.

### DIFF
--- a/lib/kamal/secrets.rb
+++ b/lib/kamal/secrets.rb
@@ -5,10 +5,13 @@ class Kamal::Secrets
 
   Kamal::Secrets::Dotenv::InlineCommandSubstitution.install!
 
+  attr_reader :secrets
   def initialize(destination: nil)
     @secrets_files = \
       [ ".kamal/secrets-common", ".kamal/secrets#{(".#{destination}" if destination)}" ].select { |f| File.exist?(f) }
     @mutex = Mutex.new
+    ::Dotenv.load
+    @secrets = set_secrets
   end
 
   def [](key)
@@ -29,7 +32,7 @@ class Kamal::Secrets
   end
 
   private
-    def secrets
+    def set_secrets
       @secrets ||= secrets_files.inject({}) do |secrets, secrets_file|
         secrets.merge!(::Dotenv.parse(secrets_file))
       end
--- a/test/secrets_test.rb
+++ b/test/secrets_test.rb
@@ -31,4 +31,25 @@ class SecretsTest < ActiveSupport::TestCase
       assert_equal "JKL", Kamal::Secrets.new(destination: "nodest")["SECRET2"]
     end
   end
+
+  test "dotenv load" do
+    create_and_remove_secret_files do
+      secrets = Kamal::Secrets.new
+      assert_equal @value, secrets[@key]
+    end
+  end
+
+  private
+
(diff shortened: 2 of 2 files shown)


nickhammond:
Kamal intentionally doesn't read from `.env` automatically anymore. https://kamal-deploy.org/docs/upgrading/secrets-changes/

junket:
@nickhammond I understand the change but one thing I haven't found any advice on:  Where should the actual secrets (meaning: the plain text values) be kept if not in .env or in .kamal/secrets? I'm currently putting them in .kamal/secrets and keeping that gitignored but that seems not to be the intended design. Thanks for any light you can shed! Kamal is a joy.

nickhammond:
@junket I'm still sort of figuring out my workflow as well since I don't often want to deal with fetching the credentials from 1P, it can just add another layer.

The intention seems to be that you always fetch from a secret store and that's what gets set dynamically in `.kamal/secrets`. That's not always the case though and sometimes it's just easier to deal with files for some apps, pre Kamal 2 style. 

What I've been doing is ignoring from git `.kamal/secrets-common` and `.kamal/secrets.*`, `.kamal/secrets` just contains example/template config with empty values.

If you wanted to keep on using `.env` you can use the `dotenv` CLI command to pre-load the env before running Kamal commands but this feels pretty clunky.

junket:
Excellent. Thanks @nickhammond! Exactly, I haven't bothered setting up secret stores for any of my one-man projects for the reasons you mentioned.

Okay, I'll follow your approach, too--it does make sense to me to just have just one .env for dev and all the deployment/destination stuff in with the Kamal config. But when I saw the advice in the `.kamal/secrets` boilerplate not to actually add the secrets there I was stymied.

vladiim:
Thanks so much for the additional info @nickhammond - sorry I didn't catch the docs!

nickhammond:
Added an idea to autoload here as well: https://github.com/basecamp/kamal/discussions/1052

vladiim:
I ended up working with lastpass, just a note that laspass' default output didn't work with `kamal secrets extract` I needed to wrap the <SECRETS-FETCH-OUTPUT> in "\"${<SECRETS-FETCH-OUTPUT>}\"" might be worth updating the docs.

kokorolx:
I'm facing the same issue, the RSA key does not wrap in `"` and it does not work correctly.

kokorolx:
> I'm facing the same issue, the RSA key does not wrap in `"` and it does not work correctly.
Here's the error: `initialize': Neither PUB key nor PRIV key (OpenSSL::PKey::RSAError)`

My workaround: use `gsub` to replace `\\n` to `\n`

`OpenSSL::PKey::RSA.new(_private_key.gsub(/\\n/, "\n"))`

djmb:
@junket - the default path is to have `.kamal/secrets` in source control and to load the secrets from a password manager/secret store at runtime. 

But removing `.kamal/secrets` from source control and putting the secrets directly in there will work if that's what you prefer.

I'm going to close this PR though as we won't be using `.env` anymore.

@vladiim, @kokorolx - if either of you had time it would be great if you could raise an issue/PR for the lastpass problems you are having.

Issue #1046, opened 2024-10-03
Expose `pause` and `resume` proxy option in Kamal


baldarn:
Hi! based on issues https://github.com/basecamp/kamal-proxy/issues/4 and comments https://github.com/basecamp/kamal-proxy/issues/4#issuecomment-2354231228 https://github.com/basecamp/kamal-proxy/issues/8#issuecomment-2354236607

I would like to have the exposure of `pause` and `resume` methods of `kamal-proxy`

dhh:
We are going to tweak these, then expose.

Issue #162, opened 2023-03-27
Suggestion: maintenance mode


rfunduk:
Something that is occasionally handy (when running a long/app-breaking migration, upgrading the database, making some fundamental change to configuration, or in the event of a security incident perhaps) is the ability to quickly lock down the application with a 503 and a 'be back soon' page.

Used to use nginx/apache reconfig with a local file for this (e.g. via something like https://github.com/capistrano/maintenance), but more recently I've been configuring AWS ALB to respond with a static blob (ex. [using terraform variables](https://github.com/serverless-rails/serverless-rails-demo/blob/9d395248ffb2f21aadb5510350075cbca128d6e5/tf/alb.tf#L77-L95)).

Doing it via the load balancer is probably not a very portable solution (Digital Ocean doesn't appear to support it)... maybe traefik entrypoints could be used for this.

dhh:
Would love to see this explored. We currently do something like this, but it's Ruby specific:

```ruby
if ENV["MAINTENANCE"]
  require "pathname"
  MAINTENANCE_HTML = Pathname.new(File.dirname(__FILE__)).join("public/maintenance.html").read

  run lambda { |env| [ 200, { "Content-Type" => "text/html" }, [ MAINTENANCE_HTML ] ] }
else
  ... 
end
```

And then set `MAINTENANCE` as an ENV in deploy. But I could totally see injecting this as a pre-baked image of some kind. I just don't want to pay a very high complexity price for it. So it's gotta be approximately as simple as what's above.

rfunduk:
Ah, you do this in `config.ru`, and just bypass the whole rails app to serve the maintenance page? That seems pretty straight-forward (and clever!). Personally I think I'd be happy with that.

I agree that it seems like it might take some ugliness to bake this in to mrsk with some kind of sidecar container or traefik hacks. Maybe I'll attempt it, just to compare. It would be cool to have a built-in solution.

dhh:
I think actually a basic solution here could be to have a mrsk-owned image that just does what ENV["MAINTENANCE"] does. Injects a Rack app that runs that lambda, with a way to overwrite the maintenance.html screen, and swaps that container in instead of the app one when you go into maintenance mode. Might be pretty simple. Do investigate if you like!

rfunduk:
I don't think a mrsk-owned image is really necessary since this basically covers what's needed:

```
[...]

PR #1047, opened 2024-10-03
Expose `pause` and `resume` methods of `kamal-proxy`

baldarn:
I created the issue to be addressed https://github.com/basecamp/kamal/issues/1046

The idea is to expose the `pause` and `resume` commands of the `kamal-proxy`

## Proxy Interface edits

This pr enables that via the some new commands in the `proxy` interface:
- `pause_app` will triggers the `pause` command in all kamal proxy containers
- `resume_app` will triggers the `resume` command in all kamal proxy containers

## Concerns

I feel like the presence in the `proxy` interface is not the best option, what do you think? i feel like this could go in the `app` interface, not really sure

## Tests
tests of all levels are there. i found some old ugly tests in the integration part, but managed to have something running ;)

this could be useful to address a `maintenance mode` as requested here https://github.com/basecamp/kamal/issues/162

## Docs
docs pr https://github.com/basecamp/kamal-site/pull/122

### DIFF
--- a/lib/kamal/commands/proxy.rb
+++ b/lib/kamal/commands/proxy.rb
@@ -16,8 +16,16 @@ class Kamal::Commands::Proxy < Kamal::Commands::Base
     docker :container, :start, container_name
   end
 
-  def stop(name: container_name)
-    docker :container, :stop, name
+  def stop
+    docker :container, :stop, container_name
+  end
+
+  def pause_app
+    docker :exec, container_name, "kamal-proxy", "pause", "#{config.service}-web"
+  end
+
+  def resume_app
+    docker :exec, container_name, "kamal-proxy", "resume", "#{config.service}-web"
   end
 
   def start_or_run
--- a/lib/kamal/commands/proxy.rb
+++ b/lib/kamal/commands/proxy.rb
@@ -21,11 +21,11 @@ class Kamal::Commands::Proxy < Kamal::Commands::Base
   end
 
   def pause_app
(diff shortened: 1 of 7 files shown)


lib/kamal/commands/proxy.rb
```diff
@@ -16,8 +16,16 @@ def start
...
   end
 
-  def stop(name: container_name)
-    docker :container, :stop, name
+  def stop
+    docker :container, :stop, container_name
+  end
+
+  def pause_app
+    docker :exec, container_name, "kamal-proxy", "pause", "#{config.service}-web"
```
nickhammond, line 24: This should probably use the `primary_role_name`, there might be a helper method already too that's service + primary role.

baldarn: @nickhammond is this what you intended?

https://github.com/basecamp/kamal/pull/1047/commits/693328cf5bc9ea3209a4920d0c78f3e51537ee46

nickhammond (approved):
Yep! The default is `web` which is why the spec automatically works but you could have the primary role be `primary` or something else. https://github.com/basecamp/kamal/blob/81f3508507511a40cfc52eb5eeb4f9cf0d8ecc44/lib/kamal/configuration/docs/role.yml#L6

dhh:
We are going to change these shortly in kamal proxy, and then expose after that. The command names are getting a refresh. So let's close this until that lands.

konung:
> Just came across this while looking for a way to use a "maintenance mode" with Kamal, is the plan still to change command names?

Also was just looking for exactly the same thing :D

PR #1050, opened 2024-10-04
Update sample template for docker setup hook.

morgoth:
"kamal" network is already created (in v2.0) so the sample code is no longer accurate.

Replacing with simple echo statement, as in few other samples.

If anyone has a real case sample that we could put in the file instead, then let me know.
cc @tsvallender as you're the author (https://github.com/basecamp/kamal/pull/585)


nickhammond:
What about a `kamal server exec "docker version"`?

dhh:
Happy to take a more realistic example too. Not sure docker version is it.

PR #1057, opened 2024-10-05
Remove extra empty lines in description

JenteSL:
This PR removes the extra empty lines in `deploy.yml` that are present in the description of the sections that are commented out. This brings it on par with the sections that aren’t commented out, and the template that will be shipped in Rails 8.


dhh:
I prefer it with the spacing and we should instead add that to the Rails version if it doesn't have it yet.

Issue #36, opened 2023-02-04
Server monitoring


scart88:
What is the best way to add monitoring to MRSK?

Coming from K8s, it was easy to add `kube-prometheus-stack` and get Prometheus and Grafana out of the box, so I was wondering what would be the best way to add something like that to MRSK?

dhh:
That's outside the scope of MRSK. But could be a neat sister project!

PR #1064, opened 2024-10-06
Update to kamal-proxy 0.8.0

djmb:
Proxy changes:
- Add option to use custom TLS certificates (#17)
- Don't buffer SSE responses (#36)
- Allow routing to wildcard subdomains (#45)

Custom TLS certificates not supported in Kamal itself yet. Buffering SSE responses and wildcard subdomains will work without any Kamal changes.

### DIFF
--- a/lib/kamal/configuration.rb
+++ b/lib/kamal/configuration.rb
@@ -14,7 +14,7 @@ class Kamal::Configuration
 
   include Validation
 
-  PROXY_MINIMUM_VERSION = "v0.7.0"
+  PROXY_MINIMUM_VERSION = "v0.8.0"
   PROXY_HTTP_PORT = 80
   PROXY_HTTPS_PORT = 443
 


karls:
I didn't want to create a new issue, so commenting in here. Am I correct that currently there is no way to specify the `kamal-proxy` version without a code change in Kamal itself?

tuladhar:
> I didn't want to create a new issue, so commenting in here. Am I correct that currently there is no way to specify the `kamal-proxy` version without a code change in Kamal itself?

@karls That's correct, it's currently hardcoded in the code. 

https://github.com/basecamp/kamal/blob/e34031f70cf933384fe0467ade20d6653f3ea908/lib/kamal/configuration.rb#L17

tuladhar:
> @karls I was about to ask the same question, you beat me to it :D. FWIW: I was looking in the CLI commands for a `kamal proxy upgrade`.

I discovered that you can upgrade the Kamal Proxy using the following command:
```
kamal proxy reboot
```

If it's running old version, and new version is available, then it will upgrade, e.g:
>  ERROR (RuntimeError): Exception while executing on host certsdiscount-app1: kamal-proxy version v0.6.0 is too old, please reboot to update to at least v0.7.0

paganotoni:
thanks @tuladhar. I tried rebooting. however, it did not work.

tuladhar:
> thanks @tuladhar. I tried rebooting. however, it did not work.

Same here. I tried rebooting again, but it didn’t update beyond what’s hardcoded in the code.

djmb:
Yes right now the version is fixed entirely. But I think we'll allow updating to later versions, or running your own image if you want to fork kamal-proxy

Issue #1036, opened 2024-10-03
SSHKit does not escape command arguments, resulting in failure to clone a path with spaces


davidstosik:
I understand this might be unconventional and frowned upon in some circles, but I store my whole source code folder in iCloud and use a symlink for convenience. I don't really have to justify, but that helps me with instant backups and when working with multiple computers. (I expect "use Git" comments, and that's okay. 😬 )

```sh-session
$ cd ~/src/my-project && pwd -P
/Users/sto/Library/Mobile Documents/com~apple~CloudDocs/src/my-project
```

So far, I have had very little issues with this quirk in my setup.

Enters Kamal. Since the announcements and talks at RailsWorld 2024, I've been very excited about trying it out on some pet projects.

I've ignored the issue at first and managed to deploy a pet project to DigitalOcean when my project's path didn't include spaces, but I thought it's now time to report this issue.

When running `kamal deploy` in a project which path includes spaces, Kamal fails to clone the repository:

```sh-session
$ kamal deploy
...                                                                                                                                                       unescaped space here!
                                                                                                                                                                    ⬇️
  INFO [a007f48f] Running /usr/bin/env git -C /var/folders/1q/mcsb93312b73zl6fm0887wzc0000gn/T/kamal-clones/my-project-a97a49be96920 clone /Users/sto/Library/Mobile Documents/com~apple~CloudDocs/src/my-project --recurse-submodules as sto@localhost
  Finished all in 255.7 seconds
  ERROR (SSHKit::Command::Failed): git exit status: 32768
git stdout: Nothing written
git stderr: fatal: repository '/Users/sto/Library/Mobile' does not exist
```

I dug a little bit and it seems related to how SSHKit sometimes constructs commands:
- `SSHKit::Backend::Local#execute_command` calls `Open3.popen3(cmd.to_command)` [here](https://github.com/capistrano/sshkit/blob/c3815c2229e782091caa8dd511969fae4dcbf7c3/lib/sshkit/backends/local.rb#L44)
- `SSHKit::Command.to_command` eventually calls `#to_s` and joins all items in the command array without escaping to produce a string [here](https://github.com/capistrano/sshkit/blob/c3815c2229e782091caa8dd511969fae4dcbf7c3/lib/sshkit/command.rb#L216)
- Instead, `Open3.popen3` could be passed an array and it would automatically escape spaces where needed:
   ```rb
    require "open3"
[...]

PR #1066, opened 2024-10-06
Support spaces in git repository path

davidstosik:
Fixes https://github.com/basecamp/kamal/issues/1036

~~I haven't written any tests for this though I'd be interested to try.~~

### DIFF
--- a/lib/kamal/commands/builder/clone.rb
+++ b/lib/kamal/commands/builder/clone.rb
@@ -1,33 +1,31 @@
 module Kamal::Commands::Builder::Clone
-  extend ActiveSupport::Concern
-
-  included do
-    delegate :clone_directory, :build_directory, to: :"config.builder"
-  end
-
   def clone
-    git :clone, escaped_root, "--recurse-submodules", path: clone_directory
+    git :clone, escaped_root, "--recurse-submodules", path: config.builder.clone_directory.shellescape
   end
 
   def clone_reset_steps
     [
-      git(:remote, "set-url", :origin, escaped_root, path: build_directory),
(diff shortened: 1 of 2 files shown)


lib/kamal/commands/builder/clone.rb
```diff
@@ -1,29 +1,31 @@
 module Kamal::Commands::Builder::Clone
-  extend ActiveSupport::Concern
-
-  included do
-    delegate :clone_directory, :build_directory, to: :"config.builder"
-  end
```
davidstosik, line 6: This delegation becomes superfluous when we need to escape the values returned by `clone_directory` and `build_directory`, so I just removed it. Inlined `config.builder.clone_directory.shellescape` in the single location it's needed, and factorized `config.builder.build_directory.shellescape` in the `escaped_build_directory` method.

test/commands/builder_test.rb
```diff
@@ -149,15 +149,26 @@ class CommandsBuilderTest < ActiveSupport::TestCase
...
+
   private
     def new_builder_command(additional_config = {})
       Kamal::Commands::Builder.new(Kamal::Configuration.new(@config.deep_merge(additional_config), version: "123"))
     end
 
-    def build_directory
-      "#{Dir.tmpdir}/kamal-clones/app/kamal/"
-    end
-
```
davidstosik, line 160: I wondered what this was, and realized it has become unnecessary when tests still passed without it.

davidstosik:
cc @djmb

djmb:
Thanks @davidstosik!

PR #1070, opened 2024-10-07
Add `kamal app open` command to open app in a web browser

jeromedalbert:
## Problem

Kamal is a no-PaaS tool with PaaS-like ergonomics, where you can open a production Rails (or other) console, a shell session, tail logs, etc. One of the things that I've been missing is a command to open the app in a web browser. That is, a Kamal equivalent to Heroku's `heroku open` or Fly.io's `fly apps open`.

This would allow for quick access to a deployed web application, from the CLI, without having to manually copy/paste the URL from one of your browser tabs or your `config/deploy.yml`. Examples where this can be useful:

- right after deploying your app
- when switching between different Kamal-enabled repos
- in copy/paste-able Readme instructions
- any time you're in the command line and don't want to switch context

## Solution

Add a `kamal app open` command. It opens the primary role (usually `web`) first domain defined in `proxy` `host(s)`, or falls back to the primary role's first IP address.

The implementation is so short that adding a dependency for finding the right opening command (with a gem like launchy) felt like too much. But I can change the PR if needed.

### DIFF
--- a/lib/kamal/cli/app.rb
+++ b/lib/kamal/cli/app.rb
@@ -227,6 +227,24 @@ class Kamal::Cli::App < Kamal::Cli::Base
     end
   end
 
+  desc "open", "open app in a web browser"
+  def open
+    host = KAMAL.primary_role.proxy.hosts.first || KAMAL.primary_host
+    protocol = KAMAL.primary_role.ssl? ? "https" : "http"
+    cmd =
+      case RbConfig::CONFIG["host_os"]
+      when /mswin|mingw|cygwin/ then "start"
+      when /darwin/ then "open"
+      when /linux|bsd/ then "xdg-open"
+      end
+
+    if cmd
+      system(cmd, "#{protocol}://#{host}")
+    else
+      say "Could not detect OS browser"
+    end
+  end
+
   desc "remove", "Remove app containers and images from servers"
   def remove
     with_lock do
--- a/test/cli/app_test.rb
+++ b/test/cli/app_test.rb
@@ -235,6 +235,20 @@ class CliAppTest < CliTestCase
     end
   end
 
+  test "open with proxy host domain" do
+    RbConfig::CONFIG.stubs(:[]).with("host_os").returns("linux")
+    Object.any_instance.expects(:system).with("xdg-open", "https://app.example.com")
+
+    run_command("open", config: :with_proxy_host)
+  end
+
+  test "open with primary host IP" do
+    RbConfig::CONFIG.stubs(:[]).with("host_os").returns("linux")
+    Object.any_instance.expects(:system).with("xdg-open", "http://1.1.1.1")
+
+    run_command("open")
+  end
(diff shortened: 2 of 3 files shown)


lib/kamal/cli/app.rb
```diff
@@ -227,6 +227,24 @@ def logs
     end
   end
 
+  desc "open", "open app in a web browser"
+  def open
+    host = KAMAL.primary_role.proxy.hosts.first || KAMAL.primary_host
+    protocol = KAMAL.primary_role.ssl? ? "https" : "http"
+    cmd =
+      case RbConfig::CONFIG["host_os"]
+      when /mswin|mingw|cygwin/ then "start"
```
tuladhar, line 236: Does Kamal work on Windows? I didn't find it mentioned in the  [Installation section](https://kamal-deploy.org/docs/installation/). If it does, maybe we could add. I haven't tried it yet.

jeromedalbert: I just spun up a Windows VM, I don't know if the whole Kamal deploy flow works (I wasn't able to fully `bundle install` Rails), but `kamal apps open` by itself did work.

djmb:
I like this, but I'm not sure. 

The command will not work properly in lots of setups - where there's no proxy host, where the host if a wildcard etc. 
Adding a command that's often broken seems a bit iffy.

jeromedalbert:
Normally `kamal app open` wouldn't break because it would fall back gracefully.

For example when there is no proxy host, this PR already falls back to the primary IP (see unit test). This is pretty handy as I often like to create short-lived apps where I don't bother to configure SSL, so I access these apps by IP.

Normal non-wildcard proxy hosts also work.

That leaves us with wildcards proxy hosts. Is this an issue though? I don't think Kamal currently supports them in the `proxy>hosts` key unless I am missing something. But if it does `kamal app open` could just find the first non-wildcard or doing the usual fallback to the primary IP.

So I think that covers it, unless you have other cases in mind?

AxelTheGerman:
You would have to duplicate the host name, but this would also be possible using [Aliases](https://kamal-deploy.org/docs/configuration/aliases/) on a per project basis.

jeromedalbert:
@AxelTheGerman Can you provide an example? Aliases are for things like `app exec` which execute on the server whereas this command executes on your local.

AxelTheGerman:
> @AxelTheGerman I am not sure how this would be possible. Can you provide an example? AFAIK aliases are for things like `app exec` which execute on the server whereas this command executes on your local.

Duh you're right... but in that case I guess it would be possible to A) write a custom rake task or B) creating a kamal-open gem that could read and parse the deploy yaml file as well.

It does seem much easier to add it into kamal but I'm agreeing with @djmb regarding

> The command will not work properly in lots of setups


One option though for implementing it in kamal would just be to print a message that the app cannot be opened if we can't find the host/IP for the app

Issue #1065, opened 2024-10-06
Bitwarden-Adapter: "bw" cli not found


bjesuiter:
## What I want

Being able to fetch secrets via the bitwarden adapter
```
kamal secrets fetch --adapter bitwarden --account jb-web-services@vserv.tech --from Kamal/DOCKER_REGISTRY_PASSWORD
```

## My System 
- MacOS Sequoia 15.0.0 
- Mac Book Pro with M3 Max Processor

## Try 1: Running the command via the docker alias 

```bash
alias kamal='docker run -it --rm -v "${PWD}:/workdir" -v "/run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock" -e SSH_AUTH_SOCK="/run/host-services/ssh-auth.sock" -v /var/run/docker.sock:/var/run/docker.sock ghcr.io/basecamp/kamal:latest'

kamal secrets fetch --adapter bitwarden --account jb-web-services@vserv.tech --from Kamal/DOCKER_REGISTRY_PASSWORD
```

```bash
 ERROR (Errno::ENOENT): No such file or directory - bw
```

=> The bw cli is probably not present in the kamal docker image

## Try 2: mount my local bw cli into the kamal docker 

by using `-v $HOME/bin/bw:/usr/local/bin/bw`

```bash 
alias kamal-bw='docker run -it --rm -v "${PWD}:/workdir" -v "/run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock" -e SSH_AUTH_SOCK="/run/host-services/ssh-auth.sock" -v /var/run/docker.sock:/var/run/docker.sock -v $HOME/bin/bw:/usr/local/bin/bw ghcr.io/basecamp/kamal:latest'

kamal secrets fetch --adapter bitwarden --account jb-web-services@vserv.tech --from Kamal/DOCKER_REGISTRY_PASSWORD
```

```bash
Cannot run macOS (Mach-O) executable in Docker: Exec format error
  ERROR (JSON::ParserError): unexpected token at ''
```

=> I probably cannot run my macOS style bw executable in my linux style kamal docker 

## Try 3 - installing kamal as ruby gem 

Problem 1: The default ruby version of my system is `ruby 2.6.10p210 (2022-04-12 revision 67958) [universal.arm64e-darwin24]` => too old for some dependencies

Problem 2: Installing newer ruby via "internet-recommended" rvm fails, because compilation fails for some reason for Mac 15.0/Arm64

[...]

PR #1075, opened 2024-10-08
Dockerfile: Include CLI binaries for supported password managers

tuladhar:
Towards https://github.com/basecamp/kamal/issues/1065

## Changes
- Update `Dockerfile` by including CLI binaries for [supported password managers](https://kamal-deploy.org/docs/commands/secrets/) (currently 1Password, LastPass, and Bitwarden)

## Image size impact: +114MB
 Before:
```
docker pull ghcr.io/basecamp/kamal:latest
docker images ghcr.io/basecamp/kamal:latest
REPOSITORY               TAG       IMAGE ID       CREATED        SIZE
ghcr.io/basecamp/kamal   latest    2c8810e720d9   36 hours ago   674MB
```
After: 
```
> docker pull ptuladhar/kamal:latest
> docker images ptuladhar/kamal:latest
REPOSITORY        TAG       IMAGE ID       CREATED          SIZE
ptuladhar/kamal   latest    690384d22c81   32 minutes ago   788MB
```

Versions:
```
docker run -it --entrypoint sh ptuladhar/kamal:latest
/workdir # lpass --version
LastPass CLI v1.3.3.GIT
/workdir # op --version
2.30.0
/workdir # bw --version
Could not find dir, "/root/.config/Bitwarden CLI"; creating it instead.
Could not find data file, "/root/.config/Bitwarden CLI/data.json"; creating it instead.
2024.9.0
```

## Idea:
- Maybe, move this enhance to `Dockerfile.extended` and publish the image separately as `docker pull ghcr.io/basecamp/kamal:extended`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,9 @@ FROM ruby:3.2.0-alpine
 # Install docker/buildx-bin
 COPY --from=docker/buildx-bin /buildx /usr/libexec/docker/cli-plugins/docker-buildx
 
+# Install 1password/op
+COPY --from=1password/op:2 /usr/local/bin/op /usr/local/bin/op
+
 # Set the working directory to /kamal
 WORKDIR /kamal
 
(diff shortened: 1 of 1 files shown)


djmb:
We could add these binaries, but there's more to it than that. The docker container that runs Kamal doesn't have any password manager configuration in it, so for example with 1Password I need to enter the full signin details every time (1password account, email address, secret key, password).

Then each manager/secret store will have its way to configure itself and its own challenges. As we add more adapters we'll have more to deal with those as well, and we'll bloat the image with more binaries.

Run kamal via a docker image has many drawbacks and I'm not sure we should be recommending it anymore. Though installing it directly is certainly annoying for users who don't want to deal with Ruby. Shame there doesn't seem to be a simple reliable method for generating standalone binaries for Ruby gems.

PR #1099, opened 2024-10-11
feat(secrets): add Doppler adapter

mrbongiolo:
Doppler organizes secrets in "projects" (like `my-awesome-project`) and "configs" (like `prod`, `stg`, etc.), so the pattern `project/config` is required when defining the `--account` option.

Doppler does not have a concept of folders, so using `--from` option or `FOLDER/SECRET` pattern is not supported and will raise an error.

Initially, I was thinking of using the `--from` option to define the Doppler "config", but this would require a much larger change in how secrets are "folderized", so not allowing to use that option and parsing `--account=project/config` accordingly seemed like a better approach for now.

Site documentation PR: https://github.com/basecamp/kamal-site/pull/134

### DIFF
--- a/lib/kamal/secrets/adapters/doppler.rb
+++ b/lib/kamal/secrets/adapters/doppler.rb
@@ -1,28 +1,53 @@
 class Kamal::Secrets::Adapters::Doppler < Kamal::Secrets::Adapters::Base
+  def requires_account?
+    false
+  end
+
   private
-    def login(account)
-      unless loggedin?(account)
+    def login(*)
+      unless loggedin?
         `doppler login -y`
         raise RuntimeError, "Failed to login to Doppler" unless $?.success?
       end
     end
 
-    def loggedin?(account)
+    def loggedin?
       `doppler me --json 2> /dev/null`
       $?.success?
     end
 
-    def fetch_secrets(secrets, account:, session:)
-      project, config = account.split("/")
+    def fetch_secrets(secrets, **)
+      project_and_config_flags = ""
+      unless service_token_set?
+        project, config, _ = secrets.first.split("/")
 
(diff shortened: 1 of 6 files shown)


lib/kamal/secrets/adapters/doppler.rb
```diff
@@ -0,0 +1,28 @@
...
+      $?.success?
+    end
+
+    def fetch_secrets(secrets, account:, session:)
+      project, config = account.split("/")
+
+      raise RuntimeError, "Missing project or config from --acount=project/config option" unless project && config
+      raise RuntimeError, "Using --from option or FOLDER/SECRET is not supported by Doppler" if secrets.any?(/\//)
+
+      items = `doppler secrets get #{secrets.map(&:shellescape).join(" ")} --json -p #{project} -c #{config}`
```
djmb, line 21: We should shell escape `project` and `config` here as well as we are injecting user input. Admittedly probably a friendly user, but best to be on the safe side.

djmb:
Oh, could we also implement the `check_dependencies!` method that was recently added. It should give a nice error message if doppler is not installed.

mrbongiolo:
Thanks for the review @djmb!! I'll fix the points that you brought up and update the PR later today.

mrbongiolo:
@djmb PR updated. Also added support for `ENV["DOPPLER_TOKEN"]`, based on https://github.com/basecamp/kamal/pull/1183

djmb:
Hey @mrbongiolo - this has been released now in Kamal 2.4.0. Could you update the docs PR and then I'll merge that?

mrbongiolo:
Nice @djmb !! The doc PR is updated, it just requires an approval for merging. Or is there something missing?

djmb:
No all good then, there were some merge conflicts but I've resolved them and merged the PR 👍

PR #1114, opened 2024-10-15
prevent escape '#' when generating env_file string

alanoliveira:
This PR prevents escape `#` character when generating the string used by docker `--env` argument.
The `#` was escaped because `String#dump` also consider ruby interpolation sequences as special characters, so a secret set as `PASSWORD=foo#$123` changed to `PASSWORD=foo\#$123` inside the container.

### DIFF
--- a/test/env_file_test.rb
+++ b/test/env_file_test.rb
@@ -11,6 +11,16 @@ class EnvFileTest < ActiveSupport::TestCase
       Kamal::EnvFile.new(env).to_s
   end
 
+  test "to_s won't escape '#'" do
+    env = {
+      "foo" => "\#$foo",
+      "bar" => "\#{bar}"
+    }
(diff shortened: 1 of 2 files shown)


test/env_file_test.rb
```diff
@@ -11,6 +11,16 @@ class EnvFileTest < ActiveSupport::TestCase
       Kamal::EnvFile.new(env).to_s
   end
 
+  test "to_s won't escape '#'" do
+    env = {
+      "foo" => "\#$foo",
```
djmb, line 16: Can you change these to `#` instead of `\#`? They mean the same thing and the `\` makes it seem like there's some extra escaping going on that isn't really there.

alanoliveira: done!

PR #1117, opened 2024-10-15
Added `kamal proxy reboot` to raised error message

pardeyke:
When proxy needs to be rebooted because of new version the error raised says:
`kamal-proxy version v0.7.0 is too old, please reboot to update to at least v0.8.1`

I think we can skip a look in the docs if we include the reboot command in this error message like this:
`kamal-proxy version v0.7.0 is too old, please reboot to update to at least v0.8.1. Run "kamal proxy reboot".`


Sija:
Maybe `kamal-proxy version v0.7.0 is too old, Run "kamal proxy reboot" in order to update to at least v0.8.1.`?

PR #1141, opened 2024-10-21
feat: add secrets adapter for aws secrets manager

justindell:
This add [AWS Secrets Manager](https://docs.aws.amazon.com/secretsmanager/latest/userguide/intro.html) as a secrets adapter. It assumes the AWS CLI is installed and has access to secrets manager, so no login step is required. Instead, the ```--account``` option is used to select a profile name, which is typically `default`.

Site documentation PR: [documentation for aws secrets manager ](https://github.com/basecamp/kamal-site/pull/139)

### DIFF
--- a/lib/kamal/secrets/adapters/aws_secretsmanager.rb
+++ b/lib/kamal/secrets/adapters/aws_secretsmanager.rb
@@ -1,0 +1,25 @@
+class Kamal::Secrets::Adapters::AwsSecretsmanager < Kamal::Secrets::Adapters::Base
+  private
+    def login(_account)
+      nil
+    end
+
+    def fetch_secrets(secrets, account:, session:)
+      {}.tap do |results|
+        JSON.parse(get_from_secrets_manager(secrets, account: account))["SecretValues"].each do |secret|
+          secret_name = secret["Name"]
+          secret_string = JSON.parse(secret["SecretString"])
+
+          secret_string.each do |key, value|
+            results["#{secret_name}/#{key}"] = value
+          end
+        end
+      end
+    end
+
+    def get_from_secrets_manager(secrets, account:)
+      `aws secretsmanager batch-get-secret-value --secret-id-list #{secrets.map(&:shellescape).join(" ")} --profile #{account}`.tap do
+        raise RuntimeError, "Could not read #{secret} from AWS Secrets Manager" unless $?.success?
+      end
+    end
+end
--- a/test/secrets/aws_secretsmanager_adapter_test.rb
+++ b/test/secrets/aws_secretsmanager_adapter_test.rb
@@ -1,0 +1,87 @@
+require "test_helper"
+
+class AwsSecretsmanagerAdapterTest < SecretAdapterTestCase
+  test "fetch" do
+    stub_ticks
+      .with("aws secretsmanager batch-get-secret-value --secret-id-list secret/KEY1 secret/KEY2 secret2/KEY3 --profile default")
+      .returns(<<~JSON)
+        {
+          "SecretValues": [
+            {
+              "ARN": "arn:aws:secretsmanager:us-east-1:aaaaaaaaaaaa:secret:secret",
+              "Name": "secret",
+              "VersionId": "vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv",
+              "SecretString": "{\\"KEY1\\":\\"VALUE1\\", \\"KEY2\\":\\"VALUE2\\"}",
+              "VersionStages": [
+                  "AWSCURRENT"
+              ],
+              "CreatedDate": "2024-01-01T00:00:00.000000"
+            },
+            {
+              "ARN": "arn:aws:secretsmanager:us-east-1:aaaaaaaaaaaa:secret:secret2",
+              "Name": "secret2",
+              "VersionId": "vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv",
+              "SecretString": "{\\"KEY3\\":\\"VALUE3\\"}",
+              "VersionStages": [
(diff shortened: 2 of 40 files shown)


lib/kamal/secrets/adapters/aws_secretsmanager.rb
```diff
@@ -0,0 +1,25 @@
+class Kamal::Secrets::Adapters::AwsSecretsmanager < Kamal::Secrets::Adapters::Base
```
djmb, line 1: The offical name is `AWS Secrets Manager`, so let's call this `AwsSecretsManager` (and rename the file to `aws_secrets_manager.rb`

lib/kamal/secrets/adapters/aws_secretsmanager.rb
```diff
@@ -0,0 +1,25 @@
...
+  private
+    def login(_account)
+      nil
```
djmb, line 4: There's no sensible default way to implement this is there? Would be nice if we could prompt for MFA from here if that was enabled, but we'd need to know all about the IAM setup for that to work I think.

test/secrets/aws_secretsmanager_adapter_test.rb
```diff
@@ -0,0 +1,87 @@
+require "test_helper"
+
+class AwsSecretsmanagerAdapterTest < SecretAdapterTestCase
```
djmb, line 3: Could we rename this too?

lib/kamal/secrets/adapters/aws_secretsmanager.rb
```diff
@@ -0,0 +1,25 @@
...
+
+          secret_string.each do |key, value|
+            results["#{secret_name}/#{key}"] = value
+          end
+        end
+      end
+    end
+
+    def get_from_secrets_manager(secrets, account:)
+      `aws secretsmanager batch-get-secret-value --secret-id-list #{secrets.map(&:shellescape).join(" ")} --profile #{account}`.tap do
```
djmb, line 21: Let's shell escape the account here as well 👍

djmb:
@justindell - thanks for the PR! 

Just a few minor suggestions, then we can get this merged.

djmb:
Sorry one more thing - could we also implement the check_dependencies! method that was recently added to the base adapter. It should give a nice error message if the cli is not installed.

lib/kamal/secrets/adapters/aws_secretsmanager.rb
justindell: It's funny, I went back and forth on the name but ultimately decided on one word because the AWS CLI calls it "secretsmanager" over "secrets-manager". I agree though, I like the two word version better and will get that changed.

lib/kamal/secrets/adapters/aws_secretsmanager.rb
justindell: I think there's just too many different ways to authenticate with AWS. I believe the preferred method is SSO using the `aws sso login` command, but the user could also be authenticated with access keys, environment variables, automatically (if they're on an EC2 instance with an IAM Role), or even an external process.

One option would be to check if the user is authenticated with `aws sts get-caller-identity`, and if that command returns an error, attempt to run `aws sso login`. However, this could be confusing if they normally authenticate a different way, or even worse have aws cli v1 installed which doesn't include sso login.

What do you think?

djmb: Yes let's leave it for now. We could I guess add a separate `SsoAwsSecretsManager` or something like that later.

justindell:
Should be all set! Let me know if you see anything else

chrisdebruin:
@justindell Thanks for this addition just what I needed. However I found an issue when trying this out. I have some passwords containing `(` and this goes wrong. For example:

```
results  = { test: 'password(' }
secrets = JSON.dump(results).shellescape =>  "\\{\\\"test\\\":\\\"password\\(\\\"\\}" 
JSON.parse(secrets) => unexpected token at '\{\"test\":\"password\(\"\}' (JSON::ParserError)
```

as you can see it is escaping the `(` in the password string but it's not unescaped when reading it and trying to parse the json

justindell:
> @justindell Thanks for this addition just what I needed. However I found an issue when trying this out. I have some passwords containing `(` and this goes wrong. For example:
> 
> ```
> results  = { test: 'password(' }
> secrets = JSON.dump(results).shellescape =>  "\\{\\\"test\\\":\\\"password\\(\\\"\\}" 
> JSON.parse(secrets) => unexpected token at '\{\"test\":\"password\(\"\}' (JSON::ParserError)
> ```
> 
> as you can see it is escaping the `(` in the password string but it's not unescaped when reading it and trying to parse the json

Could it be related to this? https://github.com/basecamp/kamal/pull/1009 What happens if you try the `kamal secrets print` command?

chrisdebruin:
@justindell 

here is output of `kamal secrets print`

```
./bin/kamal secrets print -d staging
SECRETS=\{\"staging/TEST\":\"password\(\",\"staging/TEST2\":\"bar\"\}
TEST=$(bundle exec kamal secrets extract TEST \{\"staging/TEST\":\"password\(\",\"staging/TEST2\":\"bar\"\})
```

I don;t see the JSON parse error anymore but as you can see the TEST does not contain the right value. It seems the `(`  in the value will break the `kamal secrets extract` command

justindell:
This looks like an issue outside of the scope of the aws secrets manager. I just created passwords with both 1password and AWS Secrets Manager with a password value of "test(", and it looks like it isn't properly being extracted (although "test" works fine on both). Would you like to create an issue?

chrisdebruin:
I'll create a PR to fix this issue

Issue #1143, opened 2024-10-22
[Feature] Proxy logging improvements


arthurwozniak:
Hi there!

I recently attempted to migrate from `kamal` version `1.9.1` to `2.2.2` and encountered a couple of issues that require some workaround configurations for `kamal` and `rsyslog` to maintain the original behavior.

### 1. Compliance with `logging.driver`

It seems that the container running `kamal-proxy` is not utilizing the specified `logging.driver` configuration (if set) and is instead writing logs to a file. This change is unexpected, as `traefik` adhered to the logging driver specified in the configuration.

Current Configuration:

```yaml
logging:
  driver: syslog
  options:
    tag: '{{ index (split .Name "-") 1 }}'
```

### 2. Passing custom docker options to proxy
To address the logging issue, I attempted to set docker options using the `boot_config` command.

```bash
$ kamal proxy boot_config set --docker_options log-driver=syslog -d staging
$ kamal proxy reboot -d staging
Note: This may cause a brief outage on each host. Are you sure? [y, N] (N) y
...
```

However, I encountered an error:

```bash
INFO [e1b4c244] Running docker run --name kamal-proxy --network kamal --detach --restart unless-stopped --volume kamal-proxy-config:/home/kamal-proxy/.config/kamal-proxy $(cat .kamal/proxy/options || echo "--publish 80:80 --publish 443:443 --log-opt max-size=10m") basecamp/kamal-proxy:v0.8.1 on A.B.C.D
Releasing the deploy lock...
ERROR (SSHKit::Command::Failed): Exception while executing on host A.B.C.D: docker exit status: 125
docker stdout: Nothing written
docker stderr: docker: Error response from daemon: unknown log opt 'max-size' for syslog log driver.
See 'docker run --help'.
```

It appears that Kamal [internally sets](https://github.com/basecamp/kamal/blob/main/lib/kamal/commands/proxy.rb#L76) the `max-size` option, which conflicts with the syslog log driver. This can be resolved by manually editing `.kamal/proxy/options` on host to remove `--log-opt max-size=10m`.

### 3. Definition of unwanted log fields
Another useful feature would be the ability to specify fields to exclude from log entries. Currently, the log entries contain many fields that may not be necessary for certain use cases, such as:

```json
{
  "time": "some_val",
  "level": "some_val",
  "msg": "some_val",
  "host": "some_val",
  "port": "some_val",
  "path": "some_val",
  "request_id": "some_val",
  "status": "some_val",
[...]

PR #1152, opened 2024-10-23
Allow log max size to not be set

djmb:
The max-size log opt is not valid for all logging drivers, such as syslog. Allow the option to be removed from the boot config with:

```
kamal proxy boot_config set --log-max-size=
or
kamal proxy boot_config set --log-max-size=""
```

Fix for https://github.com/basecamp/kamal/issues/1143

### DIFF
--- a/lib/kamal/cli/proxy.rb
+++ b/lib/kamal/cli/proxy.rb
@@ -37,7 +37,6 @@ class Kamal::Cli::Proxy < Kamal::Cli::Base
       ]
 
       on(KAMAL.proxy_hosts) do |host|
-        p boot_options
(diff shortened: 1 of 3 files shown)


lib/kamal/cli/proxy.rb
```diff
@@ -37,6 +37,7 @@ def boot_config(subcommand)
       ]
 
       on(KAMAL.proxy_hosts) do |host|
+        p boot_options
```
Sija, line 40: Debug leftover

```suggestion
```

Sija (changes requested):

PR #1181, opened 2024-10-31
add host-docker-setup hook

eduardinni:
Add new hook to run in host after docker is installed.

Motivation: Install docker plug-ins in the host, such as log-drivers.

`host-docker-setup` hook example:

```bash
#!/bin/sh

echo "Docker set up on $KAMAL_HOSTS..."

# Check if the Loki Docker Driver plug-in is already installed
if ! docker plugin ls | grep -q "loki"; then
  echo "Installing Loki Docker Driver plug-in..."
  docker plugin install grafana/loki-docker-driver:2.9.2 --alias loki --grant-all-permissions
else
  echo "Loki Docker Driver plug-in is already installed."
fi
```

This will enable the usage of custom logging driver in the config:

```yaml
# deploy.yml

logging:
  driver: loki
  options:
    loki-url: https://<user_id>:<password>@logs-us-west1.grafana.net/loki/api/v1/push
    loki-batch-size: "400"
```

### DIFF
--- a/lib/kamal/cli/server.rb
+++ b/lib/kamal/cli/server.rb
@@ -32,6 +32,7 @@ class Kamal::Cli::Server < Kamal::Cli::Base
           if execute(*KAMAL.docker.superuser?, raise_on_non_zero_exit: false)
             info "Missing Docker on #{host}. Installing…"
             execute *KAMAL.docker.install
+            run_hook "host-docker-setup"
           else
             missing << host
           end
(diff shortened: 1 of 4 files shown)


eduardinni:
Closing it because this can be achieved inside the existing `docker-setup` hook without adding extra complexity.

For anyone interested on installing a docker plugin from ``docker-setup` hook:

```bash
#!/bin/sh

echo "Docker set up on $KAMAL_HOSTS. Destination $KAMAL_DESTINATION"

# Check if the Loki Docker Driver plug-in is already installed
if ! kamal server exec "docker plugin ls" | grep -q "loki"; then
  echo "Installing Loki Docker Driver plug-in..."
  kamal server exec \
  "docker plugin install grafana/loki-docker-driver:2.9.2 --alias loki --grant-all-permissions" \
  -d $KAMAL_DESTINATION
else
  echo "Loki Docker Driver plug-in is already installed."
fi
```

PR #1186, opened 2024-11-02
feat: add Bitwarden Secrets Manager adapter

oandalib:
Bitwarden Secrets Manager has better access controls for exposing just the secrets needed for my project. It is an ideal alternative to using Bitwarden with my personal account where all of my other logins are also exposed, or creating a secondary Bitwarden account. They offer a free plan on top of your regular Bitwarden account which should be sufficient for most people -> https://bitwarden.com/products/secrets-manager/.

Due to Bitwarden Secrets Manager using a session token, I made the account field in the "fetch secrets" optional.


djmb:
Thanks for the PR - @oandalib!

I've made a few notes. I think also that we need to add support for projects (https://bitwarden.com/help/secrets-manager-cli/#project).

We could use the `--from` option to specify the project:

```
kamal secrets fetch --adapter bitwarden-sm --from <project-id> SECRET1 SECRET2 SECRET3
```

And also allow:
```
kamal secrets fetch --adapter bitwarden-sm  <project-id1>/SECRET1 <project-id1>/SECRET2 <project-id2>SECRET3
```
in case you wanted to fetch from multiple projects.

oandalib:
Thanks for the comments. That looks useful to add, and I will do so.

Do note that the CLI only supports getting secrets by an ID which is generated by Bitwarden Secrets Manager, and not via secret names.

oandalib:
Once secrets are created, they are identified by a UUID. The only time you can pass in a project id (also a UUID) is when doing a [list secrets operation](https://bitwarden.com/help/secrets-manager-cli/#secret-list) for a specific project. I think Bitwarden Secrets Manager does a good job of separating and limiting access to specific secrets through a combination of  projects and machine accounts. So I would propose this will be sufficient:

1. Fetch all secrets that the machine account has access to:
```shell
kamal secrets fetch --adapter bitwarden-sm *
``` 
2. Fetch all secrets from the project specified that the machine account has access to:
```shell
kamal secrets fetch --adapter bitwarden-sm * --from <project-uuid>
``` 
3. Fetch single/multiple items:
```shell
kamal secrets fetch --adapter bitwarden-sm <secret-uuid-1> <secret-uuid-2>
``` 

For 3, note that `<project-uuid>/<secret-uuid-1>` is not supported. As far as the CLI/API is concerned for GET operations, there is no link between projects and secrets.


I can update the documentation ticket once we get closer to merging this. Let me know what you think with the new approach and changes!

janosrusiczki:
@oandalib I'm very much looking forward to this (and I don't think I am the only one). I just wanted to give you a heads up that there's now a conflict.

oandalib:
Thanks for the heads up @janosrusiczki!

I resolved the conflict and decided to use the keyword "all" rather than "*" for getting all of the secrets.

Should be ready for you @djmb 😃

lib/kamal/secrets/adapters/bitwarden_secrets_manager.rb
```diff
@@ -0,0 +1,66 @@
...
+      [ key, value ]
+    end
+
+    def run_command(command, session: nil)
+      full_command = [ "bws", command ].join(" ")
+      `#{full_command}`.strip unless full_command.nil?
+    end
+
+    def login(account)
+      nil
```
djmb, line 55: Agh my review was stuck in pending 🙈, sorry!

Can we implement this method so it checks if you are logged in and prompts you to do so if you are not?

oandalib: All good!

Added.

oandalib:
Anything else needed here @djmb? 😃

djmb:
All good, thanks @oandalib!

PR #1189, opened 2024-11-03
Add support for Enpass - a password manager for secrets

egze:
I hope the maintainers would like to add support for one more password manager - [Enpass](https://www.enpass.io). It is a well known app, and I would be thrilled to use it together with Kamal. I also promise to fix any future issues with it.

Given a saved secret like this:
382569507-ed83f416-ecae-4c40-bf2e-d720dea5b712

Usage with Kamal would be:

```bash
# Fetch from item `FooBar ` in the vault `PATH_TO_VAULT `
kamal secrets fetch --adapter enpass --from <PATH_TO_VAULT> FooBar

kamal secrets extract DB_PASSWORD <SECRETS-FETCH-OUTPUT>
kamal secrets extract FooBar/DB_PASSWORD <SECRETS-FETCH-OUTPUT>
```

### DIFF
--- a/lib/kamal/secrets/adapters/enpass.rb
+++ b/lib/kamal/secrets/adapters/enpass.rb
@@ -1,19 +1,24 @@
 ##
-# Enpass is different from most password managers, in a way that it's offline. A path to a vault is treated as account.
+# Enpass is different from most password managers, in a way that it's offline and doesn't need an account.
 #
-# Pass it like so: `kamal secrets fetch --adapter enpass --account /Users/YOUR_USERNAME/Library/Containers/in.sinew.Enpass-Desktop/Data/Documents/Vaults/primary --from MY_PROD_SERVER`
+# Usage
+#
+# Fetch all password from FooBar item
+# `kamal secrets fetch --adapter enpass --from /Users/YOUR_USERNAME/Library/Containers/in.sinew.Enpass-Desktop/Data/Documents/Vaults/primary FooBar`
+#
+# Fetch only DB_PASSWORD from FooBar item
+# `kamal secrets fetch --adapter enpass --from /Users/YOUR_USERNAME/Library/Containers/in.sinew.Enpass-Desktop/Data/Documents/Vaults/primary FooBar/DB_PASSWORD`
 class Kamal::Secrets::Adapters::Enpass < Kamal::Secrets::Adapters::Base
-  private
-    def login(account)
-      # There is no concept of session in enpass-cli
-      true
-    end
+  def fetch(secrets, account: nil, from:)
+    check_dependencies!
+    fetch_secrets(secrets, from)
+  end
 
-    def fetch_secrets(secrets, account:, session:)
+  private
+    def fetch_secrets(secrets, vault)
       secrets_titles = fetch_secret_titles(secrets)
 
-      # Enpass outputs result as stderr, I did not find a way to stub backticks and output to stderr. Open3 did the job.
-      result = `enpass-cli -json -vault #{account.shellescape} show #{secrets.map(&:shellescape).join(" ")}`.strip
+      result = `enpass-cli -json -vault #{vault.shellescape} show #{secrets_titles.map(&:shellescape).join(" ")}`.strip
 
       parse_result_and_take_secrets(result, secrets)
     end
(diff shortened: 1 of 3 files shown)


lib/kamal/secrets/adapters/enpass.rb
```diff
@@ -0,0 +1,62 @@
...
+      raise RuntimeError, "Enpass CLI is not installed" unless cli_installed?
+    end
+
+    def cli_installed?
+      `enpass-cli version 2> /dev/null`
+      $?.success?
+    end
+
+    def fetch_secret_titles(secrets)
+      secrets.reduce(Set.new) do |acc, secret|
```
djmb, line 34: What does `acc` mean here?

lib/kamal/secrets/adapters/enpass.rb
```diff
@@ -0,0 +1,62 @@
...
+          acc << key
+        end
+      end.to_a
+    end
+
+    def parse_result_and_take_secrets(unparsed_result, secrets)
+      unparsed_result.split("\n").reduce({}) do |acc, line|
+        title = line[/title:\s{1}(\w+)/, 1]
+        label = line[/label:\s{1}(.*?)\s{2}/, 1]
+        password = line[/password:\s{1}([^"]+)/, 1]
```
djmb, line 49: This parsing seems very fragile - what about titles with spaces, or labels with two spaces? 

I think this is really just a function of Enpass's output format though - it would be nice if it offered JSON.

Maybe we should just ask for each password one by one to avoid ambiguity or misparsing of results? How fast is a single `enpasscli pass` call?

lib/kamal/secrets/adapters/enpass.rb
```diff
@@ -0,0 +1,62 @@
...
+    def login(account)
+      # There is no concept of session in enpass-cli
+      true
+    end
+
+    def fetch_secrets(secrets, account:, session:)
+      secrets_titles = fetch_secret_titles(secrets)
+
+      # Enpass outputs result as stderr, I did not find a way to stub backticks and output to stderr. Open3 did the job.
+      _stdout, stderr, status = Open3.capture3("enpass-cli -vault #{account.shellescape} show #{secrets_titles.map(&:shellescape).join(" ")}")
```
djmb, line 18: There no account here, so we probably:

1. Avoid passing an account at all
2. Use the from field to specify the vault

The base adapter just appends the from field and secrets together and passes them into `fetch_secrets`. 

That's not going to work here since the separator it uses is a `/`, so I think it should be ok to replace the `fetch` method for this adapter and have it pass the from field separately. 

We can refactor that to be neater later once all the other new adapter PRs have landed to avoid messing with them.

djmb (changes requested):
Thanks for the PR @egze!

I've added a few suggestions, let me know if they make sense!

lib/kamal/secrets/adapters/enpass.rb
egze: Accumulator. Common naming in Elixir. What would you recommend?

lib/kamal/secrets/adapters/enpass.rb
egze: Right now they don't offer JSON, but I have brought it up. It it gets added - then I will change the current implementation.

As to getting the password one by one - it won't be a good experience, because there is no concept of a session. So you would have to enter the master password for each item.

lib/kamal/secrets/adapters/enpass.rb
egze: Yes, if I could override the `fetch`, that would be cleaner.

egze: @djmb The `-account` option is marked as required in `Kamal::Cli::Secrets` (`option :account, type: :string, required: true`). If I want to not pass it at all, I need to make it optional. Is this a good idea to do it, just because of the way Enpass works? For other password managers it does make sense to have it required.

I think it's more practical to treat the Vault path as account for Enpass. What do you think?

Another option is to pass some dummy value as account and just not use it, but I'm also not sure about it.

lib/kamal/secrets/adapters/enpass.rb
egze: I changed it to `secrets_with_passwords`

lib/kamal/secrets/adapters/enpass.rb
egze: I went ahead and make `--account` optional and passed vault in `--from` 327d716

PR #1191, opened 2024-11-04
Add ability to specify `--shm-size` in `docker run` command

frayzil:


djmb:
You can set `shm-size` via:

```
servers:
  web:
    options:
      shm_size: ...
```

frayzil:
Is this a recent update?

Issue #1184, opened 2024-11-01
[Kamal 2]: Container Conflicts for Staging and Production Application Running on Same Host


djmb:
### Discussed in https://github.com/basecamp/kamal/discussions/1110

<div type='discussions-op-text'>

<sup>Originally posted by **mike-weiner** October 14, 2024</sup>
I've got a basic web application running on a DigitalOcean Droplet. I have both the production and staging instances of the application on the same host. I realize that's atypical. It's a non-critical, side project that I don't need redundancy for nor do I need to add complexity for 2 different hosts. 

I'm hitting a weird edge case that I'm looking for some advice on. 

When I've got a new build ready to go out to staging, I do a `kamal deploy -d staging` and it builds the container with the current commit (let's say that is `123`), builds the image, and exposes port `abcd` for my web application. That deploys perfectly. 

24 hours later I'll do a `kamal deploy` to push the changes out to production. Kamal re-builds the container with the same commit (`123`) but it exposes a different port, let's say `efgh` for production. That process then overwrites that image in my registry with a build that now exposes port `efgh` instead of `abcd`.

The production instance deploys fine, but if I ever need to restart the staging instance, it will go and re-pull the `123` container image which is now built to expose the incorrect port. 

Is there a better way to achieve this kind of setup to avoid this? Or is this a limitation that I'm going to need to live with?

Thank you!</div>

djmb:
@mike-weiner diagnosed the issue here in the discussion thread - https://github.com/basecamp/kamal/discussions/1110#discussioncomment-11124817

mike-weiner:
@djmb - I'm thinking a solution here might be as changing [this line](https://github.com/basecamp/kamal/blob/main/lib/kamal/commands/app.rb#L108) to: 

```
filters << "label=destination=#{config.destination}"
```

PR #1193, opened 2024-11-04
Filter correctly for empty destinations

djmb:
An empty destination should only filter container with empty destination, not pick up all containers.

Fixes: https://github.com/basecamp/kamal/issues/1184


mike-weiner (commented):
/lgtm

Thank you!

PR #1202, opened 2024-11-07
feat: support custom proxy image repository

songjiz:
- Allow users to specify custom proxy image repository instead of using default basecamp/kamal-proxy
- Help users in regions where Docker Hub access is restricted (e.g. China) to pull proxy image from cloud vendor repositories
- Move proxy related configurations from configuration to proxy


lib/kamal/configuration/proxy.rb
```diff
@@ -1,8 +1,13 @@
 class Kamal::Configuration::Proxy
   include Kamal::Configuration::Validation
 
+  HTTP_PORT = 80
+  HTTPS_PORT = 443
+  LOG_MAX_SIZE = "10m"
+  MINIMUM_VERSION = "v0.8.2"
```
AxelTheGerman, line 7: Nice clean up!

I was also looking at proxy configuration options, hoping too override the version with a newer one - see #1252

jethrodaniel:
I also ran into DockerHub's rate-limiting here.

My app's image is hosted on Gitlab's registry, which means that when the server tries to `docker pull basecamp/kamal-proxy`, it's not authenticated with DockerHub, since the `docker login` was for a different registry.

I've gotten around the issue by monkey-patching over this in `bin/kamal`:

```diff
diff --git a/bin/kamal b/bin/kamal
index cbe59b9..44f6b98 100755
--- a/bin/kamal
+++ b/bin/kamal
@@ -24,4 +24,14 @@ end
 require "rubygems"
 require "bundler/setup"
 
+#-- PATCH
+require "kamal"
+class Kamal::Configuration
+  def proxy_image
+    # "basecamp/kamal-proxy:#{PROXY_MINIMUM_VERSION}"
+    "registry.gitlab.com/myuser/myapp/kamal-proxy:v0.8.2"
+  end
+end
+#--
+
 load Gem.bin_path("kamal", "kamal")
```

(and pushing a copy of the `kamal-proxy` image to my own registry, of course).

Official support for this would be great.

djmb:
Implemented in https://github.com/basecamp/kamal/pull/1499

jethrodaniel:
Thanks, works great!

Issue #1108, opened 2024-10-12
Docker -p error


indigotechtutorials:
I am encountering a strange issue with my attempts to deploy to AWS and Azure the main reason I believe it is not working is because they both require a .pem credential file
<img width="453" alt="Screenshot 2024-10-12 at 5 56 33 PM" src="https://github.com/user-attachments/assets/5e3e56e1-749f-40b5-9bba-377f24399932">
<img width="1183" alt="Screenshot 2024-10-12 at 6 00 12 PM" src="https://github.com/user-attachments/assets/76cdc36f-7f74-42a4-95b9-74a8c14df8c4">

this is the code from my deploy file the username and path is correct
`ssh:
  user: indigo
  keys: [ "/Users/indigo/documents/Mac-m2.pem" ]`

indigotechtutorials:
I realized that this was due to my misunderstanding of how the new secrets file works. I thought that setting the values in .env file would automatically set them in environment for kamal but turns out I needed to set them manually on my terminal shell instead to get it to work with Kamal. I'm learning a lot and im sure this tool will be awesome once I figure it out

Issue #1001, opened 2024-09-29
SSH: Allow User and Port to Be Secrets


mike-weiner:
[Per the `ssh` docs](https://kamal-deploy.org/docs/configuration/ssh/), the `user` and `port` fields currently have to be specified in plain text, something like:

```yaml
ssh:
  user: root
  port: "22"
```

I attempted to use secrets for these values: 

```yaml
ssh:
  user:
    - KAMAL_SSH_USER
  port:
    - KAMAL_SSH_PORT
```

and received: 

```
ERROR (Kamal::ConfigurationError): ssh/user: should be a string
```

I would prefer not to specify the user or port that I'm using for SSH to do these deployments in plain-text as I'd like to commit my config to Github. Is there a technical limitation requiring these to be strings?

mike-weiner:
This is already possible via: 

```
ssh:
  user: KAMAL_SSH_USERNAME
  port: KAMAL_SSH_PORT
```

mike-weiner:
Doesn't look like the above config is a fix either. Still seeing: 

```
ERROR (Socket::ResolutionError): Exception while executing on host XXX.XXX.XXX.XXX: getaddrinfo: nodename nor servname provided, or not known
```

mike-weiner:
The goal I'm looking to achieve here is for `.kamal/secrets-common` to look something like:

```
KAMAL_REGISTRY_USERNAME=$KAMAL_REGISTRY_USERNAME
KAMAL_REGISTRY_PASSWORD=$KAMAL_REGISTRY_PASSWORD
KAMAL_SSH_USERNAME=$KAMAL_SSH_USERNAME
KAMAL_SSH_PORT=$KAMAL_SSH_PORT
```

There seems to be a slightly difference in behavior that I'd like to understand. Per https://kamal-deploy.org/docs/configuration/docker-registry/:

> A reference to secret (in this case DOCKER_REGISTRY_TOKEN) will look up the secret in the local environment.

This makes it sound as if `DOCKER_REGISTRY_TOKEN` is special and can only be placed as a secret in `.kamal/secrets`. Looking at this source for how this is fetched, I don't think I could use something like 1Password for this secret: https://github.com/basecamp/kamal/blob/main/lib/kamal/configuration/registry.rb#L25-L31

[...]

Issue #970, opened 2024-09-25
Kamal setup fails (version 2.0.0.rc3 from docker)


a3kov:
Looks like there's a problem with secrets - the registry password is empty. I have proper .env file in the project directory containing `KAMAL_REGISTRY_PASSWORD=mypassword`
```
$ kamal setup
  INFO [4a37da09] Running /usr/bin/env mkdir -p .kamal on 1.2.3.4
  INFO [4a37da09] Finished in 5.948 seconds with exit status 0 (successful).
Acquiring the deploy lock...
Ensure Docker is installed...
  INFO [312bfcba] Running docker -v on 1.2.3.4
  INFO [312bfcba] Finished in 1.201 seconds with exit status 0 (successful).
  INFO [05559197] Running docker login registry.gitlab.com -u [REDACTED] -p [REDACTED] on 1.2.3.4
Releasing the deploy lock...
  Finished all in 9.5 seconds
  ERROR (SSHKit::Command::Failed): Exception while executing on host 1.2.3.4: docker exit status: 125
docker stdout: Nothing written
docker stderr: flag needs an argument: 'p' in -p
See 'docker login --help'.
```

djmb:
@a3kov - the secrets have moved in Kamal 2.0 - you should put them in `.kamal/secrets` instead now.

a3kov:
Yes, I have `KAMAL_REGISTRY_PASSWORD=$KAMAL_REGISTRY_PASSWORD`
I thought Kamal will read the `.env` file to substitute `$KAMAL_REGISTRY_PASSWORD`. Do I need to provide environment variables myself now ? That was totally unexpected

a3kov:
Ok, now after reading https://github.com/basecamp/kamal/pull/924 it's a bit more clear..

a3kov:
Hmm, after making KAMAL_REGISTRY_PASSWORD variable available in the current environment, it still doesn't work (same symptoms). Any ideas ?

a3kov:
To anyone having the same problem: if you run Kamal from docker, you have to pass the environment inside the container, otherwise it won't be available to the Kamal process.

a3kov:
I actually had multiple issues with the docker version of Kamal so I switched to the gem one. It seems nobody is testing if the docker version even works

mgax:
> To anyone having the same problem: if you run Kamal from docker, you have to pass the environment inside the container, otherwise it won't be available to the Kamal process.

[...]

PR #1218, opened 2024-11-15
Warn user on blank secrets

andrewjanssen:
# Motivation

It's easy to misconfigure Kamal secrets, secret helpers, and environment variables. This is especially true for i) users who are new to Kamal, ii) users who [may not have noticed](https://github.com/basecamp/kamal/issues/970#issuecomment-2376520786) the changes to secrets between Kamal 1 and 2, and iii) users like myself who are using Kamal inside Docker and/or in remote environments.

To see how often people have challenges related to secrets, consider the number of issues opened [about them](https://github.com/basecamp/kamal/issues?q=secret+in%3Atitle+is%3Aissue). For example, #970, [#1001](https://github.com/basecamp/kamal/issues/1001#issuecomment-2381392373), #1108, and [discussion 977](https://github.com/basecamp/kamal/discussions/977#discussion-7238545). Secrets are simple yet easy to get wrong!

We should make it easier for users to get back on track when secrets are misconfigured. This PR makes it easier.

# Pain point

Currently, if secret values are blank, there is no warning.

A common special case is when `KAMAL_REGISTRY_PASSWORD` is not set, as easily happens if the user is trying to get Kamal to work in, say, a GitHub Action. This causes `kamal setup` and `kamal deploy` to emit the following output. (This example is from #970.)

```
$ kamal setup
  INFO [4a37da09] Running /usr/bin/env mkdir -p .kamal on 1.2.3.4
  INFO [4a37da09] Finished in 5.948 seconds with exit status 0 (successful).
Acquiring the deploy lock...
Ensure Docker is installed...
  INFO [312bfcba] Running docker -v on 1.2.3.4
  INFO [312bfcba] Finished in 1.201 seconds with exit status 0 (successful).
  INFO [05559197] Running docker login registry.gitlab.com -u [REDACTED] -p [REDACTED] on 1.2.3.4
Releasing the deploy lock...
  Finished all in 9.5 seconds
  ERROR (SSHKit::Command::Failed): Exception while executing on host 1.2.3.4: docker exit status: 125
docker stderr: flag needs an argument: 'p' in -p
See 'docker login --help'.
```

It's hard to understand the root cause there.

# Proposed change

* `Kamal::Secrets` should show a warning whenever it loads an empty value.
* If there is no registry password, `Kamal::Configuration::Registry` should raise a helpful exception instead of showing that Docker argument error.

# New behaviour introduced by this PR

1. Any time `Kamal::Secrets::[]` fetches a blank value, it displays a warning and tips, as shown in the next cases where a blank value is fetched by `Kamal::Configuration::Registry`.

2. If a blank registry password comes from an unset environment variable, the user sees the warning and tips, and a `Kamal::ConfigurationError` is raised. (Similar if the env var is set to the empty string.)
```
bundle exec ../kamal/bin/kamal deploy
Log into image registry...
Warning: Kamal secret key 'KAMAL_REGISTRY_PASSWORD' is blank.
Tip: see .kamal/secrets:8: KAMAL_REGISTRY_PASSWORD=$KAMAL_REGISTRY_PASSWORD
Tip: the environment variable KAMAL_REGISTRY_PASSWORD is not set. Did you forget to set it?
  Finished all in 0.1 seconds
  ERROR (Kamal::ConfigurationError): The required secret 'registry.password' does not have a value.
```

3. If a blank value comes from a secret helper, these are the tips:
```
Tip: see .kamal/secrets:8: KAMAL_REGISTRY_PASSWORD=$(echo '')
Tip: the shell command `echo ''` returned an empty value.
```

# Other comments

It's true that `Kamal::Configuration::Registry` [validates](https://github.com/basecamp/kamal/blob/9cf8da64c402897701a54ddd44375df166ab5233/lib/kamal/configuration/validator/registry.rb#L10) the presence of the password key, but it does not detect when that key is replaced with an empty value.

I'm happy to make any change to this PR, just let me know. Thanks for contributing to Kamal!

### DIFF
--- a/lib/kamal/configuration/registry.rb
+++ b/lib/kamal/configuration/registry.rb
@@ -24,7 +24,15 @@ class Kamal::Configuration::Registry
   private
     def lookup(key)
       if registry_config[key].is_a?(Array)
-        secrets[registry_config[key].first]
+        secret = secrets[registry_config[key].first]
+        # Although the key is present, its value may be empty due to environment
+        # variable substitution. I.e. if it refers to an empty or unset env var.
+        unless secret.present?
+          raise Kamal::ConfigurationError, "The required secret " \
+            "'registry.#{key}' does not have a value."
+        end
+
+        secret
       else
         registry_config[key]
       end
--- a/lib/kamal/secrets.rb
+++ b/lib/kamal/secrets.rb
@@ -10,9 +10,13 @@ class Kamal::Secrets
 
   def [](key)
     # Fetching secrets may ask the user for input, so ensure only one thread does that
-    @mutex.synchronize do
+    value = @mutex.synchronize do
       secrets.fetch(key)
     end
+
+    blank_key_warning(key) unless value.present?
+
+    value
   rescue KeyError
     if secrets_files.present?
       raise Kamal::ConfigurationError, "Secret '#{key}' not found in #{secrets_files.join(", ")}"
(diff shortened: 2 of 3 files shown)


loqimean:
Hey, thought about it and found this already opened issue, would be fine to this feature, because ofter I have error about failed docker login because secret/ENV variable wasn't set and it says login incorrect, but not that I have empty registry password @djmb @dhh 

CleanShot 2026-07-07 at 1  14 39@2x

PR #1227, opened 2024-11-19
kamal app exec may fail if not logged in

### LINKED ISSUE #1163: kamal app exec doesn't log in to docker
Kamal v2.2.2

Running `kamal app exec` to run migrations fails intermittently because kamal doesn't execute `docker login` on the remote host. In out case, the previously used ECR token expires at some point so depending on when the previous pipeline run was triggered, the next `app exec` may fail.

```
$ kamal app exec --version v1.2.3 --roles web --primary -d qa rails db:migrate
Launching command with version v1.2.3 from new container...
Running the pre-connect hook...
  INFO [4b63d86e] Running /usr/bin/env .kamal/hooks/pre-connect on localhost
  INFO [4b63d86e] Finished in 2.857 seconds with exit status 0 (successful).
  INFO [b2bca27f] Running docker run --rm --network kamal --env SERVICE="web" ... x.dkr.ecr.eu-central-1.amazonaws.com/image:v1.2.3
  ERROR (SSHKit::Command::Failed): Exception while executing on host x.x.x.x: docker exit status: 125
docker stderr: Unable to find image 'x.dkr.ecr.eu-central-1.amazonaws.com/image:v1.2.3' locally
docker: Error response from daemon: pull access denied for x.dkr.ecr.eu-central-1.amazonaws.com/image, repository does not exist or may require 'docker login': denied: Your authorization token has expired. Reauthenticate and try again.
See 'docker run --help'.
```

Compare it with `kamal deploy`:

```
$ kamal deploy -P --version v1.2.3 -d qa
Log into image registry...
Running the pre-connect hook...
  INFO [cd18ce5e] Running /usr/bin/env .kamal/hooks/pre-connect on localhost
  INFO [cd18ce5e] Finished in 3.591 seconds with exit status 0 (successful).
  INFO [4631cfff] Running docker login x.dkr.ecr.eu-central-1.amazonaws.com -u [REDACTED] -p [REDACTED] on x
  INFO [4631cfff] Finished in 9.694 seconds with exit status 0 (successful).
...
```

andrewjanssen:
# Current state

The central Kamal command, `kamal deploy`, automatically logs you in because it requires access to the container registry.

Unlike the deploy command, `kamal app exec` does not log you in. Yet it sometimes requires registry access too. In these cases, when the host is not auth'd, it fails.

Why would `kamal app exec` need auth? When you tell it to use an image that's not currently on the host, as in #1163.

Note that containers don't stay authorized very long. Container registry session tokens expire pretty quickly:

| Registry | Session expiry |
| ------| ----| 
| Amazon ECR | [12 hours](https://docs.aws.amazon.com/AmazonECR/latest/userguide/registry_auth.html#:~:text=valid%20for%2012%20hours) |
| Azure Container Registry | [3 hours](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication#:~:text=3%20hours) |
| Docker Hub | [24 hours](https://docs.docker.com/security/faqs/general#:~:text=24%20hours) |
| GitHub Container Registry | varies |
| Harbor | [30 minutes?](https://github.com/goharbor/harbor/blob/main/src/common/utils/test/config.go#L55) |

# Proposed state

It would be nice for all Kamal commands to have the same “it just works” auth behavior. I think this is the simplicity that attracts people to Kamal. So it would be nice if `kamal exec` took care of its own auth, like `kamal deploy` does.

# Changes introduced by this PR

This PR makes `kamal app exec` log in before executing `docker run`.

# Other comments

If this PR is approved I'd be happy to make a follow-up PR for `kamal accessory exec`.

Resolves #1163.

### DIFF
--- a/lib/kamal/cli/app.rb
+++ b/lib/kamal/cli/app.rb
@@ -97,6 +97,10 @@ class Kamal::Cli::App < Kamal::Cli::Base
   def exec(*cmd)
     cmd = Kamal::Utils.join_commands(cmd)
     env = options[:env]
+    unless options[:reuse]
+      say "Log into image registry...", :magenta
+      invoke "kamal:cli:registry:login", [], {skip_local: true}
+    end
     case
     when options[:interactive] && options[:reuse]
       say "Get current version of running container...", :magenta unless options[:version]
--- a/test/cli/app_test.rb
+++ b/test/cli/app_test.rb
@@ -299,6 +299,17 @@ class CliAppTest < CliTestCase
     end
   end
 
+  test "exec logs into the container registry remotely for new containers" do
+    run_command("exec", "ruby -v").tap do |output|
+      assert_no_match /docker login -u \[REDACTED\] -p \[REDACTED\] as .*@localhost/, output
+      assert_match /docker login -u \[REDACTED\] -p \[REDACTED\] on 1.1.1.\d/, output
+    end
+    # Login is never required if reuse
+    run_command("exec", "-i", "--reuse", "ruby -v").tap do |output|
+      assert_no_match /docker login -u \[REDACTED\] -p \[REDACTED\] on 1.1.1.\d/, output
+    end
+  end
+
   test "containers" do
     run_command("containers").tap do |output|
       assert_match "docker container ls --all --filter label=service=app", output


djmb:
@andrewjanssen, @cromega - like in https://github.com/basecamp/kamal/pull/1223, we don't want to require secrets to be able to run `kamal app exec`.

Also I would expect the image to already be on the host - what's the situation where that's not the case?

cromega:
@djmb in our case, running a migration in a new build before we get to the deploy part. A `secrets push` command could potentially help work around this problem.

The real solution, of course, would be built-in support for migrations in a way that doesn't require working around base assumptions in the tool.

djmb:
@cromega - where are you running the migration? In a `pre-deploy` hook? In that case the secrets might be out of date, but the image should be there shouldn't it?

cromega:
@djmb our workflow is the following:

Assume v1 of app is running

* CI pipeline builds image v2and pushes it to ECR, we don't use Kamal to build docker images
* CI pipeline runs `kamal app exec -r web --primary -d qa -V v2 rails db:migrate` - the v2 image needs to be pulled and it fails if the ECR token has expired
* CI pieline runs `kamal deploy -d qa -P -V v2`

andrewjanssen:
> @andrewjanssen, @cromega - like in https://github.com/basecamp/kamal/pull/1223, we don't want to require secrets to be able to run kamal app exec. Logging into the registry would need them

That's a good point, I hadn't considered that when writing this PR.

djmb:
@cromega - I think you should be able to ensure the pipeline works by adding `kamal registry login` just before your `db:migrate` command.

djmb:
Implemented in https://github.com/basecamp/kamal/pull/1511

Issue #1053, opened 2024-10-04
[Kamal 2] Error: Permission denied (publickey) only for interactive commands


augustosamame:
I managed (after a few hiccups) to successfully deploy my Rails app to an Ubuntu 24.04 server. The user Im using is ubuntu, and it works fine after I added it to the docker group. I can deploy and I can even run kamal commands remotely like:

```
kamal app exec 'ruby -v'
Get most recent version available as an image...
Launching command with version latest from new container...
  INFO [d39a7214] Running docker run --rm --network kamal --env SOLID_QUEUE_IN_PUMA="true" --env JOB_CONCURRENCY="2" --env DB_HOST="mydbhost" --env-file .kamal/apps/scraper_admin/env/roles/web.env --volume scraper_admin_storage:/rails/storage augustosamame/scraper_admin:latest ruby -v on 18.211.91.90
  INFO [d39a7214] Finished in 0.925 seconds with exit status 0 (successful).
App Host: 18.211.91.90
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-linux]
```

however, running any of the interactive commands: kamal console, kamal logs, fails with Permission denied (publickey). The SSH .pem file Im using has to be fine, otherwise it would not work with any commands.

```
kamal app exec --interactive --reuse "bin/rails console"
Get current version of running container...
  INFO [076e4960] Running /usr/bin/env sh -c 'docker ps --latest --format '\''{{.Names}}'\'' --filter label=service=scraper_admin --filter label=role=web --filter status=running --filter status=restarting --filter ancestor=$(docker image ls --filter reference=augustosamame/scraper_admin:latest --format '\''{{.ID}}'\'') ; docker ps --latest --format '\''{{.Names}}'\'' --filter label=service=scraper_admin --filter label=role=web --filter status=running --filter status=restarting' | head -1 | while read line; do echo ${line#scraper_admin-web-}; done on 18.211.91.90
  INFO [076e4960] Finished in 1.591 seconds with exit status 0 (successful).
Launching interactive command with version b7c1aec9139c8084bf9ea67f32fc12283303ad23 via SSH from existing container on 18.211.91.90...
ubuntu@18.211.91.90: Permission denied (publickey).
```

These are the relevant parts of my deploy.yml

```
aliases:
  console: app exec --interactive --reuse "bin/rails console"
  shell: app exec --interactive --reuse "bash"
  logs: app logs -f
[...]

PR #1087, opened 2024-10-09
[FIX] [Kamal 2] Pemission denied (publickey) only for interactive commands

augustosamame:
This should fix #1053

PR #959, opened 2024-09-20
[FIX] - Make kamal use ssh keys from config when executing commands

federicoaldunate:
When using the SSH keys options, commands are being executed without applying these options.

We should consider these commands to ensure they can run properly.

Example:
The command 
```
kamal app exec -i 'bin/rails console'
``` 
was not using the `-i path_to_key.pem` option for execution.

PR #1229, opened 2024-11-19
[FIX] - Make kamal use ssh keys from config when performing commands

matjack1:
Hey!

I've taken this PR: https://github.com/basecamp/kamal/pull/959 that worked fine and fixed the conflicts trying to maintain the approach of the previous code change for `ssh_proxy_args`.

I guess this PR should also fix https://github.com/basecamp/kamal/issues/1053 and so we should probably be able to close this too: https://github.com/basecamp/kamal/pull/1087

### DIFF
--- a/lib/kamal/commands/base.rb
+++ b/lib/kamal/commands/base.rb
@@ -96,12 +96,13 @@ module Kamal::Commands
       end
 
       def ssh_keys_args
-        args = ""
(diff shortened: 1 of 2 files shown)


matjack1:
Oh sorry, I didn't know about that, not sure if the fix is the idiomatic way, let me know if there's a better approach :)

Issue #1252, opened 2024-11-25
Unable to run newer kamal-proxy release


AxelTheGerman:
It looks like the "minimum" proxy version is indeed hard coded to a specific one.

https://github.com/basecamp/kamal/blob/74960499c053c0e0790f039b6f139ed46b3250cb/lib/kamal/configuration.rb#L265

https://github.com/basecamp/kamal/blob/74960499c053c0e0790f039b6f139ed46b3250cb/lib/kamal/configuration.rb#L17

Would it be possible to somehow run a newer proxy version or is it by design that the gem and proxy get upgraded together - meaning each proxy bump requires a gem bump as well?

Reason I'm asking is that there might be features in the proxy which are maybe not required for all kamal users but something I'd like to deploy to my apps already.

AxelTheGerman:
E.g. latest release `0.8.3` brings improvements and fixes minor issues but maybe shouldn't force all kamal users to reboot the proxy - https://github.com/basecamp/kamal-proxy/releases/tag/v0.8.3

morgoth:
@AxelTheGerman since https://github.com/basecamp/kamal/pull/1253 was merged, can you close the issue?

AxelTheGerman:
Yes, happy to close it.

Though when proxy reaches 1.0 we might want some versioning that allows running all minor and patch versions of a major version (maybe enforcing a minimum version).

Happy to revisit when the time comes though and not add that complexity yet.

PR #1253, opened 2024-11-26
Bump proxy minimum version to 0.8.3

AxelTheGerman:
https://github.com/basecamp/kamal-proxy/releases/tag/v0.8.3

> - Populate X-Request-Start header
> - Log client disconnections with status 499
> - Use an identifiable `User-Agent` header on healthcheck requests
> - When target fails to become healthy, mention timeout in error message

Note: I'm also happy to implement a mechanism that doesn't require bumping the minimum version for every patch release. See #1252

### DIFF
--- a/lib/kamal/configuration.rb
+++ b/lib/kamal/configuration.rb
@@ -14,7 +14,7 @@ class Kamal::Configuration
 
   include Validation
 
-  PROXY_MINIMUM_VERSION = "v0.8.2"
+  PROXY_MINIMUM_VERSION = "v0.8.3"
   PROXY_HTTP_PORT = 80
   PROXY_HTTPS_PORT = 443
   PROXY_LOG_MAX_SIZE = "10m"


AxelTheGerman:
@djmb do you think this could get merged or would you prefer having a solution that does not require bumping this version constantly and saving folks who don't care about it a proxy reboot?

djmb:
I've merged that, but looks like we need to bump it again for 0.8.4 😅

Regarding allowing users to choose the version, we should certainly allow patch version updates. 

Kamal proxy hasn't reached 1.0 yet so we may want to make some breaking changes before then. But we don't have a backward or forward compatibility policy yet so it is safest to force the versions to be in sync right now.

AxelTheGerman:
> Kamal proxy hasn't reached 1.0 yet so we may want to make some breaking changes before then. But we don't have a backward or forward compatibility policy yet so it is safest to force the versions to be in sync right now.

I assume at that point it would make sense to have a ruby gem style syntax for the proxy, e.g. `~> 1.2` or `'>= 2.2.1', '< 3.0'`

PR #1265, opened 2024-11-30
Add proxy boot_config --publish-ip argument

phoozle:
Implements https://github.com/basecamp/kamal-proxy/issues/74

Allows for `kamal proxy boot_config set --publish-ip=x.x.x.x`

Context:
By default Docker will bind to all available interfaces.

### DIFF
--- a/lib/kamal/cli/proxy.rb
+++ b/lib/kamal/cli/proxy.rb
@@ -23,6 +23,7 @@ class Kamal::Cli::Proxy < Kamal::Cli::Base
 
   desc "boot_config <set|get|reset>", "Manage kamal-proxy boot configuration"
   option :publish, type: :boolean, default: true, desc: "Publish the proxy ports on the host"
+  option :publish_ip, type: :string, desc: "IP address to bind HTTP/HTTPS traffic to. Defaults to all interfaces"
   option :http_port, type: :numeric, default: Kamal::Configuration::PROXY_HTTP_PORT, desc: "HTTP port to publish on the host"
   option :https_port, type: :numeric, default: Kamal::Configuration::PROXY_HTTPS_PORT, desc: "HTTPS port to publish on the host"
   option :log_max_size, type: :string, default: Kamal::Configuration::PROXY_LOG_MAX_SIZE, desc: "Max size of proxy logs"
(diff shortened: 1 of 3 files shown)


julianrubisch:
This looks great 👏🏻 will try it out asap, thanks!

lib/kamal/cli/proxy.rb
```diff
@@ -23,6 +23,7 @@ def boot
 
   desc "boot_config <set|get|reset>", "Manage kamal-proxy boot configuration"
   option :publish, type: :boolean, default: true, desc: "Publish the proxy ports on the host"
+  option :publish_ip, type: :string, desc: "IP address to bind HTTP/HTTPS traffic to. Defaults to all interfaces"
```
djmb, line 26: Can we call this `publish_host_ip`?

phoozle: Absolutely, I think that is clearer - I'll fix that up.

lib/kamal/configuration.rb
```diff
@@ -249,8 +249,9 @@ def env_tag(name)
     env_tags.detect { |t| t.name == name.to_s }
   end
 
-  def proxy_publish_args(http_port, https_port)
-    argumentize "--publish", [ "#{http_port}:#{PROXY_HTTP_PORT}", "#{https_port}:#{PROXY_HTTPS_PORT}" ]
+  def proxy_publish_args(http_port, https_port, bind_ip = nil)
+    bind_ip = bind_ip ? "#{bind_ip}:" : ""
```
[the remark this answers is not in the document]
phoozle: Yeah much better! I'll implement.

phoozle:
@djmb how would you feel about me allowing support for multiple IP addresses to be specified? Would use Thor's `:repeatable` option

I was just thinking about the folks that may want to bind to a specific interface that supports both IPv4 and IPv6.

Example:
````
$ kamal proxy boot_config set --publish-host-ip 127.0.0.1 --publish-host-ip '[::1]'
Host my_host_ip: --publish 127.0.0.1:80:80 --publish 127.0.0.1:443:443 --publish [::1]:80:80 --publish [::1]:443:443 --log-opt max-size=10m
```

phoozle:
@djmb hey mate, I have just added support for multiple IP addresses. There was a bit of a gotcha with using option `default: [ nil ]` when also using `repeatable: true`. Having the default array with nil is needed for the map iteration to run for wildcard interface binding (no IPs specified by user).

`option :publish_host_ip, type: :string, repeatable: true, default: [ nil ]`
Would produce the following when --publish-host-ip was provided:
`(bind_ips => [ nil, "127.0.0.1" ]`

So I went ahead and implemented the logic inside `#proxy_publish_args` with `(bind_ips || [ nil ])`

I'm of course open to feedback / alternative solutions on this.

In addition I added a simple IPv4/v6 validator for the argument value (not sure if we want to support this or not).

PR #1269, opened 2024-12-02
run_over_ssh does not respect the ssh keys in deploy.yml

chrisdebruin:
When `deploy.yml` is configured with 
```
ssh:
  keys: [~/.ssh/somekey]
```

Deployment is working but running a command like `app exec --interactive --reuse "bin/rails console"` failed to connect because it did not send the ssh key(s) configured in `deploy.yml`


djmb:
Thanks for the PR @chrisdebruin! However - I've just merged #1229 to resolve this so I'll close this one.

PR #1271, opened 2024-12-02
Don't use `shellescape` with secrets extract

### LINKED ISSUE #1007: `kamal secrets extract` JSON parse error on shell escaped secrets

dgmora:
Fixes #1007 

First time Kamal user, apologies if I'm missing the point / use case of the existing code. At least I hope it opens up the conversation 🙃. I've hit this issue some others reported too when following the initial instructions #1007

I'm trying to get credentials out of 1password:
```bash
SECRETS=$(kamal secrets fetch --verbose --adapter 1password  --account "$ACCOUNT" --from "Dev/Test-Credential" credential credentialtwo)
echo $SECRETS
\{\"Dev/Test-Credential/credential\":\"test-credential-very-secure\",\"Dev/Test-Credential/credentialtwo\":\"test-credential-very-secure-two\"\}
```
What is outputted is escaped. When running the following, there's an error:
```bash
kamal secrets extract credential $SECRETS
ERROR (JSON::ParserError): unexpected token at '\{\"Dev/Test-Credential/credenti'
```

 Setting a breakpoint here:
https://github.com/basecamp/kamal/blob/fbc451588814b04d6c87476f9bdc6f0c9a2bbc49/lib/kamal/cli/secrets.rb#L21-L23

I can see that secrets are double escaped:
```
"\\{\\\"Dev/Test-Credential/credential\\\":\\\"test-credential-very-secure\\\",\\\"Dev/Test-Credential/credentialtwo\\\":\\\"test-credential-very-secure-two\\\"\\}"
```

I think at least for this case, `shellscape` should not be necessary for `fetch`. When using `extract`, the json will be interpreted correctly with `JSON.parse` already. Looking at the tests, there are calls `shellunescape` that are not performed by anything when running `kamal secrets extract`.

Also corrected `hidden` to `hide`. `inline` was being displayed in the help and I was getting confused.

### DIFF
--- a/lib/kamal/cli/secrets.rb
+++ b/lib/kamal/cli/secrets.rb
@@ -3,7 +3,7 @@ class Kamal::Cli::Secrets < Kamal::Cli::Base
   option :adapter, type: :string, aliases: "-a", required: true, desc: "Which vault adapter to use"
   option :account, type: :string, required: false, desc: "The account identifier or username"
   option :from, type: :string, required: false, desc: "A vault or folder to fetch the secrets from"
-  option :inline, type: :boolean, required: false, hidden: true
+  option :inline, type: :boolean, required: false, hide: true
   def fetch(*secrets)
     adapter = initialize_adapter(options[:adapter])
 
@@ -13,11 +13,11 @@ class Kamal::Cli::Secrets < Kamal::Cli::Base
 
     results = adapter.fetch(secrets, **options.slice(:account, :from).symbolize_keys)
 
-    return_or_puts JSON.dump(results).shellescape, inline: options[:inline]
+    return_or_puts JSON.dump(results), inline: options[:inline]
   end
 
   desc "extract", "Extract a single secret from the results of a fetch call"
-  option :inline, type: :boolean, required: false, hidden: true
+  option :inline, type: :boolean, required: false, hide: true
   def extract(name, secrets)
     parsed_secrets = JSON.parse(secrets)
     value = parsed_secrets[name] || parsed_secrets.find { |k, v| k.end_with?("/#{name}") }&.last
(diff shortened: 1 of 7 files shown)


davidgm0:
Nevermind. This was a problem when testing out `kamal extract` in my machine, but when executing `kamal deploy` it's actually working fine there.

dgmora:
Nevermind. This was a problem when testing out `kamal extract` in my machine, but when executing `kamal deploy` the existing code works fine.

Issue #432, opened 2023-08-26
Docker won't install on MacOS


aravindreddykrishna:
kamal setup
Acquiring the deploy lock...
  INFO [9a8f3599] Running docker -v on 192.*******
  INFO [9a8f3599] Finished in 0.021 seconds with exit status 127 (failed).
  INFO [84e85df9] Running [ "${EUID:-$(id -u)}" -eq 0 ] on 192********
  INFO [84e85df9] Finished in 0.024 seconds with exit status 0 (successful).
  INFO Missing Docker on 192.*****. Installing…
  INFO [eb976864] Running /usr/bin/env curl -fsSL https://get.docker.com | sh on 192.1******
Releasing the deploy lock...
  Finished all in 0.6 seconds
  ERROR (SSHKit::Command::Failed): Exception while executing on host 192******: curl exit status: 1
curl stdout: # Executing docker install script, commit: c2de0811708b6d9015ed1a2c80f02c9b70c8ce7b

ERROR: Unsupported operating system 'macOS'
Please get Docker Desktop from https://www.docker.com/products/docker-desktop
curl stderr: Nothing written

------------------
I have installed docker desktop manually and docker -v responds with version but still seeing this issue.
ENV- Mac mini M2

dhh:
Kamal was not designed for deployment on macOS. Only from macOS.

filipesilva:
This seems related to https://github.com/capistrano/sshkit?tab=readme-ov-file#the-command-map, how the SSH session won't have the same `$PATH` as an interactive session. In my mac, docker is added to `$PATH` under `/usr/local/bin/docker`.

Perhaps supporting the `command_map` or `default_env` in deploy.yml's `sshkit` section could help, but even then there might be other issues.
```
sshkit:
  default_env:
    path: "/usr/local/bin:$PATH"
```

I see how this wouldn't make much sense since this isn't really a supported case. Just leaving my attempts here in case someone else is trying it. 

I debugged using `kamal setup -v`, which gave me a lot more logging. I tried to make it work by adding the `default_env` to https://github.com/basecamp/kamal/blob/69867e2650051abd3ed4e9d1e1ee7215296f62dd/lib/kamal/commander.rb#L169

I found the locally installed file in my rbenv gems at `/Users/my-user/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/kamal-2.3.0/lib/kamal/commander.rb`:

```ruby
[...]

PR #1279, opened 2024-12-09
Support sshkit.default_env

filipesilva:
Enables deploy to MacOS hosts.

Will still need further configuration though:
- `default_env` needs to be set to a PATH that contains docker and coreutils `cp`
- desktop docker config needs to have no `credsStore`
- shell needs to be changed to `bash`

All this can be done by the user though, whereas passing on `default_env` to SSHKit needed to happen on Kamal.

cc @dhh, @HLFH

See https://github.com/basecamp/kamal/issues/432#issuecomment-2505075808 for details.


filipesilva:
I don't have any Ruby experience, so if I'm doing something very silly here please let me know so I can fix it.

HLFH:
@filipesilva I have fixed the zsh issue in this PR: https://github.com/basecamp/kamal/pull/1281. You can probably keep zsh?

PR #1280, opened 2024-12-09
Enable IPv6 connectivity for the kamal network

HLFH:
This commit enables IPv6 connectivity for the kamal network.


frenkel:
@djmb you can close this one as well as it has the same issue as #1381.

PR #1284, opened 2024-12-11
Fix JSON ParserError with kamal secrets extract $SECRETS

chrisdebruin:
When running `kamal secrets extract TEST $SECRETS` and value of TEST contains a `(` the code was throwing a `JSON::ParserError`

```
secrets = { TEST: "value(" }.to_json.shellescape
JSON.parse(secrets)
```

### DIFF
--- a/lib/kamal/cli/secrets.rb
+++ b/lib/kamal/cli/secrets.rb
@@ -19,7 +19,7 @@ class Kamal::Cli::Secrets < Kamal::Cli::Base
   desc "extract", "Extract a single secret from the results of a fetch call"
   option :inline, type: :boolean, required: false, hidden: true
   def extract(name, secrets)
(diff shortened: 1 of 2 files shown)


djmb:
This feels like a problem with the output from `kamal secrets fetch` command rather than something to fix in the extract command. Do you know why the output is escaped like this?

chrisdebruin:
> This feels like a problem with the output from `kamal secrets fetch` command rather than something to fix in the extract command. Do you know why the output is escaped like this?

Yes it's because of fetch is doing a `shellescape` on the json data. In this case the `(` is also escaped even though it is inside `""`


```
{ TEST: "value(" }.to_json.shellescape => "\\{\\\"TEST\\\":\\\"value\\(\\\"\\}"

PR #1319, opened 2024-12-22
Fix bug in role validate_servers.

ShPakvel:
There were typo-bug during `validate_servers!` invocation for role. It wasn't discovered, because it never met condition. Because role_config wasn't correctly extracted for validation.

Also remove not used anymore `accessories_on`. Leftover from previous changes.

### DIFF
--- a/lib/kamal/commander.rb
+++ b/lib/kamal/commander.rb
@@ -76,11 +76,6 @@ class Kamal::Commander
     config.accessories&.collect(&:name) || []
   end
 
-  def accessories_on(host)
-    config.accessories.select { |accessory| accessory.hosts.include?(host.to_s) }.map(&:name)
-  end
-
-
   def app(role: nil, host: nil)
     Kamal::Commands::App.new(config, role: role, host: host)
   end
@@ -128,7 +123,6 @@ class Kamal::Commander
   def alias(name)
     config.aliases[name]
   end
-
 
   def with_verbosity(level)
     old_level = self.verbosity
--- a/lib/kamal/configuration/role.rb
+++ b/lib/kamal/configuration/role.rb
@@ -10,7 +10,7 @@ class Kamal::Configuration::Role
   def initialize(name, config:)
     @name, @config = name.inquiry, config
     validate! \
-      specializations,
+      role_config,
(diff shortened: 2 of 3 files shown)


djmb:
Hi @ShPakvel - is it possible to write a test for this condition?

ShPakvel:
> Hi @ShPakvel - is it possible to write a test for this condition?

### Regarding to fix for syntax bug with wrong number of arguments

Existing tests already cover it. Here is example of test execution.
As I wrote, it wasn't caught by tests because of another logical bug interfered and prevented flow to validation.
```bash
$ ./bin/test test/configuration/role_test.rb

# ...

Error:
ConfigurationRoleTest#test_cmd:
ArgumentError: wrong number of arguments (given 2, expected 1)
    lib/kamal/configuration/validator.rb:110:in `validate_servers!'
    lib/kamal/configuration/validator/role.rb:6:in `validate!'
    lib/kamal/configuration/validation.rb:21:in `validate!'
    lib/kamal/configuration/role.rb:12:in `initialize'

# ...

bin/test /home/shpakvel/projects/github/kamal/test/configuration/role_test.rb:32

Finished in 0.069962s, 285.8696 runs/s, 57.1739 assertions/s.
20 runs, 4 assertions, 0 failures, 18 errors, 0 skips
```

And tests are passed when code is corrected.
```bash
$ ./bin/test test/configuration/role_test.rb
Run options: --seed 49549

# Running:

....................

Finished in 0.190348s, 105.0708 runs/s, 262.6770 assertions/s.
20 runs, 50 assertions, 0 failures, 0 errors, 0 skips
```

### Regarding logical bug.

- It looks like simple overlooked difference between `specializations` and `role_config`. And after this change, difference is clearly stated in code logic.
- And right now it is private part of class.

Which leads us to choose between these ways to proceed:
- A - to consider this part as internal (private) that shouldn't be tested.
- B - to make `specializations` and `role_config` methods public, and right tests on what they should return.

@djmb, Let me know which way you prefer `A` or `B`. If it is `B`, I will proceed with changes and tests creation.

djmb:
Ok, that makes sense, thanks! All good to merge as it is then

PR #1320, opened 2024-12-22
[Feature] Registry for accessory 

ShPakvel:
#  Context

Currently we can specify only one root level registry with main purpose of deploying our application.

Accessories by default use Docker Hub registry to pull from.
And there is option to specify root level registry server as a prefix for an accessory image, to pull them from the same root level registry.

# Feature proposal

**This PR proposes an option to specify different registry per accessory.**
- The accessory's registry data structure are the same as root level registry has.
- The option is used and validated when provided.
- It shouldn't break any existing configurations or behavior.
- Test cases to cover new option is added. Most of tests are unchanged, so previous behavior is not broken. Extended deploy config `test/fixtures/deploy_with_accessories_with_different_registries.yml` is added to test `test/cli/accessory_test.rb` with more thorough coverage of registries.

## Example

Note: Anchor was used to eliminate duplication. 

```yml
service: app
image: the_cake/is_a_lie
servers:
  - 1.1.1.1
registry:
  server: client.registry
  username: client_user
  password: client_password
builder:
  arch: amd64
x-organization-registry: &organization-registry
  server: organization.registry
  username: organization_user
  password: organization_password
accessories:
  private-organization-tool-xxx:
    image: organization_tool_xxx:latest
    host: 1.1.1.1
    port: 3333
    registry:
      <<: *organization-registry
  private-organization-tool-yyy:
    image: organization_tool_yyy:latest
    host: 1.1.1.1
    port: 3334
    registry:
      <<: *organization-registry
  private-client-tool:
    image: client.registry/client_tool:latest
    host: 1.1.1.1
    port: 4444
  redis:
    image: redis:latest
    roles:
      - web
    port: 6379
    directories:
      - data:/data
```

### DIFF
--- a/lib/kamal/commands/registry.rb
+++ b/lib/kamal/commands/registry.rb
@@ -1,14 +1,16 @@
 class Kamal::Commands::Registry < Kamal::Commands::Base
-  delegate :registry, to: :config
+  def login(registry_config: nil)
+    registry_config ||= config.registry
 
-  def login
     docker :login,
-      registry.server,
-      "-u", sensitive(Kamal::Utils.escape_shell_value(registry.username)),
-      "-p", sensitive(Kamal::Utils.escape_shell_value(registry.password))
+      registry_config.server,
+      "-u", sensitive(Kamal::Utils.escape_shell_value(registry_config.username)),
+      "-p", sensitive(Kamal::Utils.escape_shell_value(registry_config.password))
   end
 
-  def logout
-    docker :logout, registry.server
+  def logout(registry_config: nil)
+    registry_config ||= config.registry
+
+    docker :logout, registry_config.server
   end
 end
--- a/lib/kamal/cli/accessory.rb
+++ b/lib/kamal/cli/accessory.rb
@@ -292,7 +292,7 @@ class Kamal::Cli::Accessory < Kamal::Cli::Base
     def prepare(name)
       with_accessory(name) do |accessory, hosts|
         on(hosts) do
-          execute *KAMAL.registry.login
+          execute *KAMAL.registry.login(registry_config: accessory.registry)
           execute *KAMAL.docker.create_network
         rescue SSHKit::Command::Failed => e
           raise unless e.message.include?("already exists")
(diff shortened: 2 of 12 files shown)


lib/kamal/commands/registry.rb
```diff
@@ -1,14 +1,16 @@
 class Kamal::Commands::Registry < Kamal::Commands::Base
-  delegate :registry, to: :config
+  def login(registry_config: nil)
+    registry_config ||= config.registry
```
djmb, line 3:
```suggestion
  def login(registry_config: config.registry)
```

lib/kamal/commands/registry.rb
```diff
@@ -1,14 +1,16 @@
...
-      "-p", sensitive(Kamal::Utils.escape_shell_value(registry.password))
+      registry_config.server,
+      "-u", sensitive(Kamal::Utils.escape_shell_value(registry_config.username)),
+      "-p", sensitive(Kamal::Utils.escape_shell_value(registry_config.password))
   end
 
-  def logout
-    docker :logout, registry.server
+  def logout(registry_config: nil)
+    registry_config ||= config.registry
```
djmb, line 12:
```suggestion
  def logout(registry_config: config.registry)
```

djmb (changes requested):
This is great @ShPakvel!

Just a couple of very minor suggestions.

lib/kamal/commands/registry.rb
ShPakvel: With this change we will have bugs in case when accessory doesn't have own registry configured. Because it is optional config for accessories.
`@registry = initialize_registry if accessory_config["registry"].present?`

So when invoke for accessory login to registry it can pass nil which will override default argument value suggested here.
`execute *KAMAL.registry.login(registry_config: accessory.registry)`
And it will lead to errors on following logic, like at `registry_config.username` invocation.

So the current implementation with registry set to default one after argument is passed is intentional to eliminate issues on cases when accessory doesn't have own registry and we fallback to login main registry instead.

There is another way to go: assign main registry to accessory in case when specific registry is not provided. Which can be carefully considered and applied in separate PR I think. Because current implementation seems to be simple enough (just one line to fallback on login and logout).

@djmb is it ok for you to proceed with what is implemented in this PR for now, instead of investigating main registry assignment for accessories during initialization?

djmb: Ah yes I see what you mean, please ignore my suggestions then!

PR #1345, opened 2025-01-08
fix the digital ocean registry example

exterm:
It didn't work as shown.

Actually I'm a little confused - the same example exists in https://github.com/basecamp/kamal-site. Do I need to fix it there too?

### DIFF
--- a/lib/kamal/configuration/docs/registry.yml
+++ b/lib/kamal/configuration/docs/registry.yml
@@ -9,11 +9,10 @@
 # A reference to a secret (in this case, `DOCKER_REGISTRY_TOKEN`) will look up the secret
 # in the local environment:
 registry:
-  server: registry.digitalocean.com
-  username:
-    - DOCKER_REGISTRY_TOKEN
+  server: registry.digitalocean.com/<your-registry>
+  username: <your-digitalocean-username>
   password:
-    - DOCKER_REGISTRY_TOKEN
+    - DIGITALOCEAN_PERSONAL_ACCESS_TOKEN
 
 # Using AWS ECR as the container registry
 #


nickhammond:
@exterm https://github.com/basecamp/kamal-site/pull/155 Does this update work for you?

exterm:
@nickhammond partially - I do think the registry URL needs to include the name of the registry (`registry.digitalocean.com/<your-registry>`).

exterm:
Also - why is the documentation here and there? Where do we need to change it?

djmb:
@exterm - we should change it here, there's a `bin/docs` script that generates the files for the doc site from this, that I run when doing a new release.

kmctown:
This is what I have for DOCR and it has been working:

```
registry:
  server: registry.digitalocean.com
  username: <%= ENV["DOCR_PUSH_API_TOKEN"] %>
  password: <%= ENV["DOCR_PUSH_API_TOKEN"] %>
```

exterm:
@kmctown curious! Does the value for your `image:` key include your registry name?

jorgegonzalez:
Sorry to hijack this issue, but is anyone else unable to authenticate with `dotenv bin/kamal registry login` while trying to reference `KAMAL_REGISTRY_PASSWORD`? I have it set in my `.env`, and the same setup works if I were using the default Dockerhub registry. Seems to only be an issue with DigitalOcean. I am passing in my email as the username. If I add the password directly into my `deploy.yml` then the registry login will work. I also have `KAMAL_REGISTRY_PASSWORD=$KAMAL_REGISTRY_PASSWORD` in my `.kamal/secrets`.

exterm:
@kmctown could you post your image name too? I'm curious why this would work for you and not for me.

kmctown:
> @kmctown could you post your image name too? I'm curious why this would work for you and not for me.

@exterm interesting. So my setup looks like this:

```
service: my-app
image: my-org/my-app

registry:
  server: registry.digitalocean.com
  username: <%= ENV["DOCR_PUSH_API_TOKEN"] %>
  password: <%= ENV["DOCR_PUSH_API_TOKEN"] %>
```

And my repository name in the DO Container Registry is `my-app`.

PR #1355, opened 2025-01-14
Use local docker registry to push and pull app images

npezza93:
Allow applications to be deployed without needing to set up a repository in a remote Docker registry.

If the registry server starts with `localhost`, Kamal will start a local docker registry on that port and push the app image to it.

Then when pulling the image onto the servers, we use net-ssh to forward the that port from the app server to the deployment server.

This is branched off @djmb 's initial work. I have this working on a simple app but curious to open this up to try out to find issues and possibly optimize. Saw an overall speed up of around 10-15 seconds. 

To get this to work update your deploy.yml files registry section to look something like:
```
registry:
  server: "localhost:5555"
```


lib/kamal/commands/registry.rb
```diff
@@ -11,4 +14,18 @@ def login
   def logout
     docker :logout, registry.server
   end
+
+  def setup
+    combine \
+      docker(:start, "kamal-docker-registry"),
+      docker(:run, "--detach", "-p", "127.0.0.1:#{local_port}:5000", "--name", "kamal-docker-registry", "registry:2"),
```
djmb, line 21: Publishing the port on your local machine will make it available in the same network segment (see https://github.com/moby/moby/issues/45610). That's why I didn't go any further with this at the time.

I think we can mitigate this though by generating a password for the registry when we boot it.

npezza93: Hmm looks like a pr was merged a few days ago to resolve that. Perhaps in the next release of buildkit it'll be fixed for us?

djmb: Oh awesome! This will be great 🎉

ShPakvel (commented):
Thanks again for working on it ❤️ 🚀

lib/kamal/cli/build.rb
```diff
@@ -60,14 +60,16 @@ def push
...
   def pull
-    if (first_hosts = mirror_hosts).any?
-      #  Pull on a single host per mirror first to seed them
-      say "Pulling image on #{first_hosts.join(", ")} to seed the #{"mirror".pluralize(first_hosts.count)}...", :magenta
-      pull_on_hosts(first_hosts)
-      say "Pulling image on remaining hosts...", :magenta
-      pull_on_hosts(KAMAL.hosts - first_hosts)
-    else
-      pull_on_hosts(KAMAL.hosts)
+    Kamal::Cli::PortForwarding.new(KAMAL.hosts, KAMAL.config.registry.local_port).forward do
```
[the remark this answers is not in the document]
npezza93: Hey @ShPakvel, I like the suggestion around making the PortForwarding class usable elsewhere and not tying it to if the registry is local. Didnt really follow the code suggestion but made some edits to make the class usable elsewhere. Let me know what you think

test/fixtures/deploy_with_local_registry_and_accessories.yml
```diff
@@ -0,0 +1,37 @@
...
+      - test/fixtures/files/my.cnf:/etc/mysql/my.cnf
+    directories:
+      - data:/var/lib/mysql
+  redis:
+    image: redis:latest
+    roles:
+      - web
+    port: 6379
+    directories:
+      - data:/data
```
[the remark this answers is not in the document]
npezza93: should be covered now

lib/kamal/commands/registry.rb
npezza93: Looks like this is now released: https://docs.docker.com/engine/release-notes/28/

driveton:
I'm a solodeveloper and I was using Cloudflare tunnels to connect to a local registry but this is much better. Thank you @npezza93 

I wonder if can use the same configuration for local and remote registry to avoid "works on machine". Changes here:  https://github.com/driveton/kamal/tree/local-registry-branch

- Fail if there's no image set. Remove this so there's no magic here.
```
def image
    name = raw_config&.image.presence
    name ||= raw_config&.service if registry.local?

    name
end
```
- We can keep login/logout from registry and add setup/remove.
```
➜   git:(main) ✗ bundle exec kamal registry help 
Commands:
  kamal registry help [COMMAND]  # Describe subcommands or one specific subcommand
  kamal registry login           # Log in to registry locally and remotely
  kamal registry logout          # Log out of remote registry locally and remotely
  kamal registry remove          # Remove local registry
  kamal registry setup           # Setup local registry

```
- We can have a local: true/false and a registry_port in deploy.yml and/or have a deploy.localhost.yml and secrets-localhost.
```
def local?
    registry_config["local"] == true
end

def registry_port ## Instead of local_port
   registry_config.fetch("registry_port", 5000)
end
```
- Fail if username/password are not, same as remote registry even if they are not used.
```
registry:
  server: localhost:5000
  username: localuser
  password:
    - KAMAL_REGISTRY_PASSWORD
  registry_port: 5000 ## Similar to app_port
  local: true
```
- Forward ports explicitly only when the registry is local
```
def pull(forward_registry_port: KAMAL.config.registry.local?)
    login_to_registry_remotely

    if forward_registry_port
      return Kamal::Cli::PortForwarding.new(KAMAL.app_hosts, KAMAL.config.registry.registry_port).forward do
        pull(forward_registry_port: false)
      end
    end

    if (first_hosts = mirror_hosts).any?
      # Pull on a single host per mirror first to seed them
      say "Pulling image on #{first_hosts.join(", ")} to seed the #{"mirror".pluralize(first_hosts.count)}...", :magenta
      pull_on_hosts(first_hosts)
      say "Pulling image on remaining hosts...", :magenta
      pull_on_hosts(KAMAL.app_hosts - first_hosts)
    else
      pull_on_hosts(KAMAL.app_hosts)
    end
  end
```

simonleva:
Can we make this work when running Kamal inside the devcontainer?

npezza93:
> Can we make this work when running Kamal inside the devcontainer?

I have no need for such a feature but you are more than welcome to open a pr against this and I can include it

simonleva:
Sure, it's kind of tricky though. I'm not sure how to solve it. 

My understanding is that Kamal, at least for Rails, is "best used" inside the devcontainer

- If we create the local registry in the host, we need to the app devcontainer where kamal is running to see the local registry in the host.
- If we create the local registry as a service the devcontainer, we need to tunnel the remote sever to the registry running in the devcontainer (in this scenario you don't need to setup the registry, just add it as service in the compose file).

Docker expects https so for any of those scenarios to work, we need to "hack" to accept http or to install certificates which is cumbersome.

airblade:
It would be great to merge this.  Currently I use ghcr.io and it is randomly totally unusable due to `docker pull` slowing to a crawl.  I wish I could just push my images directly to my server from my local machine.

khaled-badenjki:
This would be a great feature to have. what is blocking this MR to be merged? is any help needed?

PR #1357, opened 2025-01-15
Introduce a "build dev" command

flavorjones:
## Context

The problem I'm trying to solve with this PR is: During the development process, how can a developer iterate quickly on their container?

The primary obstacles to using Kamal to build development images in this way are:

- `build push` uses the config to determine the build context, and the default is to make a pristine build from a clean git clone. There's no command option override that configuration for a single invocation.
- `build push` will push to the registry every time! For this use case, it's more likely the developer will simply want to push to the local docker image store.

A secondary obstacle is the danger of accidentally using a development image (built from the dirty working directory) in production! If we're going to support development images, we should probably tag them explicitly to avoid confusion.

### Alternatives considered

It's possible, but not trivial, to build the container in a manner that addresses this use case by running `docker buildx` from the commandline. However, for containers with build steps that require secrets to be injected, this can quickly become very complicated.

### Proposed solution

I've broken this feature out into two commits.

First, I introduce some flexibility into the `kamal build push` command with an `--output` option which allows Kamal users to specify where they would like the build result to be exported. The default value is "registry" to reflect the current behavior of `build push`.

Any value provided to this option will be passed to the `docker buildx build` command as a `--output=type=<VALUE>` flag. For example, the following command will push to the local docker image store:

    kamal build push --output=docker

Second, I introduce a new command, `build dev` which is similar to `build push` but:

- always uses the local working directory as the build context
- exports the build result to the local docker image store
- tags the build result with a `-dirty` suffix

The command also supports the `--output` option.

### Other notes

Documentation PR is https://github.com/basecamp/kamal-site/pull/158 (or will be once I update it).

### DIFF
--- a/lib/kamal/cli/build.rb
+++ b/lib/kamal/cli/build.rb
@@ -109,6 +109,28 @@ class Kamal::Cli::Build < Kamal::Cli::Base
     end
   end
 
+  desc "dev", "Build using the working directory, tag it as dirty, and push to local image store."
+  option :output, type: :string, default: "docker", banner: "export_type", desc: "Exported type for the build result, and may be any exported type supported by 'buildx --output'."
+  def dev
+    cli = self
+
+    ensure_docker_installed
+
+    uncommitted_changes = Kamal::Git.uncommitted_changes
+    if uncommitted_changes.present?
+      say "WARNING: building with uncommitted changes:\n #{uncommitted_changes}", :yellow
+    end
+
+    with_env(KAMAL.config.builder.secrets) do
+      run_locally do
+        build = KAMAL.builder.push(cli.options[:output], tag_as_dirty: true)
+        KAMAL.with_verbosity(:debug) do
+          execute(*build)
+        end
+      end
+    end
+  end
+
   private
     def connect_to_remote_host(remote_host)
       remote_uri = URI.parse(remote_host)
--- a/lib/kamal/cli/build.rb
+++ b/lib/kamal/cli/build.rb
@@ -116,9 +116,22 @@ class Kamal::Cli::Build < Kamal::Cli::Base
 
     ensure_docker_installed
 
-    uncommitted_changes = Kamal::Git.uncommitted_changes
-    if uncommitted_changes.present?
-      say "WARNING: building with uncommitted changes:\n #{uncommitted_changes}", :yellow
+    docker_included_files = Set.new(Kamal::Docker.included_files)
+    git_uncommitted_files = Set.new(Kamal::Git.uncommitted_files)
+    git_untracked_files = Set.new(Kamal::Git.untracked_files)
+
+    docker_uncommitted_files = docker_included_files & git_uncommitted_files
+    if docker_uncommitted_files.any?
+      say "WARNING: Files with uncommitted changes will be present in the dev container:", :yellow
+      docker_uncommitted_files.sort.each { |f| say "  #{f}", :yellow }
+      say
+    end
+
+    docker_untracked_files = docker_included_files & git_untracked_files
+    if docker_untracked_files.any?
+      say "WARNING: Untracked files will be present in the dev container:", :yellow
+      docker_untracked_files.sort.each { |f| say "  #{f}", :yellow }
+      say
     end
 
     with_env(KAMAL.config.builder.secrets) do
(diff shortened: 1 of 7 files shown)


nickhammond:
@flavorjones Does this introduce a different behavior than setting your build context to the current directory with `context: .`? 

https://kamal-deploy.org/docs/configuration/builders/#build-context

flavorjones:
@nickhammond The idea behind this is to build a one-off container that doesn't get pushed anywhere. Part of the behavior is to build with the current directory as context. The different behaviors are described in the description above.

nickhammond:
@flavorjones Is the use case to warm the build cache? Or just to verify that your image is building successfully without dealing with the login and push process? 

It's nice to have the standalone build process but `dev` made me think it was something you'd use in your development environment, similar to say if you develop locally utilizing containers(https://github.com/rails/docked). 

What about `kamal build local`?

flavorjones:
@nickhammond The use case here is that I want to iterate on the contents of my Dockerfile.

In particular, I was running into an issue with packages installed on the base image, so I wanted to quickly run through the process of:

1. make change to `Dockerfile`
2. build image
3. run test in the image using `docker run`

It's possible to build the image with `docker build`, but in particular if I've got secrets used during the build process, it's troublesome to either inject them or work around their absence.

I got some feedback on this PR from my colleagues at 37signals that it could probably be merged into `build push` with a `--local` flag or something similar. I'll be revising this taking that feedback into account.

flavorjones:
I've rewritten this PR to decouple the build result export from the dirty working directory features. See the description for a longer explanation.

flavorjones:
Converting to draft until I have time to come back and get it green.

lib/kamal/cli/build.rb
```diff
@@ -108,6 +109,28 @@ def details
...
+  desc "dev", "Build using the working directory, tag it as dirty, and push to local image store."
+  option :output, type: :string, default: "docker", banner: "export_type", desc: "Exported type for the build result, and may be any exported type supported by 'buildx --output'."
+  def dev
+    cli = self
+
+    ensure_docker_installed
+
+    uncommitted_changes = Kamal::Git.uncommitted_changes
+    if uncommitted_changes.present?
+      say "WARNING: building with uncommitted changes:\n #{uncommitted_changes}", :yellow
```
djmb, line 121: If there are any files that are git-ignored, but not docker-ignored, then they'll end up in the image as well but we won't report them here. This is the reason for the git clone before building.

The worry I'd have is that the files could contain secrets. Maybe not the end of the world if the images are not pushed remotely, but it would be nice to warn the user about them.

I'm not sure there's a way to work out dockerignored files without creating an image anyway so maybe we could just add a generic warning here?

flavorjones: I tried to communicate "don't push this image anywhere" by naming it with the `-dirty` tag. I can make this warning more generic (and just skip the uncommited changes check in that case, yes?) for sure.

What if I also renamed the command from `kamal build dev` to `kamal build dirty` just to reinforce the notion that this container is not for polite company or pushing to a registry?

djmb: I think `kamal build dev` communicates the purpose or the command better. And I guess the `-dirty` suffix is a pretty good warning not to push it further I think.

My main worry if that the uncommitted file list creates the impression those are the only files that you need to worry about. So I think maybe a generic warning might be best. We could suggest running `git status --ignored` to see all extra files that are potentially going to end up in the image.

flavorjones: @djmb I've pushed a commit that provides a detailed look at untracked and uncommitted files that are being packaged in the container. Output looks like:

```
WARNING: Files with uncommitted changes will be present in the dev container:
  Rakefile
  bin/kamal

WARNING: Untracked files will be present in the dev container:
  .ruby-lsp/.gitignore
  .ruby-lsp/Gemfile
  .ruby-lsp/Gemfile.lock
  .ruby-lsp/last_updated
  .ruby-lsp/main_lockfile_hash
  build.log
  db/cache_primary_shard1_schema_cache.yml
  db/cache_primary_shard2_schema_cache.yml
  db/cache_secondary_shard1_schema_cache.yml
  db/cache_secondary_shard2_schema_cache.yml
  db/schema_cache.yml
  db/solid_queue_primary_schema_cache.yml
  foo.log
  node_modules/read-pkg/node_modules/resolve/test/resolver/symlinked/_/symlink_target/.gitkeep
  node_modules/resolve/test/resolver/symlinked/_/symlink_target/.gitkeep
  node_modules/tailwindcss/types/generated/.gitkeep
  test.log
```

djmb: That's great @flavorjones!

exterm:
Thank you a lot for this @flavorjones this is just what I was missing last week when debugging our image build!

PR #1381, opened 2025-01-27
Enable IPv6 connectivity for the kamal network

frenkel:
Based on #1280 I've updated the tests as well.
This change is needed if you want to receive the client IPv6 address in your Rails app. Without it you'll only get the IPv4 address of the docker proxy (172.x.x.x).

### DIFF
--- a/lib/kamal/commands/docker.rb
+++ b/lib/kamal/commands/docker.rb
@@ -20,7 +20,7 @@ class Kamal::Commands::Docker < Kamal::Commands::Base
   end
 
   def create_network
-    docker :network, :create, :kamal
+    docker :network, :create, "--ipv6", :kamal
   end
 
   private
--- a/test/cli/accessory_test.rb
+++ b/test/cli/accessory_test.rb
@@ -32,9 +32,9 @@ class CliAccessoryTest < CliTestCase
       assert_match "docker login private.registry -u [REDACTED] -p [REDACTED] on 1.1.1.1", output
       assert_match "docker login private.registry -u [REDACTED] -p [REDACTED] on 1.1.1.2", output
       assert_match "docker login other.registry -u [REDACTED] -p [REDACTED] on 1.1.1.3", output
-      assert_match /docker network create kamal.*on 1.1.1.1/, output
-      assert_match /docker network create kamal.*on 1.1.1.2/, output
-      assert_match /docker network create kamal.*on 1.1.1.3/, output
+      assert_match /docker network create --ipv6 kamal.*on 1.1.1.1/, output
+      assert_match /docker network create --ipv6 kamal.*on 1.1.1.2/, output
+      assert_match /docker network create --ipv6 kamal.*on 1.1.1.3/, output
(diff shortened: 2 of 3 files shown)


djmb:
The integration tests are failing with "Error response from daemon: could not find an available, non-overlapping IPv6 address pool among the defaults to assign to the network".

This might just be from running docker inside docker, but it highlights that maybe this isn't a safe change in all contexts.

From the latest changes in main, the pre-deploy hook will always run before attempting to create the network, so the best solution here might be to use that to create a kamal network with whatever options you want.

frenkel:
Indeed, you would need some options in your docker file (or pass them as arguments). Hadn't thought about that and locally I did have those probably.

pre-deploy seems like a better solution, thanks for the hint! You might want to close #1280 then as well.

disbelief:
So neither of the PRs addressing this appear to have been merged, or have I missed something?

acidtib:
@disbelief did you read the comments in the closed PR?

disbelief:
@acidtib I don't see any comments. What am I missing?

https://github.com/basecamp/kamal/pull/1280

frenkel:
Just create the kamal network yourself with the ipv6 option before deploying anything and you're good to go. You might need to enable some options in daemon.json, like this:
```
{
	"userland-proxy": false,
	"ipv6": true,
	"ip6tables": true,
	"fixed-cidr-v6": "2001:db8::/32",
	"experimental": true,
	"default-address-pools": [
		{ "base": "172.17.0.0/16", "size": 16 },
		{ "base": "172.18.0.0/16", "size": 16 },
		{ "base": "172.19.0.0/16", "size": 16 },
		{ "base": "172.20.0.0/14", "size": 16 },
		{ "base": "172.24.0.0/14", "size": 16 },
		{ "base": "172.28.0.0/14", "size": 16 },
		{ "base": "192.168.0.0/16", "size": 20 },
		{ "base": "2001:db8::/104", "size": 112 }
	]
}
```

unasuke:
The CI failure that led to closing this PR no longer reproduces. It was caused by Docker < 27.0 lacking automatic IPv6 address pool allocation. Docker 27.0 ([released June 2024](https://redirect.github.com/moby/moby/discussions/48053)) introduced automatic ULA allocation, so `docker network create --ipv6` now works without explicit `--subnet` or `default-address-pools` configuration. (investigated by claude, but I've confirmed too)

I've verified this on a fork — all CI jobs pass

* https://github.com/unasuke/kamal/commit/a5771405ee1c94d4c08964ef84933aa80fc02ac7
* https://github.com/unasuke/kamal/actions/runs/23303350611

The suggested workaround (pre-creating the network via a hook) has a discoverability problem: https://kamal-deploy.org/docs/ does not mention that Kamal creates a `kamal` Docker network, nor how to customize it. Users on IPv6-only hosts (it's my own use-case!) have no way to discover this workaround without reading the source code.

The `--ipv6` flag is safe on dual-stack hosts and essential on IPv6-only hosts. Would you be open to reconsidering this change?

PR #1392, opened 2025-02-03
Allow omitting AWS account parameter while fetching secrets

neiljohari:
# Problem

When using OIDC in GHA to assume a role to execute commands via AWS cli, you do not have a profile.

For instance, running `aws configure list` after using OIDC:
```
      Name                    Value             Type    Location
      ----                    -----             ----    --------
   profile                <not set>             None    None
access_key     ****************AMST              env    
secret_key     ****************JxL9              env    
    region                us-east-1              env    ['AWS_REGION', 'AWS_DEFAULT_REGION']
```

You don't get any profile.

Also you'd expect `kamal secrets fetch --adapter aws_secrets_manager` to use whatever profile is exported as `AWS_DEFAULT_PROFILE`, even if it's pointing to something other than the actual profile named `default` if it exists. 

# Solution

When calling `aws secretsmanager batch-get-secret-value`, it's not actually required to pass an account id. If omitted, it'll follow the CLI's general precedence order for determining what account to use.

So, we can modify the adapter to allow omitting the account option and defaulting it to `nil`. Then, we can only add the argument conditionally if passed in.

This will correctly enable use cases like deployment via GHA and OIDC.

### DIFF
--- a/lib/kamal/secrets/adapters/aws_secrets_manager.rb
+++ b/lib/kamal/secrets/adapters/aws_secrets_manager.rb
@@ -48,4 +48,3 @@ class Kamal::Secrets::Adapters::AwsSecretsManager < Kamal::Secrets::Adapters::Ba
       $?.success?
     end
 end
-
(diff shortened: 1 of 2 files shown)


neiljohari:
@djmb Thanks for running CI, I've gone ahead and fixed the formatting issue. Could we re-run the workflows? Should all pass now hopefully 😄

PR #1402, opened 2025-02-06
Fix the docker build

djmb:
Somewhere along the way the docker build broke, it now needs yaml-dev to be installed. Maybe the underlying image changed?

Update to 3.4-alpine while we are here.


morgoth:
@djmb Yes, there was a change in the source image, however it was later reverted.
See https://github.com/rails/rails/pull/54237

Issue #1399, opened 2025-02-04
Issue parsing --env for `app exec`


mktcode:
I'm trying to reset my production database with this command:

`kamal app exec --primary --env=DISABLE_DATABASE_ENVIRONMENT_CHECK:1 'bin/rails db:reset'`

It turns into this docker command:

`docker run --rm --network kamal --env SOLID_QUEUE_IN_PUMA="true" --env JOB_CONCURRENCY="3" --env WEB_CONCURRENCY="2" --env DB_HOST="xxx.xxx.xxx.xxx" --env-file .kamal/apps/market/env/roles/web.env --env DISABLE_DATABASE_ENVIRONMENT_CHECK="1" --env bin/rails db="reset" --log-opt max-size="10m" --volume market_storage:/rails/storage mktcode/market:latest `

Throwing this error:

```
ERROR (SSHKit::Command::Failed): Exception while executing on host xxx.xxx.xxx.xxx: docker exit status: 125
docker stdout: Nothing written
docker stderr: docker: invalid reference format.
```

This seems to be wrong: `--env bin/rails db="reset"`

`kamal app exec --help` says:

```
-e, [--env=key:value]     # Set environment variables for the command
```

Anything I'm doing wrong or is this a bug in the parsing of the env argument?

mike-weiner:
This does look (and smell) like a parsing issue.

What `docker` command is generated when `kamal app exec --primary 'bin/rails db:reset' --env=DISABLE_DATABASE_ENVIRONMENT_CHECK:1` is used?

After looking at the description of `kamal app exec` command, I would expect that the command you want to run should be first followed by any options.

```
weiner ~ % kamal app exec --help 
Usage:
  kamal app exec [CMD...]

...
```

mike-weiner:
I tried to recreate this bug with my own deployment by running: `kamal app exec --primary --env=TEST:1 '/usr/bin/cat package.json'`

The command worked without issue on `kamal v2.5.0`:

```
weiner % kamal version                                                    
2.5.0
```

```
weiner % kamal app exec --primary --env=TEST:1 '/usr/bin/cat package.json'
Get most recent version available as an image...
Launching command with version latest from new container...
  INFO [3538b338] Running docker run --rm --network kamal --env PORT="3000" --env TEST="1" --log-opt max-size="10m" registry.hub.docker.com/<REDACTED>/<REDACTED>:latest /usr/bin/cat package.json on <REDACTED>
[...]

PR #1405, opened 2025-02-08
Fail kamal app exec without a CMD

mike-weiner:
Related to: https://github.com/basecamp/kamal/issues/1399

It appears that there is an issue with `Thor` and parsing `key:value` pairs that results in the vague and confusing error below. 

```
weiner % kamal app exec --primary --env=TEST:1 '/usr/bin/echo :test'      
Get most recent version available as an image...
Launching command with version latest from new container...
  INFO [ab8c5ab1] Running docker run --rm --network kamal --env PORT="3000" --env TEST="1" --env /usr/bin/echo ="test" --log-opt max-size="10m" registry.hub.docker.com/<REDACTED>/<REDACTED>:latest  on <REDACTED>
  ERROR (SSHKit::Command::Failed): Exception while executing on host <REDACTED>: docker exit status: 125
docker stderr: docker: invalid reference format.
See 'docker run --help'.
```

The parsing error is not present when command and options are reordered slightly: `kamal app exec '/usr/bin/echo :test' --primary --env=TEST:1`

`thor` appears to be incorrectly parsing `/usr/bin/echo :test` as an `--env` option. My stance is that we can return a better error to the user when this invalid parsing takes place.

### DIFF
--- a/lib/kamal/cli/app.rb
+++ b/lib/kamal/cli/app.rb
@@ -108,6 +108,10 @@ class Kamal::Cli::App < Kamal::Cli::Base
       raise ArgumentError, "Detach is not compatible with #{incompatible_options.join(" or ")}"
     end
 
+    if cmd.empty?
+      raise ArgumentError, "No command provided. You must specify a command to execute."
+    end
+
     cmd = Kamal::Utils.join_commands(cmd)
     env = options[:env]
     detach = options[:detach]


mike-weiner:
@djmb - This is ready for another workflow approval for PR checks. Thanks!

PR #1420, opened 2025-02-21
Fix Docker Build under Ruby 3.4

### LINKED ISSUE #1411: Running Kamal via Docker - update the package
Hello,

I would be nice to update the package to the latest kamal's version.

Thanks

smartygus:
Docker Build was failing after updating to use 3.4-alpine. 
This was related to nokogiri.
A nokogiri update was required for Ruby 3.4 compatibility,
but also they changed some stuff about their precompiled native gems
that was causing bundler not to be able to find the right gem, because
the declared platforms in the lockfile didn't include the necessary ones
to match the docker build environment in Github Actions.
Based on the info in the nokogiri release notes, a newer version of
Bundler is also required to support these changes with the precompiled
native gems, so this pull request also updates Bundler to the current
version.

Based on my testing, these changes should get the docker build working
again under Ruby 3.4.

Reference nokogiri release notes: https://github.com/sparklemotion/nokogiri/releases/tag/v1.18.0

Closes #1411 

CC: @djmb

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ COPY lib/kamal/version.rb /kamal/lib/kamal/version.rb
 # Install system dependencies
 RUN apk add --no-cache build-base git docker openrc openssh-client-default yaml-dev \
     && rc-update add docker boot \
-    && gem install bundler --version=2.4.3 \
+    && gem install bundler --version=2.6.5 \
     && bundle install
 
 # Copy the rest of our application code into the container.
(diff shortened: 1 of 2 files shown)


djmb:
That [fixed it](https://github.com/basecamp/kamal/actions/runs/13564419149/job/37914192045), thank you @smartygus!

smartygus:
@djmb glad I could help!

PR #1425, opened 2025-02-22
feat: ssh: use secret in key_data

idrista:
### What’s changed?  
- `key_data` can now be a **string** to reference a secret, just like `registry.username` and `registry.password`.  
- To avoid breaking changes, the behavior is **reversed** compared to `registry`:  
  - **Registry**.username | **Registry**.password : `Array → lookup secret`, `String → raw value`  
  - **SSH**.key_data: `Array → raw value`, `String → lookup secret`  

### Why?  
- This makes it easier to manage SSH keys securely using `.kamal/secrets`.  
- Avoids exposing sensitive data in Git.


lucashfreitas:
would be very nice!

djmb:
Superseded by #1620

PR #1427, opened 2025-02-23
[Feature] Preserve kamal default options

### LINKED ISSUE #1424: Set default deploy config on kamal init.
# Context

Right now all kamal deployment related files stored in `.kamal` dir, except main `deploy.yml` config, which is stored in `config/deploy.yml` by default.

Some of us want to have all files related to kamal deployment in one dir `.kamal`.
Others also can have different locations for some reason.

We can specify config file via `--config-file=CONFIG_FILE` option on each `kamal` command invocation.
But it feels redundant and exhausting over time, especially with mostly manually invoking aliases-commands.

Taking into account that in most cases there will be one main deploy config, it would be nice to have this options saved on `kamal init` run and used automatically on other `kamal` commands.

# Proposal

Feels like it can be achieved with minimal changes.

For example. There could be generated `.kamal/options.yml` or `.kamal/defaults.yml` file to save general/default options to use with all `kamal` future invocations.
We can read this `config_file` option from mentioned file and apply in the only place it is used (in file `lib/kamal/cli/base.rb`):
```ruby
          commander.configure \
            config_file: Pathname.new(File.expand_path(options[:config_file])),
```

@djmb Let me know your thoughts on this extension proposal. I can take care of implementation if it sounds interesting.

ShPakvel:
## Summary

This PR adds `.kamal/options.yml` file as storage for default options used with kamal commands.

It implements [proposal-issue](https://github.com/basecamp/kamal/issues/1424) to set default deploy config file with different location. To eliminate necessity applying option with each kamal command, like `kamal deploy -c .kamal/deploy.yml`, or with often run manually alias-commands.

Closes #1424

This foundation should be easily applicable for other options. E.g. to allow `destination` default from `.kamal/options.yml` file, change should be as minimal as following:

```diff
-            destination: options[:destination],
+            destination: option_with_default_from_file(:destination),
```

## Changes

- Add check for default options from `.kamal/options.yml` file.
- Add `config_file` option check, with backward-compatible fallback to `config/deploy.yml` value.
- Add `.kamal/options.yml` creation on `kamal init` command. Existing option `-c` can be used during initialization to provide default config file. It will be stored into `.kamal/options.yml`. If not provided, backward-compatible `config/deploy.yml` will be stored.
- Cover `.kamal/options.yml` with tests.

**Note: This PR preserves backward-compatibility with current behavior. This feature is extension without breaking stuff.**

### DIFF
--- a/lib/kamal/cli/base.rb
+++ b/lib/kamal/cli/base.rb
@@ -1,5 +1,6 @@
 require "thor"
 require "kamal/sshkit_with_ext"
+require "active_support/core_ext/hash/keys"
 
 module Kamal::Cli
   class Base < Thor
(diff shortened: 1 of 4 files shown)


grafst:
This is would also be very useful when using a project that has a "config" file like [radicale](https://github.com/Kozea/Radicale) for example.

ShPakvel:
@djmb, Do you have this idea in consideration?

Issue #1448, opened 2025-03-07
kamal proxy - can't boot with a sysctl with a quoted string


mdkent:
After issuing:
```
$ kamal proxy boot_config set -d staging --docker-options='sysctl net.ipv4.ip_local_port_range="10000 60999"'
```
we get the correct setting on disk:
```
$ kamal proxy boot_config get -d staging
<snip>
Host foo-cable-app-01: --publish 80:80 --publish 443:443 --log-opt max-size=10m --sysctl net.ipv4.ip_local_port_range="10000 60999"
```
but we can't boot it:
```
$ kamal proxy reboot --rolling -d staging
<snip>
  INFO [57d18bc1] Running docker run --name kamal-proxy --network kamal --detach --restart unless-stopped --volume kamal-proxy-config:/home/kamal-proxy/.config/kamal-proxy $(cat .kamal/proxy/options || echo "--publish 80:80 --publish 443:443 --log-opt max-size=10m") basecamp/kamal-proxy:v0.8.4 on foo-cable-app-01
Releasing the deploy lock...
  ERROR (SSHKit::Command::Failed): Exception while executing on host foo-cable-app-01: docker exit status: 125
docker stdout: Nothing written
docker stderr: docker: invalid reference format.
See 'docker run --help'.
```
looks like the quoting coming out of the shell expansion is messing it up:
```
$ set -x; docker run --name kamal-proxy --network kamal --detach --restart unless-stopped --volume kamal-proxy-config:/home/kamal-proxy/.config/kamal-proxy $(cat .kamal/proxy/options || echo "--publish 80:80 --publish 443:443 --log-opt max-size=10m") basecamp/kamal-proxy:v0.8.4;
++ cat .kamal/proxy/options
+ docker run --name kamal-proxy --network kamal --detach --restart unless-stopped --volume kamal-proxy-config:/home/kamal-proxy/.config/kamal-proxy --publish 80:80 --publish 443:443 --log-opt max-size=10m --sysctl 'net.ipv4.ip_local_port_range="10000' '60999"' basecamp/kamal-proxy:v0.8.4
docker: invalid reference format.
See 'docker run --help'.
```
note the quoting on
```
--sysctl 'net.ipv4.ip_local_port_range="10000' '60999"'
```
I've tried all manor of quoting here without luck, the only real fix seems to be to full resolve the entire command first, eg:
```
[...]

PR #1450, opened 2025-03-10
Eval proxy options

djmb:
To ensure that the proxy options are evaluated as if they were added verbatim to the command, we'll eval the docker run command.

Fixes: https://github.com/basecamp/kamal/issues/1448


djmb:
Closing in favour of https://github.com/basecamp/kamal/pull/1459

PR #1455, opened 2025-03-13
Refactor using_version method for improved readability

sharmapankaj2512:
This refactor reduces nesting and simplifies the logic of the using_version method by using an early return pattern. The functional behavior remains unchanged while making the code more maintainable.

### DIFF
--- a/lib/kamal/cli/app.rb
+++ b/lib/kamal/cli/app.rb
@@ -314,17 +314,15 @@ class Kamal::Cli::App < Kamal::Cli::Base
   end
 
   private
+
     def using_version(new_version)
-      if new_version
-        begin
-          old_version = KAMAL.config.version
-          KAMAL.config.version = new_version
-          yield new_version
-        ensure
-          KAMAL.config.version = old_version
-        end
-      else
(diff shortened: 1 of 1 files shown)


lib/kamal/cli/app.rb
```diff
@@ -314,17 +314,15 @@ def version
...
-        begin
-          old_version = KAMAL.config.version
-          KAMAL.config.version = new_version
-          yield new_version
-        ensure
-          KAMAL.config.version = old_version
-        end
-      else
-        yield KAMAL.config.version
+      return yield KAMAL.config.version unless new_version
```
djmb, line 319: Generally we avoid the early return pattern in Kamal - there's usually some other way to clean things up which keeps the code more "symmetrical".

In this case we could use the Active Support [with](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/object/with.rb) method and do something like:

```ruby
if new_version
  KAMAL.config.with(version: new_version) { yield KAMAL.config.version }
else
  yield KAMAL.config.version
end
```

PR #1481, opened 2025-04-07
feat: Add shell completions and shell completion command

hschne:
This PR adds [shell completion functions](https://en.wikipedia.org/wiki/Command-line_completion) for ZSH and Bash for Kamal. These are originally hosted in [this repository](https://github.com/hschne/kamal-complete), but I thought getting them in Kamal main made sense. 

Apart from adding the completions this PR adds a command to print the completions. This is akin to what [GitHub CLI](https://cli.github.com/manual/gh_completion) and other tools do. 

## Testing

You can try those completions by adding them to your respective shell. Depends on your local setup. This is roughly what you'd usually do:

```
### bash

First, ensure that you install `bash-completion` using your package manager.

After, add this to your `~/.bash_profile`:

	eval "$(kamal completion -s bash)"

### zsh

Generate a `_kamal` completion script and put it somewhere in your `$fpath`:

	kamal completion -s zsh > /usr/local/share/zsh/site-functions/_gh

Ensure that the following is present in your `~/.zshrc`:

	autoload -U compinit
	compinit -i
```

## Considerations

- I opted to put the completion files into the `completions` directory, which made it necessary to amend the gemspec. Happy to move it to a different directory if that makes more sense. 
- Happy to update the docs after this has been merged :+1: 
- Would love to improve auto completion for aliases, but that would require a command to list aliases by name. Just a nice TODO for the future.

### DIFF
--- a/completions/kamal.bash
+++ b/completions/kamal.bash
@@ -1,0 +1,298 @@
+#!/usr/bin/env bash
+# ------------------------------------------------------------------------------
+# Description
+# -----------
+#
+#  Completion script for Kamal deployment tool (https://kamal-deploy.org/).
+#
+# ------------------------------------------------------------------------------
+
+_init_completion() {
+  COMPREPLY=()
+  cur="${COMP_WORDS[COMP_CWORD]}"
+  prev="${COMP_WORDS[COMP_CWORD - 1]}"
+  words=("${COMP_WORDS[@]}")
+  cword="${COMP_CWORD}"
+  return 0
+}
+
+# Helper function to generate completions for common options
+_kamal_common_options() {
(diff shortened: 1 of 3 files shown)


airblade:
As far as I can tell this hard-codes some accessory names instead of extracting the accessories from deploy.yml?

hschne:
> As far as I can tell this hard-codes some accessory names instead of extracting the accessories from deploy.yml?

That is correct. Ideally, Kamal would expose a command such as `accessory list` that could be used by completion functions, but I thought implementing such a thing without any completions in place. Hence hard-coded common accessories as first step.

PR #1530, opened 2025-04-28
command 'kamal app visit' to visit the app in your browser

grafst:
`kamal app visit`

It would be very cool if we could establish a standard so that all apps deployable with kamal can be opened like this.
Would be very convenient.

I am very new to ruby and also kamal, so I am happy for any pointers.
Also if some people with more complex setups could test this!

Looking forward :)


djmb:
Closing as it duplicates #1070

Issue #1485, opened 2025-04-09
piping into `kamal exec` does not work


prullmann:
When trying to pipe data into commands executed in containers using `kamal exec`, no data is recieved.

### Use Case:

I am trying to restore a database backup via the following command from my dev machine:
`kamal accessory exec -i --reuse mysql 'mysql example_db"' < example_db_backup.sql`
_(ignoring password handling for this example)_

Kamal translates this to:
`ssh -t example_host docker exec -it example_project-mysql 'mysql example_db' < example_db_backup.sql`

However, to make piping through docker work, `-i` has to be set, but not `-t`.

### Proposed Solution:

I suggest that besides the `--interactive`/`-i` flag, a `--tty`/`-t` flag is added to control both independently.

*Alternative Solution:*

Kamal could also check if STDIN is attached to a file/pipe when interactive is enabled and set `-i` instead of `-ti` in this case.

If one of those solutions sound acceptable, I am happy to write a PR.

djmb:
Was fixed by https://github.com/basecamp/kamal/pull/1544

PR #1544, opened 2025-05-08
Allow piping into kamal exec #1485 

prullmann:
This PR enables data to be piped to remote commands via STDIN.

An example use case would be to restore a database backup from your local dev machine:
`kamal accessory exec -i --reuse mysql 'mysql example_db"' < example_db_backup.sql`

As discussed in #1485, the implementation for interactive mode now checks if STDIN is connected to a TTY and only then it uses the current '-it' options for docker, else it uses '-i', which enables the transfer of data via STDIN.

The implementation should be straight forward, just the tests were a bit tricky.
I made sure the tests also still work when not run from a tty (eg within CI).
I hope it's OK like this, any comments are welcome.

### DIFF
--- a/lib/kamal/commands/accessory.rb
+++ b/lib/kamal/commands/accessory.rb
@@ -55,14 +55,14 @@ class Kamal::Commands::Accessory < Kamal::Commands::Base
 
   def execute_in_existing_container(*command, interactive: false)
     docker :exec,
-      ("-it" if interactive),
+      (docker_interactive_args if interactive),
       service_name,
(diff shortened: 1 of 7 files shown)


prullmann:
Hi @djmb, Do you have any comments? Is there anything I can do to make this easier for you to review?

fdev:
This is a very welcome change. It enables me to perform all common management tasks through kamal aliases, instead of just a subset. Hopefully it gets merged soon.

PR #1547, opened 2025-05-09
Pin accessories to tags

dhh:
Allow accessories to specify which hosts they'll run on by using tags.

Example:

```yaml
servers:
  web:
    hosts:
      - fizzy-app-101: writer
      - fizzy-app-102: reader

accessories:
  beamer-primary:
    <<: *beamer-accessory
    service: beamer-primary
    cmd: beamer primary --remove-after=1h --dir=/storage
    tag: writer

  beamer-replica:
    <<: *beamer-accessory
    service: beamer-replica
    cmd: beamer replica --primary=http://fizzy-app-101:5000/ --dir=/storage
    tag: reader
```


dhh:
cc @kevinmcconnell

dhh:
Added that you can either specify a singular `tag` or an array of `tags`, like `host` and `hosts`. We should follow up to do the same for `roles` by adding a singular `role`.

PR #1531, opened 2025-04-28
Add support for custom certificates

acidtib:
This PR adds support for custom SSL certificates in the proxy configuration, building upon the foundation laid by @kpumuk in PR #969. Users can now securely provide their own SSL certificates and private keys through secrets using the following configuration format:

```yaml
proxy:
  ssl:
    certificate_pem: CERTIFICATE_PEM
    private_key_pem: PRIVATE_KEY_PEM
```

### Background
This enhancement addresses scenarios where Let's Encrypt automated certificate management isn't feasible or when users already possess SSL certificates from other Certificate Authorities. The implementation maintains compatibility with existing SSL options:

 - Enable SSL with Let's Encrypt (unchanged)
```yaml
proxy:
  ssl: true
```

- Disable SSL (unchanged)
```yaml
proxy:
  ssl: false
```

- Enable custom SSL with provided certificates
```yaml
proxy:
  ssl:
    certificate_pem: CERTIFICATE_PEM
    private_key_pem: PRIVATE_KEY_PEM
```

The feature allows for a more flexible SSL setup while maintaining security best practices by using secrets for certificate management.

Site documentation PR: https://github.com/basecamp/kamal-site/pull/174

PR #1583, opened 2025-06-17
Custom certs per role

djmb:
- Upload the cert with `sshkit.upload!`
- Use the role name to create a directory for each role's certs
- Add an integration test for the custom certs


djmb:
Follow up to https://github.com/basecamp/kamal/pull/1531

PR #1585, opened 2025-06-18
Update custom cert docs

djmb:
Remove ssl: ..., reword some of the docs and ensure the formatting is correct.

### DIFF
--- a/lib/kamal/configuration/docs/proxy.yml
+++ b/lib/kamal/configuration/docs/proxy.yml
@@ -45,27 +45,23 @@ proxy:
   # unless you explicitly set `forward_headers: true`
   #
   # Defaults to `false`:
-  ssl: ...
+  ssl: true
 
   # Custom SSL certificate
   #
   # In some cases, using Let's Encrypt for automatic certificate management is not an
-  # option, for example if you are running from host than one host. Or you may already
-  # have SSL certificates issued by a different Certificate Authority (CA).
-  # Kamal supports loading custom SSL certificates
-  # directly from secrets.
+  # option, for example if you are running from host than one host.
   #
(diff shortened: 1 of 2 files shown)


mhenrixon:
In #1531 and #1583, I couldn't find any support for fallback or handling of multiple domains, which may require certificates for some domains and use Let's Encrypt for others. 

Did I read that right? Will the certificates provided be used for all hosts?

djmb:
> Did I read that right? Will the certificates provided be used for all hosts?

Yes they will. Doing anything else would require updating kamal-proxy to allow per domain TLS configuration (cc @kevinmcconnell).

djmb:
@mhenrixon - one option is to have separate roles for each domain if that is feasible.

Each one can then have it's own container and configuration and a single kamal-proxy will happily handle both on the same host.

lib/kamal/configuration/docs/proxy.yml
```diff
@@ -45,27 +45,23 @@ proxy:
...
-  # Kamal supports loading custom SSL certificates
-  # directly from secrets.
-  #
-  # Examples:
-  #   ssl: true              # Enable SSL with Let's Encrypt
-  #   ssl: false             # Disable SSL
-  #   ssl:                   # Enable custom SSL
-  #     certificate_pem: CERTIFICATE_PEM
-  #     private_key_pem: PRIVATE_KEY_PEM
+  # option, for example if you are running from host than one host.
```
frenkel, line 53: "from host than one host" doesn't seem right.

djmb: Thanks! Should say "from more than one host"

PR #1594, opened 2025-06-24
Deploy env is authoritative (1-9-stable)

jeremy:
Otherwise, having a stray `AWS_ACCESS_KEY_ID` for your dev env can leak into your production deploys. The given deploy env should always be authoritative.

Only affects Kamal 1.x due to Dotenv usage.


djmb:
Backported https://github.com/basecamp/kamal/pull/1494 to fix the tests

PR #1599, opened 2025-06-27
Enable cache-from support for GHA cache options

martinzamuner:
Previously, cache options were only passed to `cache-to`, not `cache-from`. This PR updates the builder to pass allowed options to `cache-from` as well. Specifically:
* `url`
* `url_v2`
* `token`
* `scope`
* `timeout`

Reference: https://docs.docker.com/build/cache/backends/gha/#synopsis

### DIFF
--- a/lib/kamal/configuration/builder.rb
+++ b/lib/kamal/configuration/builder.rb
@@ -177,8 +177,15 @@ class Kamal::Configuration::Builder
       [ server, cache_image ].compact.join("/")
     end
 
+    def cache_options
+      builder_config["cache"]&.fetch("options", nil)
+    end
+
     def cache_from_config_for_gha
-      "type=gha"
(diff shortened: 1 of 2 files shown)


djmb:
Can we assume that you would always want to pass the same values to both cache-to and cache-from? 

Maybe it would be better to have something like:

```
options:
  cache-to:
    ...
  cache-from:
    ...
```

We could keep passing options not under cache-to/cache-from to cache-to for backward compatibility.

martinzamuner:
I think it's pretty safe to assume that.

`url`, `url_2` & `token` are used for auth with the cache server, which needs to be the same for read/write.
`scope` is basically the key for such read/write, can't think of a good reason to use different keys.
`timeout` could maybe benefit from different values, but a priori seems overkill.

I might be missing something, though. Happy to apply your suggestion if you prefer.

PR #1601, opened 2025-07-01
Expand secret environment variables in accessory files

aliismayilov:
When accessory file ends with `.erb`, it's possible to interpolate environment variables in it. Currently, only clear environment variables are expanded. Not sure if it was a deliberate choice.

This PR includes all env variables (including the aliased ones) in the ERB template processing.

Background: I was trying to run PgBouncer (a PostgreSQL connection pooler) as an accessory, which required the database connection string to be present in its configuration file. I managed to work around this by generating the final config file through a Docker entrypoint script. However, if this PR is accepted, the setup would become much simpler.

### DIFF
--- a/lib/kamal/configuration/accessory.rb
+++ b/lib/kamal/configuration/accessory.rb
@@ -142,17 +142,17 @@ class Kamal::Configuration::Accessory
 
     def expand_local_file(local_file)
       if local_file.end_with?("erb")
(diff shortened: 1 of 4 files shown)


djmb:
Thank's @aliismayilov!

Happy to merge this one since https://github.com/basecamp/kamal/pull/1724 now allows you to lock down permissions on those files.

PR #1606, opened 2025-07-11
Add support for Bitwarden Secret Manager EU vaults

frenkel:
Currently it is not possible to override the Bitwarden Security Manager server url on a case by case basis. You can only do this through the global config. This PR fixes that by adding a new `--server-url=` argument to the kamal secrets command.

It has touched quite some files because these functions had all the optional arguments named. I've replaced the ones that were not used with `**` so any future additions can have a smaller diff.

Fixes https://github.com/basecamp/kamal/discussions/1577.

I'll open a PR for the documentation repo in a few minutes.

### DIFF
--- a/lib/kamal/cli/secrets.rb
+++ b/lib/kamal/cli/secrets.rb
@@ -4,6 +4,7 @@ class Kamal::Cli::Secrets < Kamal::Cli::Base
   option :account, type: :string, required: false, desc: "The account identifier or username"
   option :from, type: :string, required: false, desc: "A vault or folder to fetch the secrets from"
   option :inline, type: :boolean, required: false, hidden: true
+  option :server_url, type: :string, aliases: "-u", required: false, desc: "Override the default server-url"
   def fetch(*secrets)
     adapter = initialize_adapter(options[:adapter])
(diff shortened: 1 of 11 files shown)


dawidof:
no need you can do in secrets:


```
SECRETS_RAW=$(kamal secrets fetch --adapter bitwarden-sm <PROJECT_ID>/all)

DATABASE_URL=$(kamal secrets extract DATABASE_URL ${SECRETS_RAW})
SECRET_KEY_BASE=$(kamal secrets extract SECRET_KEY_BASE ${SECRETS_RAW})
```

dawidof:
> no need you can do in secrets:
> 
> ```
> SECRETS_RAW=$(kamal secrets fetch --adapter bitwarden-sm <PROJECT_ID>/all)
> 
> DATABASE_URL=$(kamal secrets extract DATABASE_URL ${SECRETS_RAW})
> SECRET_KEY_BASE=$(kamal secrets extract SECRET_KEY_BASE ${SECRETS_RAW})
> ```

Actually i was wrong, you cannot do this way. It will be good to have eu region

PR #1609, opened 2025-07-16
SSL private key should be locked down just in case

frenkel:


frenkel:
Apparently the tests fail in CI. I had tested this manually on a server, where it worked without problems. Any idea why it would fail? The user that owns the file is the same one as the one that execute the proxy process.

frenkel:
I understand now. The uploads are done as the ssh user, but then the container is started with the directory as volume and the user in the container cannot always access them. In my testing this coincidentally was uid 1000 for both, which made it work.

I think this cannot be fixed when using the simple `upload!` method, as only when your ssh user is root you are allowed to change the owner uid. Any ideas on how to best solve this @djmb? If you can give some hint at what you prefer, I can update this PR with it.

PR #1624, opened 2025-08-11
Add example of rails credentials:fetch to secrets template

MatheusRich:
This was [recently introduced in Rails][1] and it's worth teaching this new concept.

[1]: https://github.com/rails/rails/pull/53119


supercobra:
More documentation is always good.

PR #1629, opened 2025-08-17
Add plural path prefixes

olivermt:
Hello,

This builds on the work from @ACPK to add path prefix plurals.

This has been used in production now for about two months without any hiccups and I have found no bugs or errors.


olivermt:
@djmb any chance to look at this? :) I would like to stop running prod off a fork now after soon three months hehe

PR #1631, opened 2025-08-18
Preserve file extension in destination config files

### LINKED ISSUE #1630: Destination config files don't preserve base file extension
When using custom config files with the `--config-file` option combined with the `-d` (destination) flag, Kamal hardcodes the `.yml` extension for destination-specific config files, regardless of the base config file's extension.

## Current Behavior

If you use a config file with `.yaml` extension:
```bash
kamal deploy --config-file my-app.yaml -d staging
```

Kamal will:
- Load `my-app.yaml` ✅ (works correctly)
- Look for `my-app.staging.yml` ❌ (hardcoded .yml extension)

This inconsistency means users must use different extensions for their base and destination files when using `.yaml`.

## Expected Behavior

The destination config file should preserve the same extension as the base config file:
- Base file: `my-app.yaml` → Destination file: `my-app.staging.yaml`
- Base file: `my-app.yml` → Destination file: `my-app.staging.yml`

## Root Cause

In `lib/kamal/configuration.rb:42`, the `destination_config_file` method hardcodes the `.yml` extension:

```ruby
def destination_config_file(base_config_file, destination)
  base_config_file.sub_ext(".#{destination}.yml") if destination  # ❌ hardcoded .yml
end
```

delkopiso:
Fixes #1630

## Summary

This PR fixes an issue where destination config files always used the `.yml` extension, even when the base config file used `.yaml`. Now the destination file preserves the same extension as the base config file.

### Core Fix
- **`lib/kamal/configuration.rb`**: Modified `destination_config_file` method to preserve the base file's extension instead of hardcoding `.yml`

### Test Coverage
- **`test/configuration_test.rb`**: Added tests to verify both `.yml` and `.yaml` extensions work correctly with destinations
- **Test fixtures**: Added `.yaml` versions of existing test files (`deploy_for_dest.yaml`, `deploy_for_dest.world.yaml`)

### Before
```bash
kamal deploy --config-file my-app.yaml -d staging
# Looked for: my-app.staging.yml ❌
```

### After  
```bash
kamal deploy --config-file my-app.yaml -d staging
# Looks for: my-app.staging.yaml ✅

kamal deploy --config-file my-app.yml -d staging  
# Looks for: my-app.staging.yml ✅ (unchanged)
```

## Backward Compatibility

✅ **Fully backward compatible** - existing `.yml` config files continue to work exactly as before. Only affects users who want to use `.yaml` extensions consistently.

## Test Plan

- [x] Existing tests pass (ensures no regression)
- [x] New tests verify `.yaml` extension preservation
- [x] New tests verify `.yml` extension still works (backward compatibility)

The fix is minimal, focused, and maintains full backward compatibility while providing the expected behavior for users who prefer `.yaml` extensions.

### DIFF
--- a/lib/kamal/configuration.rb
+++ b/lib/kamal/configuration.rb
@@ -39,7 +39,10 @@ class Kamal::Configuration
       end
 
       def destination_config_file(base_config_file, destination)
-        base_config_file.sub_ext(".#{destination}.yml") if destination
+        if destination
+          base_extension = base_config_file.extname
+          base_config_file.sub_ext(".#{destination}#{base_extension}")
+        end
       end
   end
 
--- a/test/configuration_test.rb
+++ b/test/configuration_test.rb
@@ -262,6 +262,25 @@ class ConfigurationTest < ActiveSupport::TestCase
     end
   end
 
+  test "destination yaml config preserves extension" do
+    dest_config_file = Pathname.new(File.expand_path("fixtures/deploy_for_dest.yaml", __dir__))
+
+    config = Kamal::Configuration.create_from config_file: dest_config_file, destination: "world"
+    assert_equal "2.1.1.1", config.all_hosts.first
+  end
+
+  test "destination file preserves base file extension" do
+    # Test with .yml extension (existing behavior should still work)
+    yml_config_file = Pathname.new(File.expand_path("fixtures/deploy_for_dest.yml", __dir__))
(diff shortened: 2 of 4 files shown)


djmb:
Thanks for the PR @delkopiso! 

This is not a backward compatible change though, so I think we'll need to wait for Kamal 3.

PR #1664, opened 2025-10-06
Don't login to registry locally for remote build: take 2

djmb:
Reverts basecamp/kamal#1662

Then set BUILDKIT_NO_CLIENT_TOKEN=1 in the env for the remote build push command to allow the remote builder to push without a local login.

### DIFF
--- a/lib/kamal/cli/build.rb
+++ b/lib/kamal/cli/build.rb
@@ -19,7 +19,7 @@ class Kamal::Cli::Build < Kamal::Cli::Base
     pre_connect_if_required
 
     ensure_docker_installed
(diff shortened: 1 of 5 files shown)


djmb:
@mhenrixon - if you had a chance, I'd appreciate if you could check if this works for your remote builds?

mhenrixon:
> @mhenrixon - if you had a chance, I'd appreciate if you could check if this works for your remote builds?

I'll give it a go tonight. Will get back to you latest tomorrow morning.

mhenrixon (approved):
Image

PR #1669, opened 2025-10-14
Set minimum proxy version to 0.9.0

julienanne:


julienanne:
Sorry bad branche name

PR #1670, opened 2025-10-14
Set minimum proxy version to 0.9.1

julienanne:
Inspired by https://github.com/basecamp/kamal/pull/1546


morgoth:
I believe this is only bumped when there are some incompatibilities. 0.9.1 looks like a small bugfix release, so there is no need to force it in kamal.

julienanne:
Ok fine I just make it because, I want to update my Kamal proxy by following the documentation here https://kamal-deploy.org/docs/commands/proxy/ : "When you want to upgrade kamal-proxy, you can call kamal proxy reboot..." and see that my kamal proxy version is not the last kamal proxy released version, now I know why and keep it like this for the moment.

Thanks.

Issue #1674, opened 2025-10-20
Cannot use Kamal local docker registry on macOS when deploying a Rails app


jeromedalbert:
After https://github.com/basecamp/kamal/pull/1355 got merged and released, I tried deploying a Rails app using the local docker registry with Kamal configured the same way as this pending Rails PR from DHH: https://github.com/rails/rails/pull/55934. I am on macOS. But when I run `bin/kamal setup`, I eventually get the following error:

```
ERROR (SSHKit::Command::Failed): Exception while executing on host 5.78.64.88: docker exit status: 1
docker stdout: Nothing written
docker stderr: Error response from daemon: Get "http://localhost:5555/v2/": dial tcp [::1]:5555: connect: connection refused
```

### Steps to repro

Clone the following repro Rails app, and follow its README instructions: https://github.com/jeromedalbert/rails-55934. 

### Additional information

I am on macOS, deploying a Rails 8.1 RC1 app, with the latest Docker Desktop 4.48.0, and the latest Kamal 2.8.0.

brunoprietog:
I can confirm from a primary source that it works on Mac. It's probably something else that's not working on your end.

adrienpoly:
I tried also to use it a I get a similar error 

here is a public build if this can help https://github.com/rubyevents/rubyevents/actions/runs/18659102655/job/53195241858

In my can the kamal server was already setup with an external Docker registry. Not sure if in such cases I would need to anything specific. I did try to run kamal setup but I had the same error.

Looking at the logs it seems we can push the image but we are not able to pull it ?

djmb:
I think this looks like a timing issue - we set up port forwarding so the app servers can access the registry as localhost. It should confirm the ports are all up before doing the docker pull. 

I'm working on a fix now

djmb:
@jeromedalbert - I've released Kamal 2.8.1 with that fix.

adrienpoly:
Thanks I have tested 2.8.1.

I can now deploy from my [CI task](https://github.com/rubyevents/rubyevents/actions/runs/18674015904/job/53240624900) but I am still getting a similar error when I try to deploy from my Mac 

```
#24 pushing layers done
[...]

PR #1679, opened 2025-10-21
Add local registry support for remote builders

djmb:
When running with a remote builder and a local registry:
1. Forward the local registry port to the remote builder
2. Use host networking on the remote builder
3. Use a different builder name as the settings are different

A remote builder can only have one port at a time forwarded, so concurrent builds on the same registry port will fail.

See: https://github.com/basecamp/kamal/issues/1674#issuecomment-3424910818

cc @jeromedalbert, @adrienpoly

### DIFF
--- a/lib/kamal/cli/build.rb
+++ b/lib/kamal/cli/build.rb
@@ -70,7 +70,7 @@ class Kamal::Cli::Build < Kamal::Cli::Base
   def pull
     login_to_registry_remotely unless KAMAL.registry.local?
 
-    forward_local_registry_port(KAMAL.hosts) do
+    forward_local_registry_port(KAMAL.hosts, **KAMAL.config.ssh.options) do
       if (first_hosts = mirror_hosts).any?
         #  Pull on a single host per mirror first to seed them
         say "Pulling image on #{first_hosts.join(", ")} to seed the #{"mirror".pluralize(first_hosts.count)}...", :magenta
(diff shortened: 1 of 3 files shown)


darkamenosa:
Is this released in 2.8.1. I got the same issue, and here is my config: 
```
# Credentials for your image host.
registry:
  server: "localhost:5555"

# Configure the image builder.
builder:
  # Build for the server's native architecture (aarch64 == arm64)
  arch: arm64
  # Enable a registry cache so apt/bundler layers can be reused across deploys.
  cache:
    type: registry
    # Use an arch-specific cache image to avoid mixing caches across platforms
    image: tuyenhx/async_example-buildcache-arm64
    options: mode=max,image-manifest=true,oci-mediatypes=true

  remote: ssh://root@46.224.21.86
  local: false
```

adrienpoly:
thanks @djmb I tested this branch 

It now works when I deploy from my mac but it fails when I [deploy from the CI](https://github.com/rubyevents/rubyevents/actions/runs/18713355952/job/53367277072)

To add more context 
Mac: arm64
Host : amd64
CI: amd64

I tried both with local: true and false and the results where the same

Here is the draft PR https://github.com/rubyevents/rubyevents/pull/1096

and the kamal config https://github.com/rubyevents/rubyevents/blob/main/config/deploy.staging.yml

let me know if you want me to run any more tests

djmb:
@adrienpoly - I only see remote builds failing in the action tab. I think the "connection reset by peer" error probably indicates that port forwarding was initially set up correctly, so possibly the SSH connection is being dropped?

I've added the keepalive configuration, hopefully that might fix it, but I've also added some logging for any errors in the port forwarding thread, if anything goes wrong there right now I think we are missing it.

Could you try building with this branch again when you have a chance?

Issue #1675, opened 2025-10-20
Kamal 2.8.0 doesn't know "registry login"


JamesChevalier:
When I run `kamal registry login` with Kamal 2.8.0 I get this error:

> Could not find command "login".

When I run `kamal registry help` it doesn't include login as an option:

```
Commands:
  kamal registry help [COMMAND]  # Describe subcommands or one specific subcommand
  kamal registry remove          # Remove local registry or log out of remote registry locally and remotely
  kamal registry setup           # Setup local registry or log in to remote registry locally and remotely

Options:
  -v, [--verbose], [--no-verbose], [--skip-verbose]  # Detailed logging
  -q, [--quiet], [--no-quiet], [--skip-quiet]        # Minimal logging
      [--version=VERSION]                            # Run commands against a specific app version
  -p, [--primary], [--no-primary], [--skip-primary]  # Run commands only on primary host instead of all
  -h, [--hosts=HOSTS]                                # Run commands on these hosts instead of all (separate by comma, supports wildcards with *)
  -r, [--roles=ROLES]                                # Run commands on these roles instead of all (separate by comma, supports wildcards with *)
  -c, [--config-file=CONFIG_FILE]                    # Path to config file
                                                     # Default: config/deploy.yml
  -d, [--destination=DESTINATION]                    # Specify destination to be used for config file (staging -> deploy.staging.yml)
  -H, [--skip-hooks]                                 # Don't run hooks
                                                     # Default: false
```

The command is still documented in https://kamal-deploy.org/docs/commands/registry/

djmb:
Ah sorry, those should not have been removed. I'll add them back as aliases for setup/remove.

PR #1680, opened 2025-10-21
Restore kamal registry login/logout commands

djmb:
They were removed for the local registry in favour of the more generic setup/remove.

Keep the commands for backward compatibility. They can be removed the next major version.

Fixes: https://github.com/basecamp/kamal/issues/1675

### DIFF
--- a/lib/kamal/cli/registry.rb
+++ b/lib/kamal/cli/registry.rb
@@ -24,4 +24,26 @@ class Kamal::Cli::Registry < Kamal::Cli::Base
       on(KAMAL.hosts) { execute *KAMAL.registry.logout } unless options[:skip_remote]
     end
   end
+
+  desc "login", "Log in to remote registry locally and remotely"
+  option :skip_local, aliases: "-L", type: :boolean, default: false, desc: "Skip local login"
+  option :skip_remote, aliases: "-R", type: :boolean, default: false, desc: "Skip remote login"
+  def login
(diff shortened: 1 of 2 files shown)


JamesChevalier:
Apologies if I'm misunderstanding or linked to the wrong code below. Is there a plan to remove `kamal registry login` from future versions?
The setup [docs](https://kamal-deploy.org/docs/commands/setup/) and [code](https://github.com/basecamp/kamal/blob/bdf38a4af0e8aa2bbf603b62ebc6e0131514d4cd/lib/kamal/commands/registry.rb#L19-L26) seem to cover a much different/wider set of actions compared to the simpler/focused registry login [docs](https://kamal-deploy.org/docs/commands/registry/) and [code](https://github.com/basecamp/kamal/blob/bdf38a4af0e8aa2bbf603b62ebc6e0131514d4cd/lib/kamal/commands/registry.rb#L2-L11).
If I'm [Using AWS ECR as the container registry](https://kamal-deploy.org/docs/configuration/docker-registry/#using-aws-ecr-as-the-container-registry), I currently have to run `kamal registry login` prior to running `kamal build deliver` and then `kamal deploy`. It seems like `kamal setup` would do too much in that case.

djmb:
@JamesChevalier - sorry, I wasn't clear there - by setup/remove I mean `kamal registry setup`, `kamal registry remove`.

Issue #1646, opened 2025-09-08
`kamal app version` should not need registry login


samialdury:
basically what the title says - when I run `kamal app version` and I'm not logged in to the registry, kamal throws  `ERROR (Kamal::ConfigurationError): registry/password: is required`. 

Since I run Kamal in Github Actions, skipping the registry login for commands that do not need it saves a lot of time.

mike-weiner:
@samialdury - I agree that it feels unnecessary to require image registry credentials for a command that doesn't require them. With that being said, your ask really feels like a complete change to how `kamal` currently behaves. It doesn't know what commands you want to run, so it does all of the up-front validation of your configuration file first. It's trying to protect you from misconfiguration issues.

~~Logging into a registry shouldn't be slowing CI builds down _that much_, nor should you have to sign in manually. You should be pointing your `kamal` config to where it can find your credentials (https://kamal-deploy.org/docs/configuration/docker-registry/) and it'll log you in.~~ In fact, `kamal` doesn't even log into the registry.

What is your use case where you can't (or won't) just include your image registry details in the configuration file that you're already having to supply?

mike-weiner:
@samialdury - Hey! I've been thinking about this one for a while, and I've changed my position. I think there are a couple of valid use cases for wanting this - especially for scenarios where you want folks that don't have access to a protected image registry to be able to use `kamal` to simply get information about the apps running.

I went ahead and opened a PR with a change to allow this: https://github.com/basecamp/kamal/pull/1685

PR #1685, opened 2025-10-25
Introduce new option to bypass validation of registry configuration

mike-weiner:
Closes https://github.com/basecamp/kamal/issues/1646. 

This PR adds a new field to the `registry` section of the configuration file to bypass all validation of the `registry` config.

The primary use case here is to be able to run commands like `kamal app containers` or `kamal app version` without having to specify image registry credentials in the configuration file.

This new bypass is opt-in _only_. By default, users will get the safeguard of the validation. `skip_validation: false` is **not** an option. You either include `skip_validation: true` or you leave it out.

For those that want to potentially shoot themselves in the foot by being able to bypass the `registry` config validation, they can now do something like:

```
registry:
  skip_validation: true
```

### DIFF
--- a/test/configuration_test.rb
+++ b/test/configuration_test.rb
@@ -27,7 +27,7 @@ class ConfigurationTest < ActiveSupport::TestCase
     ENV.delete("VERSION")
   end
 
-    %i[ service image registry ].each do |key|
+  %i[ service image registry ].each do |key|
     test "#{key} config required" do
       assert_raise(Kamal::ConfigurationError) do
         Kamal::Configuration.new @deploy.tap { |config| config.delete key }


djmb:
I'm not clear on why we would need this. You can use secrets to specify the values in the config (`DOCKER_REGISTRY_TOKEN` etc).

Those secrets should only be read when attempting to login to the registry, so if you want to run commands that don't login to the registry like `kamal app containers`  then that should still work, even if you don't have access to them. If it doesn't work like that right now, we should fix that rather than add an option to skip the validations.

mike-weiner:
@djmb - My PR description was confusing. Let me try to clear things up. 

If I wanted to run `kamal app version` (which shouldn't require any registry login), and my configuration file doesn't specify `registry` at all, I get this error: 

```
weiner kenyonwx (main) % kamal app version
  ERROR (Kamal::ConfigurationError): registry/username: is required
```

If I then update my configuration file to list: 
```
registry:
  username:
    - ""
  password:
    - ""
```

I can then successfully run `kamal app containers`. I don't want to speak for the author of https://github.com/basecamp/kamal/issues/1646, but I think that is the heart of the issue. 

I'm not a fan of including blank values. With this PR I suggested being able to do something like:

```
registry:
  skip_validation: true
```

That would get around the `ERROR (Kamal::ConfigurationError): registry/username: is required`.

Bottom line is that I don't feel like `kamal` should require registry configuration to be present for commands that don't invoke registry operations. With the way `kamal` is currently structured, it does all validation up-front before any command logic is run.

What are your thoughts?

djmb:
I'm not sold on this - it means that you have two different configuration files, one with the registry details and the other with `skip_validation: true`.

It seems simpler to have one configuration file and inject the secrets into it at runtime. Kamal lazy loads secrets, so you don't need access to them for commands that don't use them.

mike-weiner:
@djmb - 

What is the expected way to handle the situation where I don't want folks to have access to a container registry but want them to be able to use Kamal to view app logs? 

Just put in empty strings?

djmb:
Hi @mike-weiner!

It would be to have something like:

```
# config/deploy.yml
registry:
  username: <name>
  password:
    - KAMAL_REGISTRY_PASSWORD
```

```
# .kamal/secrets-common
KAMAL_REGISTRY_PASSWORD=$(command-to-fetch-password)
```

Then only kamal commands that need the password should call `command-to-fetch-password`. `kamal app version` shouldn't do that so it should work ok without access.

mike-weiner:
Hi, @djmb!

Okay, I'm fine with that solution and closing this PR.

PR #1697, opened 2025-11-04
Security Fix: Prevent OS Command Injection in Passbolt adapter (CWE-78)

NiceGuy720:
Hi team,

This PR fixes a critical OS Command Injection vulnerability in the `Passbolt` adapter.

The `item["id"]` variable was interpolated directly into a shell command without escaping. An attacker who can control folder metadata in Passbolt could escalate this to RCE on the machine running `kamal secrets fetch`.

This patch adds the missing `.to_s.shellescape` to sanitize the input, fixing the vulnerability.


NiceGuy720:
You can keep it but I would suggest you make it not vurenable

Donal McBreen ***@***.***> schrieb am Fr. 7. Nov. 2025 um
15:22:

Issue #1200, opened 2024-11-07
Secrets with special chars rise error unexpected token at


vldmr-k:
Hi All
I'm using kamal version 2.3.0.
When I try to extract secret from bitwarden

```
kamal secrets print
```
Kamal throw error
```
  ERROR (JSON::ParserError): unexpected token at '{"SENTRY/EMAIL_URL":"*****:~ArhTd7+3D8\'
```

Full secret is 
```
smtp://*****:~ArhTd7+3D8)@mail.***.***:25
```

My `.kamal/secrets`
```
SECRETS=$(kamal secrets fetch --adapter bitwarden --account **** SENTRY/EMAIL_URL)
```

3v0k4:
I cannot reproduce.

Could you share the output of `bw get item SENTRY`?

vldmr-k:
> bw get item SENTRY

Yes, sure. But I can't share sesative data

```json
{
   "passwordHistory":[
      {
         "lastUsedDate":"2024-12-15T23:13:18.434Z",
         "password":"smtp://sentry@*****.com:~ArhTd7+3D8)@mail.***.***:25"
      }
   ],
   "revisionDate":"2024-12-15T23:13:48.736Z",
   "creationDate":"2024-11-07T01:34:43.958Z",
   "deletedDate":null,
   "object":"item",
   "id":"d0506d38-4f0b-4f58-8d5f-b551d073884e",
   "organizationId":"afde1b41-7117-446b-a3bf-d0859f123dd1",
   "folderId":null,
   "type":1,
   "reprompt":0,
   "name":"SENTRY",
   "notes":"sentry.etradeua.com",
   "favorite":false,
   "fields":[
      {
         "name":"EMAIL_URL",
         "value":"smtp://sentry@*****.com:~ArhTd7+3D8)@mail.***.***:25",
         "type":1,
         "linkedId":null
      }
   ],
   "login":{
      "fido2Credentials":[
         
      ],
      "uris":[
         
      ],
      "username":null,
      "password":null,
      "totp":null,
      "passwordRevisionDate":"2024-12-15T23:13:18.434Z"
   },
   "collectionIds":[
      "0da0fb66-7385-4b28-ad8e-22020dc0c321"
   ]
}
```


all `*` is sensative data contains `[a-z]`

jhclouse:
Seeing the same issue with 1password. I'm getting around it by querying 1password-cli directly for now as recommended in #1007.

PR #1346, opened 2025-01-08
Handle parentheses in variables in commands

i7an:
Variables with parentheses are not handled correctly in commands:

```
SECRETS=\\{\\\"PASSWORD\\\":\\\"pa\\(\\(word\\\"\\}
PASSWORD=$(kamal secrets extract PASSWORD ${SECRETS})
```

Expected `kamal secrets print` output:

```
SECRETS=\{\"PASSWORD\":\"pa\(\(word\"\}
PASSWORD=pa((word
```

Actual `kamal secrets print` output:

```
SECRETS=\{\"PASSWORD\":\"pa\(\(word\"\}
PASSWORD=$(kamal secrets extract PASSWORD \{\"PASSWORD\":\"pa\(\(word\"\})
```

This PR depends on another one https://github.com/bkeepers/dotenv/pull/526. The dependency was approved https://github.com/bkeepers/dotenv/pull/526#issuecomment-2787755200 but not merged so as not to break `kamal`.

**The plan**
* Merge and release this patch. It is backward compatible and will work with patched and with unpatched `dotenv`
* Merge and release the `dotnenv` [patch](https://github.com/bkeepers/dotenv/pull/526)

PR #1726, opened 2025-12-01
Handle parentheses in secrets

djmb:
Dotenv inline command substitutions incorrectly matches escaped parentheses when parsing commands, so let's replace its regex with one that doesn't.

This means that when you have a secrets file like:

```
SECRETS=$(kamal secrets fetch --adapter myadapter SECRET_WITH_PARENS)
SECRET_WITH_PARENS=$(kamal secrets extract SECRET_WITH_PARENS ${SECRETS})
```

Then we correctly add the secret even if it contains parentheses.

Currently if SECRET_WITH_PARENS=pass)word, then we get:

```
SECRET_WITH_PARENS=$(kamal secrets extract SECRET_WITH_PARENS pass\)word)
```

And dotenv's regex thinks the command is
`kamal secrets extract SECRET_WITH_PARENS pass\`.

Fixes: #1200


i7an:
Related PRs:
* https://github.com/basecamp/kamal/pull/1346
* https://github.com/bkeepers/dotenv/pull/540
  * https://github.com/bkeepers/dotenv/pull/526

PR #1750, opened 2026-01-10
Add healthcheck port configuration option

lucasmeijer:
Allow healthchecks to use a different port than the main application port by adding a configurable `port` option under `proxy.healthcheck.port`.

When not specified, the healthcheck port is not set and kamal-proxy will use its default behavior. The port can be set to any valid port number (1-65535) and is passed to kamal-proxy via the --health-check-port parameter.

This is useful when applications expose health endpoints on a dedicated port separate from the main application port.


shiny:
I'm also looking for `proxy.healthcheck.port` for my garage server https://github.com/basecamp/kamal-proxy/discussions/177

Issue #1752, opened 2026-01-14
Skip hooks doesnt work for server command


zdehasek:
When I run `kamal server --skip-hooks` it does not work 

```
…/kamal/zenbpm master kamal server exec docker exec kamal-proxy kamal-proxy list --skip-hooks
Running the pre-connect hook...
  INFO [33147b5c] Running /usr/bin/env .kamal/hooks/pre-connect as zde@localhost
  ERROR (Kamal::Cli::HookError): Hook `pre-connect` failed:
.kamal/hooks/pre-connect exit status: 256
.kamal/hooks/pre-connect stdout: [tag] No new tag, skipping deploy (current: v0.3.9)
.kamal/hooks/pre-connect stderr: Nothing written
```

other commands like `kamal app logs -f`

zdehasek:
I already create a fix for it https://github.com/basecamp/kamal/pull/1754

zdehasek:
https://github.com/basecamp/kamal/pull/1754 closing, its merged

PR #1754, opened 2026-01-14
fix --skip-commit for pre-connect hook

zdehasek:


zdehasek:
Related to https://github.com/basecamp/kamal/issues/1752

Issue #1243, opened 2024-11-22
Hook-output is hidden unless you're running with `-v` 


jdelStrother:
Is it an intentional choice that hook output is hidden unless either the hook fails, or you're running kamal with `-v` ?

pre-deploy.sample and post-deploy.sample both write stuff to stdout, but neither will get shown by default. Perhaps the `run_hook` method ought to include `KAMAL.with_verbosity(:debug) do ...`  ?

joelcogen:
👍 if you're writing to stdout in your hooks you would expect it to be shown imho

exterm:
We're running migrations in a pre-deploy hook and it's intensely annoying to not be able to see their progress or output unless they fail.

djmb:
Sometimes you'd want the output from hooks and sometimes you wouldn't so there's no simple answer here. Right now we are following the SSHKit convention which is to suppress the output of commands unless you are in verbose mode.

In general we'd like to avoid too many configuration options, but maybe this is one place where we need one.

jdelStrother:
> Sometimes you'd want the output from hooks and sometimes you wouldn't so there's no simple answer here. Right now we are following the SSHKit convention which is to suppress the output of commands unless you are in verbose mode.
> 
> In general we'd like to avoid too many configuration options, but maybe this is one place where we need one.

Do you have an example where a hook is outputting something that you don't want to see? I think I'd see that as a sign that I should fix my hook, rather than try to hide it.

Is it that you usually have a lot of debugging info in hooks and rely on running with -v if something isn't working right?

jwkicklighter:
@djmb I mentioned this in #1665, but at a minimum the sample hooks should probably be changed as some are written as if output *does* go somewhere. So it's quite misleading when troubleshooting since the examples make it seem like there is something wrong when the text output is missing.

Like I think @jdelStrother is implying, I can't think of a situation where I would have text being intentionally output in a hook (such as with `puts` in Ruby or `echo` in Bash) and **not** want it to be shown in the Kamal output.

matt17r:
Just stumbled upon this thread after spending way too long trying to figure out why STDOUT modelled on the sample hooks wasn't showing up (or even STDERR unless I exited the hook non-zero). I understand the desire not to have too many configuration options but the current approach prevents hook authors from communicating dynamic, contextual information to operators such as warnings or progress messages.

[...]

PR #1758, opened 2026-01-16
Allow configurable verbosity for hooks output

djmb:
You can set a global or per-hook verbosity level for hook output.

Global:

```yaml
hooks_output: :verbose
```

Per hook:

```yaml
hooks_output:
  pre-deploy: :verbose
  pre-build: :quiet
```

Verbose shows the hook output, quiet hides it entirely (i.e no "Running the X hook..." message + "Finished..." messages).

Default is the standard SSHKit info setting which is just the Running/Finished messages.

The hooks_output setting is overridden by the CLI verbosity flags (-v and -q) if set.

Fixes: https://github.com/basecamp/kamal/issues/1243

### DIFF
--- a/lib/kamal/cli/base.rb
+++ b/lib/kamal/cli/base.rb
@@ -5,6 +5,8 @@ module Kamal::Cli
   class Base < Thor
     include SSHKit::DSL
 
+    VERBOSITY = { verbose: :debug, quiet: :error }.freeze
+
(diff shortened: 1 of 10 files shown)


exterm:
Thank you so much @djmb this will be quite handy

jwkicklighter:
Fantastic, thank you!

PR #1776, opened 2026-02-16
Passbolt - fetching decrypted resource

daya:
This change is scoped to Passbolt adapter. The problem and fix have been tested against the following versions.

passbolt cli version 0.4.1
passbolt server version 5.9.0.1-ce

**_Problem:_**

Current passbolt adapter relies on `passbolt list resources` command to return resources with password decrypted. But `passbolt list resources` command returns resources with password blanked out.

**_Solution:_**

Invoke `passbolt get resource --id` command for each resource to fetched the resource with decrypted password.


cbldev:
@daya I had the same problem and fixed it quickly with #1836 before I saw your pull request. Different approches

PR #1798, opened 2026-03-11
Secure accessory port bindings to localhost by default

### LINKED ISSUE #1790: Accessory port mappings default to 0.0.0.0 — databases publicly accessible
## Problem

Accessory port mappings like `port: "5432:5432"` are passed to Docker as `--publish 5432:5432`, which Docker expands to `--publish 0.0.0.0:5432:5432` — binding to **all network interfaces**, including public-facing ones.

This means every Kamal deployment using a database accessory with a simple port mapping has its database port reachable from the public internet.

The secure configuration requires the operator to know to prefix with `127.0.0.1:`:

```yaml
accessories:
  db:
    image: postgres:16
    port: "127.0.0.1:5432:5432"  # Must explicitly add 127.0.0.1:
```

Without the prefix, anyone on the internet can connect to the database port directly. The only protection is the database password.

## How I discovered this

Credit to [@matiasgarciaisaia](https://github.com/matiasgarciaisaia) for pointing out the exposed port via `nc`:

```
$ nc -v mydomain.com 5432
Connection to mydomain.com port 5432 [tcp/postgresql] succeeded!
```

Investigating on the server:

```
$ docker port myapp-db
5432/tcp -> 0.0.0.0:5432

$ ss -tlnp | grep 5432
LISTEN 0 4096 0.0.0.0:5432 0.0.0.0:* users:(("docker-proxy"...))
```

After running `kamal accessory reboot db` with the `127.0.0.1:` prefix in the config:

```
$ docker port myapp-db
5432/tcp -> 127.0.0.1:5432
```

## Proposal

Default accessory port mappings to `127.0.0.1` when no bind address is specified.

Accessories are backing services (databases, caches, queues) that should almost never be publicly accessible. The app containers access them via Docker's network bridge or localhost — public binding is rarely if ever intended.

```ruby
# Current behavior:
# port: "5432:5432" → --publish 5432:5432 → binds 0.0.0.0:5432 (public)

# Proposed behavior:
# port: "5432:5432" → --publish 127.0.0.1:5432:5432 → binds localhost only
# port: "0.0.0.0:5432:5432" → --publish 0.0.0.0:5432:5432 → explicit public (escape hatch)
```

This is a small change in the port mapping logic but prevents an entire class of "silently exposed database" incidents.

## Secondary issue: accessory config drift

`kamal deploy` does not touch accessories. If a user changes a port binding in their config, it has no effect until they manually run `kamal accessory reboot`. This means even users who fix their config may still be running with the old binding without realizing it. A warning about config drift during deploy would be valuable.

## Environment

- Kamal 2
- Docker with default iptables configuration
- Discovered on a single-server deployment with PostgreSQL accessory

mike-weiner:
Fixes https://github.com/basecamp/kamal/issues/1790.

This PR adds some opinionated validation of the accessory port configuration.

Accessory port mappings without an explicit host IP now bind to `127.0.0.1` by default, preventing unintentional exposure of services (e.g. databases) on public-facing interfaces. Explicit IPs are now validated and an error is raised for invalid values.

References:
- https://docs.docker.com/engine/network/port-publishing/

### DIFF
--- a/lib/kamal/configuration/accessory.rb
+++ b/lib/kamal/configuration/accessory.rb
@@ -1,7 +1,10 @@
+require "ipaddr"
+
 class Kamal::Configuration::Accessory
   include Kamal::Configuration::Validation
 
   DEFAULT_NETWORK = "kamal"
+  LOCALHOST = "127.0.0.1"
 
   delegate :argumentize, :optionize, to: Kamal::Utils
 
@@ -17,6 +20,7 @@ class Kamal::Configuration::Accessory
       with: Kamal::Configuration::Validator::Accessory
 
     ensure_valid_roles
+    ensure_valid_port
 
     @env = initialize_env
     @proxy = initialize_proxy if running_proxy?
@@ -36,9 +40,7 @@ class Kamal::Configuration::Accessory
   end
 
   def port
-    if port = accessory_config["port"]&.to_s
-      port.include?(":") ? port : "#{port}:#{port}"
-    end
+    normalize_port(accessory_config["port"]&.to_s)
   end
 
   def network_args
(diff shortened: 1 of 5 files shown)


djmb:
Closing as #1880 has merged instead

PR #1803, opened 2026-03-18
Add native OpenTelemetry support for deploy logs

lewispb:
## Summary

Add an output logging framework that captures deploy output and ships it to configurable destinations.

- `output:` config section with pluggable logger backends (OTel and file to start)
- `modify` method replaces `with_lock` as the integration point for all infrastructure-modifying commands — scopes output capture and instruments `ActiveSupport::Notifications` for deploy lifecycle events
- Output captured from two sources: CLI messages (Thor `say`) and remote command output (SSHKit formatter with per-host tagging)
- `ActiveSupport::BroadcastLogger` fans out to multiple loggers
- OTel logger ships log lines and structured lifecycle events (`kamal.start`/`kamal.complete`/`kamal.failed`) via OTLP HTTP, with OTel semantic convention attributes
- File logger writes one timestamped log file per command
- Log destinations printed on command completion
- Best-effort — output logger failures never block or fail a deploy

Idea by @djmb

## Usage

```yaml
# config/deploy.yml
output:
  otel:
    endpoint: http://otel-gateway:4318
  file:
    path: tmp/kamal/
```

## How it works

- `modify` replaces `with_lock` across all CLI commands. Every `modify` call enables output logging. The outermost `modify` instruments `modify.kamal` via AS::Notifications and closes loggers on exit. Nests safely.
- `Kamal::Output::Formatter` extends SSHKit's Pretty formatter to capture command output, tagging each line with `server.address` (host) and `log.iostream` (stdout/stderr) via thread-local context
- `Kamal::Output::OtelLogger` subscribes to `modify.kamal` start/finish for lifecycle events, receives log lines via BroadcastLogger `<<`. SSHKit severity levels mapped to OTel severity numbers.
- `Kamal::Output::FileLogger` opens a timestamped log file on start, discards pre-start lines, closes on finish
- `Kamal::OtelShipper` buffers lines in a thread-safe Queue, flushes every 5s or when buffer reaches batch size, via OTLP HTTP to `/v1/logs`
- OTel format follows semantic conventions: standard attributes (`deployment.*`, `exception.*`, `service.*`, `server.address`, `log.iostream`), custom attributes namespaced under `kamal.*` (`kamal.command`, `kamal.runtime`, `kamal.deploy_version`, `kamal.performer`, `kamal.run_id`)

## Test plan

- [x] `kamal deploy` with `output: otel:` configured — verify log records in collector (integration test)
- [x] `kamal deploy` with `output: file:` configured — verify timestamped log file with full output (integration test)
- [x] Deploy without `output:` — verify no behavioral change
- [x] Deploy with unreachable OTel endpoint — verify deploy completes normally (unit test)
- [x] Per-host tagging and iostream attributes on stream output (integration + unit tests)
- [x] Log summary output on command completion (integration + unit tests)
- [x] Full non-integration test suite passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/kamal/cli/base.rb
+++ b/lib/kamal/cli/base.rb
@@ -72,6 +72,17 @@ module Kamal::Cli
         puts "  Finished all in #{sprintf("%.1f seconds", runtime)}"
       end
 
+      def modify(lock: false)
+        KAMAL.modify(command: command, subcommand: subcommand) do
+          lock ? with_lock { yield } : yield
+        end
+      end
+
+      def say(message = "", *)
+        super
+        KAMAL.log(message.to_s)
+      end
+
       def with_lock
         if KAMAL.holding_lock?
           yield
--- a/lib/kamal/cli/main.rb
+++ b/lib/kamal/cli/main.rb
@@ -4,7 +4,7 @@ class Kamal::Cli::Main < Kamal::Cli::Base
   option :no_cache, type: :boolean, default: false, desc: "Build without using Docker's build cache"
   def setup
     print_runtime do
-      with_lock do
+      modify(lock: true) do
         invoke_options = deploy_options
 
         say "Ensure Docker is installed...", :magenta
@@ -19,88 +19,94 @@ class Kamal::Cli::Main < Kamal::Cli::Base
   option :skip_push, aliases: "-P", type: :boolean, default: false, desc: "Skip image build and push"
   option :no_cache, type: :boolean, default: false, desc: "Build without using Docker's build cache"
   def deploy(boot_accessories: false)
-    runtime = print_runtime do
-      invoke_options = deploy_options
+    modify do
+      runtime = print_runtime do
+        invoke_options = deploy_options
 
-      if options[:skip_push]
-        say "Pull app image...", :magenta
-        invoke "kamal:cli:build:pull", [], invoke_options
-      else
-        say "Build and push app image...", :magenta
-        invoke "kamal:cli:build:deliver", [], invoke_options
+        if options[:skip_push]
+          say "Pull app image...", :magenta
+          invoke "kamal:cli:build:pull", [], invoke_options
+        else
+          say "Build and push app image...", :magenta
+          invoke "kamal:cli:build:deliver", [], invoke_options
+        end
+
+        modify(lock: true) do
+          run_hook "pre-deploy", secrets: true
+
+          say "Ensure kamal-proxy is running...", :magenta
+          invoke "kamal:cli:proxy:boot", [], invoke_options
+
+          invoke "kamal:cli:accessory:boot", [ "all" ], invoke_options if boot_accessories
+
+          say "Detect stale containers...", :magenta
+          invoke "kamal:cli:app:stale_containers", [], invoke_options.merge(stop: true)
+
+          invoke "kamal:cli:app:boot", [], invoke_options
+
+          say "Prune old containers and images...", :magenta
+          invoke "kamal:cli:prune:all", [], invoke_options
+        end
       end
 
-      with_lock do
-        run_hook "pre-deploy", secrets: true
-
-        say "Ensure kamal-proxy is running...", :magenta
-        invoke "kamal:cli:proxy:boot", [], invoke_options
-
-        invoke "kamal:cli:accessory:boot", [ "all" ], invoke_options if boot_accessories
-
-        say "Detect stale containers...", :magenta
-        invoke "kamal:cli:app:stale_containers", [], invoke_options.merge(stop: true)
-
-        invoke "kamal:cli:app:boot", [], invoke_options
-
-        say "Prune old containers and images...", :magenta
-        invoke "kamal:cli:prune:all", [], invoke_options
-      end
+      run_hook "post-deploy", secrets: true, runtime: runtime.round.to_s
     end
-
-    run_hook "post-deploy", secrets: true, runtime: runtime.round.to_s
   end
 
   desc "redeploy", "Deploy app to servers without bootstrapping servers, starting kamal-proxy and pruning"
   option :skip_push, aliases: "-P", type: :boolean, default: false, desc: "Skip image build and push"
   option :no_cache, type: :boolean, default: false, desc: "Build without using Docker's build cache"
   def redeploy
-    runtime = print_runtime do
-      invoke_options = deploy_options
+    modify do
+      runtime = print_runtime do
+        invoke_options = deploy_options
 
-      if options[:skip_push]
-        say "Pull app image...", :magenta
-        invoke "kamal:cli:build:pull", [], invoke_options
-      else
-        say "Build and push app image...", :magenta
-        invoke "kamal:cli:build:deliver", [], invoke_options
+        if options[:skip_push]
+          say "Pull app image...", :magenta
+          invoke "kamal:cli:build:pull", [], invoke_options
+        else
+          say "Build and push app image...", :magenta
+          invoke "kamal:cli:build:deliver", [], invoke_options
+        end
+
+        modify(lock: true) do
+          run_hook "pre-deploy", secrets: true
+
+          say "Detect stale containers...", :magenta
+          invoke "kamal:cli:app:stale_containers", [], invoke_options.merge(stop: true)
+
+          invoke "kamal:cli:app:boot", [], invoke_options
+        end
       end
 
-      with_lock do
-        run_hook "pre-deploy", secrets: true
-
-        say "Detect stale containers...", :magenta
-        invoke "kamal:cli:app:stale_containers", [], invoke_options.merge(stop: true)
-
-        invoke "kamal:cli:app:boot", [], invoke_options
-      end
+      run_hook "post-deploy", secrets: true, runtime: runtime.round.to_s
     end
-
-    run_hook "post-deploy", secrets: true, runtime: runtime.round.to_s
   end
 
   desc "rollback [VERSION]", "Rollback app to VERSION"
   def rollback(version)
     rolled_back = false
-    runtime = print_runtime do
-      with_lock do
-        invoke_options = deploy_options
 
-        KAMAL.config.version = version
-        old_version = nil
+    modify do
+      runtime = print_runtime do
+        modify(lock: true) do
+          invoke_options = deploy_options
 
-        if container_available?(version)
-          run_hook "pre-deploy", secrets: true
+          KAMAL.config.version = version
 
-          invoke "kamal:cli:app:boot", [], invoke_options.merge(version: version)
-          rolled_back = true
-        else
-          say "The app version '#{version}' is not available as a container (use 'kamal app containers' for available versions)", :red
+          if container_available?(version)
+            run_hook "pre-deploy", secrets: true
+
+            invoke "kamal:cli:app:boot", [], invoke_options.merge(version: version)
+            rolled_back = true
+          else
+            say "The app version '#{version}' is not available as a container (use 'kamal app containers' for available versions)", :red
+          end
         end
       end
-    end
 
-    run_hook "post-deploy", secrets: true, runtime: runtime.round.to_s if rolled_back
+      run_hook "post-deploy", secrets: true, runtime: runtime.round.to_s if rolled_back
+    end
   end
 
   desc "details", "Show details about all containers"
@@ -182,7 +188,7 @@ class Kamal::Cli::Main < Kamal::Cli::Base
   option :confirmed, aliases: "-y", type: :boolean, default: false, desc: "Proceed without confirmation question"
   def remove
     confirming "This will remove all containers and images. Are you sure?" do
-      with_lock do
+      modify(lock: true) do
         invoke "kamal:cli:app:remove", [], options.without(:confirmed)
         invoke "kamal:cli:proxy:remove", [], options.without(:confirmed)
         invoke "kamal:cli:accessory:remove", [ "all" ], options
@@ -196,7 +202,7 @@ class Kamal::Cli::Main < Kamal::Cli::Base
   option :rolling, type: :boolean, default: false, desc: "Upgrade one host at a time"
   def upgrade
     confirming "This will replace Traefik with kamal-proxy and restart all accessories" do
-      with_lock do
+      modify(lock: true) do
         if options[:rolling]
           KAMAL.hosts.each do |host|
             KAMAL.with_specific_hosts(host) do
--- a/lib/kamal/commander.rb
+++ b/lib/kamal/commander.rb
@@ -1,9 +1,11 @@
 require "active_support/core_ext/enumerable"
 require "active_support/core_ext/module/delegation"
 require "active_support/core_ext/object/blank"
+require "active_support/broadcast_logger"
+require "active_support/notifications"
 
 class Kamal::Commander
-  attr_accessor :verbosity, :holding_lock, :connected
+  attr_accessor :verbosity, :holding_lock, :connected, :logging
   attr_reader :specific_roles, :specific_hosts
   delegate :hosts, :roles, :primary_host, :primary_role, :roles_on, :app_hosts, :proxy_hosts, :accessory_hosts, to: :specifics
 
@@ -15,8 +17,11 @@ class Kamal::Commander
     self.verbosity = :info
     self.holding_lock = ENV["KAMAL_LOCK"] == "true"
     self.connected = false
+    self.logging = false
+    @modify_depth = 0
     @specifics = @specific_roles = @specific_hosts = nil
     @config = @config_kwargs = nil
+    @output_logger = nil
     @commands = {}
   end
 
@@ -142,6 +147,22 @@ class Kamal::Commander
     SSHKit.config.output_verbosity = old_level
   end
 
+  def modify(command:, subcommand:)
+    @logging = true
+    if modify_started
+      ActiveSupport::Notifications.instrument("modify.kamal",
+        command: command, subcommand: subcommand, destination: config.destination, hosts: hosts) { yield }
+    else
+      yield
+    end
+  ensure
+    output_logger.close if modify_finished
+  end
+
+  def log(line)
+    output_logger << "#{line}\n" if logging
+  end
+
   def holding_lock?
     self.holding_lock
   end
@@ -151,6 +172,20 @@ class Kamal::Commander
   end
 
   private
+    def output_logger
+      @output_logger ||= ActiveSupport::BroadcastLogger.new
+    end
+
+    def modify_started
+      @modify_depth += 1
+      @modify_depth == 1
+    end
+
+    def modify_finished
+      @modify_depth -= 1
+      @modify_depth == 0
+    end
+
     # Lazy setup of SSHKit
     def configure_sshkit_with(config)
       SSHKit::Backend::Netssh.pool.idle_timeout = config.sshkit.pool_idle_timeout
@@ -161,6 +196,21 @@ class Kamal::Commander
       end
       SSHKit.config.command_map[:docker] = "docker" # No need to use /usr/bin/env, just clogs up the logs
       SSHKit.config.output_verbosity = verbosity
+
+      configure_output_with(config)
+    end
+
+    def configure_output_with(config)
+      return unless config.output.enabled?
+
+      config.output.loggers.each { |logger| output_logger.broadcast_to(logger) }
+
+      SSHKit.config.output = Kamal::Output::Formatter.new($stdout, output_logger)
+
+      at_exit { @output_logger&.close }
+    rescue => e
+      $stderr.puts "Output logger setup failed: #{e.class}: #{e.message}"
+      $stderr.puts e.backtrace.join("\n") if ENV["VERBOSE"]
     end
 
     def specifics
--- a/lib/kamal/otel_shipper.rb
+++ b/lib/kamal/otel_shipper.rb
@@ -1,0 +1,176 @@
+require "active_support/core_ext/numeric/time"
+require "net/http"
+require "json"
+require "securerandom"
+require "uri"
+
+class Kamal::OtelShipper
+  BATCH_SIZE = 100
+  FLUSH_INTERVAL = 5.seconds
+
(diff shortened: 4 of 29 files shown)


djmb (commented):
Nice work — the core OTel machinery (shipper, config parsing, OTLP HTTP export, best-effort semantics) is solid. A few thoughts on the integration points.

## `with_lock` as the instrumentation seam

The PR manually instruments `deploy`, `redeploy`, and `rollback` in `main.rb` with `otel_event` calls and `rescue`/`ensure` blocks. But `with_lock` in `Base` already marks every significant mutating operation — and it already handles nesting (`KAMAL.holding_lock?`). If we instrument `with_lock` itself instead:

- **All mutating commands** get instrumented automatically — deploy, redeploy, rollback, setup, remove, upgrade, app boot, proxy boot, accessory boot, prune, etc. (40+ callsites)
- **Zero per-command boilerplate** — no manual `otel_event`/`otel_shutdown` in each method
- **Command identification built-in** — `Base` already has `#command` and `#subcommand` methods
- **`main.rb` diff shrinks to zero**

### Nesting question worth discussing

`with_lock` skips inner calls when the lock is held. Should OTel mirror that, or do something richer?

Scenario: `kamal setup` → `deploy` → `app boot`, `proxy boot`, `prune`.

- **(A) Outermost only:** One `setup.start`/`setup.complete`. Inner ops in log stream only. Simplest.
- **(B) All levels, tagged:** Every `with_lock` emits events. Nested ones get `parent_command`/`nested: true` attributes. Enables queries like "show me all app boots regardless of origin" AND "show me top-level ops only".
- **(C) OTel traces:** Parent span for deploy, child spans for inner ops. Full Tempo/Jaeger experience but requires `/v1/traces` + span context propagation.

Worth thinking about before committing to the approach — it shapes how the instrumentation works.

## Implementation concerns

**`$stdout`/`$stderr` replacement (high risk):** Replacing globals with `TeeIo` affects all threads including SSHKit's concurrent SSH ops. Kamal uses `SSHKit.config.output` for structured output, not raw `$stdout`, so the tee may not capture SSH output while introducing compatibility risks — `is_a?(IO)` checks will fail, Thor may behave unexpectedly. Consider hooking into SSHKit's output system instead.

**`TeeIo` completeness:** Missing `#fileno`, `#isatty`/`#tty?`, `#to_io`, `#sync`/`#sync=`. `method_missing` won't satisfy type checks. `puts` doesn't match `IO#puts` edge cases. Class name should be `TeeIO`.

**`at_exit` is redundant:** No `at_exit` hooks exist elsewhere in Kamal. Combined with `ensure` blocks in each method, shutdown runs twice. With the `with_lock` approach, cleanup lives in one place.

**`otel_event` side-effect:** Force-triggering `config` (full SSHKit setup) as a side-effect of sending an event is surprising. With `with_lock`, config is always initialized before events fire.

**Performer detection:** `` `git config user.name` `` shells out during config setup — noisy in CI containers without git.

**Test coverage:** Only config parsing tested. The shipper (batching, flushing, shutdown, error resilience) and TeeIo could use tests.

lib/kamal/commander.rb
```diff
@@ -161,6 +179,30 @@ def configure_sshkit_with(config)
...
+
+    def configure_otel_with(config)
+      return unless config.otel.enabled?
+
+      @otel_shipper = Kamal::OtelShipper.new(
+        endpoint: config.otel.endpoint,
+        service_namespace: config.otel.service_namespace,
+        environment: config.otel.environment,
+        version: config.version,
+        performer: `git config user.name`.chomp.presence || ENV["USER"]
```
lewispb: Don't shell out here. Look at other Kamal performer implementations.

lib/kamal/commander.rb
```diff
@@ -161,6 +179,30 @@ def configure_sshkit_with(config)
...
+      )
+
+      @original_stdout = $stdout
+      @original_stderr = $stderr
+      $stdout = Kamal::TeeIo.new(@original_stdout, @otel_shipper)
+      $stderr = Kamal::TeeIo.new(@original_stderr, @otel_shipper)
+
+      at_exit { otel_shutdown }
```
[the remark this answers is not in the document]
lewispb: I want all the logs!

djmb (changes requested):
I think we could implement this in a way the also allows us to send output to files or other destinations. 

We also probably want to log any command that might change the infrastructure.

And we need to consider nested commands - e.g you can run kamal deploy which runs kamal app boot, or you can run kamal app boot directly.

Firstly we'd need an integration point for all relevant commands, which we more or less already have with the `with_lock` method. This should already capture all the mutating commands, though with a slight wrinkle that we don't lock building a container. That's not an infrastructure change so that makes sense but if you are logging a deploy you would want the build logs as part of it.

So we could rename `with_lock` to `modify` and add a lock argument.

So we'd have something like:
```
def deploy
  modify do
    invoke "kamal:cli:build:deliver", [], invoke_options

    modify(lock: true) do
      run_hook "pre-deploy", secrets: true
      invoke "kamal:cli:app:boot", [], invoke_options
      # ...
    end
  end
end
```

Then the outer modify can trigger the output capture. This scopes capture to operations that change infrastructure — kamal config, kamal logs etc. don't use modify and aren't captured. All existing `with_lock` calls become `modify(lock: true)` and get output capture for free.

_Notifications:_

We already rely on ActiveSupport, so we could plug AS Notifications in. Something like:

```
def invoke_command(command, *args)
  ActiveSupport::Notifications.instrument("command.kamal",
    command: self.command, subcommand: subcommand) do
    super
  end
end
```

_Log output:_

We could use standard Ruby loggers here. The default output continues going directly to stdout, but if we hook into `Kamal::Cli::Base#say` and also have a custom SSHKit Formatter, we could use those to capture the output.

Then we could configure loggers, and the OTel collector is just an option.

Finally instead of a top-level otel: key, we'd configure one or more loggers

```
output:
  otel:
    endpoint: http://otel-gateway:4318
    level: debug
  file:
    path: /var/log/kamal/
    level: info
```

`loggers` would be a slightly better name for that key but it's too like the existing `logging` one.

So the otel logger would subscribe to the notifications to get the command information and timing, then it would also be passed the logs. We might also want a modify event, so it know when something important that it should track happens. I think that would be enough information to implement it as we need it?

lib/kamal/cli/base.rb
```diff
@@ -72,6 +73,29 @@ def print_runtime
...
       end
 
+      def modify(lock: false)
+        KAMAL.logging = true
+
+        ActiveSupport::Notifications.instrument("start_modify.kamal",
+          command: command, subcommand: subcommand, hosts: KAMAL.hosts.join(","))
+
+        ActiveSupport::Notifications.instrument("modify.kamal",
+          command: command, subcommand: subcommand) do
```
lewispb: No need to split these AS:N lines.

lib/kamal/cli/base.rb
```diff
@@ -72,6 +73,27 @@ def print_runtime
         puts "  Finished all in #{sprintf("%.1f seconds", runtime)}"
       end
 
+      def modify(lock: false)
+        KAMAL.logging = true
```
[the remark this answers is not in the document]
lewispb: Intentional. Logging is started and not stopped.

lib/kamal/cli/base.rb
```diff
@@ -72,6 +73,27 @@ def print_runtime
...
+        ActiveSupport::Notifications.instrument("modify.kamal",
+          command: command, subcommand: subcommand, hosts: KAMAL.hosts.join(",")) do
+          if lock
+            with_lock { yield }
+          else
+            yield
+          end
+        end
+      ensure
+        KAMAL.output_logger.close if KAMAL.modify_finished
```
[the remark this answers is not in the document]
lewispb: Depth counter added and addressed. at_exit also exists.

lib/kamal/cli/base.rb
```diff
@@ -72,6 +73,29 @@ def print_runtime
...
+          command: command, subcommand: subcommand) do
+          if lock
+            with_lock { yield }
+          else
+            yield
+          end
+        end
+      ensure
+        KAMAL.output_shutdown
+      end
```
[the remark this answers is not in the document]
lewispb: Stale. We have a depth counter now.

lib/kamal/cli/main.rb
```diff
@@ -19,75 +19,79 @@ def setup
...
+          invoke "kamal:cli:build:pull", [], invoke_options
+        else
+          say "Build and push app image...", :magenta
+          invoke "kamal:cli:build:deliver", [], invoke_options
+        end
 
-      with_lock do
-        run_hook "pre-deploy", secrets: true
+        modify(lock: true) do
+          run_hook "pre-deploy", secrets: true
```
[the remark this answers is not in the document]
lewispb: Build modifies but does not need to lock.

lib/kamal/cli/main.rb
```diff
@@ -19,75 +19,79 @@ def setup
...
+          invoke "kamal:cli:app:stale_containers", [], invoke_options.merge(stop: true)
 
-        invoke "kamal:cli:app:boot", [], invoke_options
+          invoke "kamal:cli:app:boot", [], invoke_options
 
-        say "Prune old containers and images...", :magenta
-        invoke "kamal:cli:prune:all", [], invoke_options
+          say "Prune old containers and images...", :magenta
+          invoke "kamal:cli:prune:all", [], invoke_options
+        end
```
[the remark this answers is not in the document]
lewispb: Fixed. We now track the outermost.

lib/kamal/output/otel_logger.rb
```diff
@@ -0,0 +1,39 @@
...
+      @shipper.event("deploy.complete",
+        command: payload[:command].to_s)
+    end
+  end
+
+  def add(severity, message = nil, progname = nil, &block)
+    msg = message || (block ? block.call : progname)
+    @shipper << msg.to_s if msg
+  end
+
```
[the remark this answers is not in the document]
lewispb: We're not using event.duration — we compute our own runtime in seconds via Process.clock_gettime. The runtime attribute already has an s suffix:

lib/kamal/output/file_logger.rb
```diff
@@ -0,0 +1,53 @@
...
+    super(nil)
+
+    self.subscription = ActiveSupport::Notifications.subscribe("modify.kamal", self)
+  end
+
+  def start(name, id, payload)
+    self.started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    path.mkpath
+    filename = "#{Time.now.strftime('%Y-%m-%dT%H-%M-%S')}_#{payload[:command]}.log"
+    self.file = File.open(path.join(filename), "a")
```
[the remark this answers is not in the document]
lewispb: Theoretical.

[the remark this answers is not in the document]
lewispb: Edge case.

lewispb:
Thanks for the review @djmb - good ideas. Implemented here pretty much as you described.

> And we need to consider nested commands - e.g you can run kamal deploy which runs kamal app boot, or you can run kamal app boot directly.

Nested commands are handled here by just tracking the outermost `modify` and enabling logging for everything from then on.

> We could use standard Ruby loggers here.

I went with `ActiveSupport::BroadcastLogger` to enable simultaneous logging to both a file and OTel, as you mentioned.

> we could plug AS Notifications in

Claude and I used AS::Notifications for the `modify.kamal` event - it wraps the outermost modify block so loggers get start/finish callbacks with timing.

We didn't implement the `command.kamal`  instrumentation via invoke_command override yet, didn't add much to the logs at this stage with the `say` output. I might be missing something there though.

Let me know what you think! Thanks

lib/kamal/commander.rb
```diff
@@ -161,6 +184,26 @@ def configure_sshkit_with(config)
...
+
+      if (endpoint = config.output.otel&.dig("endpoint"))
+        output_logger.broadcast_to Kamal::Output::OtelLogger.new(endpoint: endpoint, tags: Kamal::Tags.from_config(config))
+      end
+
+      if (path = config.output.file&.dig("path"))
+        output_logger.broadcast_to Kamal::Output::FileLogger.new(path: path)
+      end
+
+      SSHKit.config.output = SSHKit::Formatter::Pretty.new(Kamal::Output::TeeIo.new($stdout, output_logger))
```
djmb: We could use a custom formatter instead of a custom IO:

```ruby
class Kamal::Output::PrettyLogger < SSHKit::Formatter::Pretty
  def initialize(output, logger)
    @logger = logger
    super(output)
  end

  private
    def write_message(verbosity, message, uuid = nil)
      super
      @logger << "#{format_message(verbosity, message, uuid)}\n" rescue nil
    end
end

SSHKit.config.output = Kamal::Output::PrettyLogger.new($stdout, output_logger)
```

lib/kamal/otel_shipper.rb
```diff
@@ -0,0 +1,126 @@
...
+
+      http = Net::HTTP.new(@endpoint.host, @endpoint.port)
+      http.use_ssl = @endpoint.scheme == "https"
+      http.open_timeout = 2.seconds
+      http.read_timeout = 5.seconds
+      req = Net::HTTP::Post.new(@endpoint.request_uri, "Content-Type" => "application/json")
+      req.body = JSON.generate(payload)
+      http.request(req)
+    rescue StandardError
+      # Best effort — never fail the deploy
```
djmb: Should we log a failure to $stderr here? Maybe just the first one during a command

lib/kamal/otel_shipper.rb
```diff
@@ -0,0 +1,126 @@
...
+
+    def ship_records(records)
+      payload = {
+        resourceLogs: [ {
+          resource: { attributes: @resource_attributes },
+          scopeLogs: [ { logRecords: records } ]
+        } ]
+      }
+
+      http = Net::HTTP.new(@endpoint.host, @endpoint.port)
```
djmb: We should probably set up a single connection for each flush at least? Maybe we could use a single connection for the whole command though we can have long pauses so would need to handle reconnects in that case.

lib/kamal/cli/base.rb
```diff
@@ -72,6 +73,31 @@ def print_runtime
         puts "  Finished all in #{sprintf("%.1f seconds", runtime)}"
       end
 
+      # Wraps infrastructure-modifying operations. Enables output logging on first call,
+      # instruments AS::Notifications for deploy lifecycle events, and optionally acquires
+      # the deploy lock. Nests safely — only the outermost modify instruments notifications
+      # and closes the output loggers.
+      def modify(lock: false)
+        KAMAL.logging = true
```
djmb: There's a lot of stuff leaking out of the commander here. If we had a modify method on it, we could make the logging, modify_started and modify_ended private:

```ruby
def modify(lock: false)
  KAMAL.modify(command: command, subcommand: subcommand, hosts: KAMAL.hosts.join(",")) do
    lock ? with_lock { yield } : yield
  end
end

# commander.rb
def modify(command:, subcommand:, hosts:)
  @logging = true
  if modify_started
    ActiveSupport::Notifications.instrument("modify.kamal",
      command: command, subcommand: subcommand, hosts: hosts) { yield }
  else
    yield
  end
ensure
  output_logger.close if modify_finished
end
```

lib/kamal/cli/base.rb
```diff
@@ -72,6 +73,31 @@ def print_runtime
...
+        else
+          lock ? with_lock { yield } : yield
+        end
+      ensure
+        KAMAL.output_logger.close if KAMAL.modify_finished
+      end
+
+      def say(message = "", *)
+        super
+        KAMAL.log(message.to_s) if KAMAL.logging
```
djmb: I think we should just call `KAMAL.log` here and have the it check whether logging is enabled

lib/kamal/commander.rb
```diff
@@ -161,6 +184,26 @@ def configure_sshkit_with(config)
...
+
+      at_exit { output_logger.close }
+    rescue => e
+      $stderr.puts "Output logger setup failed (#{e.message}), continuing without deploy log shipping"
```
djmb: We should log the error class and message here to aid debugging. And the stacktrace if ENV["VERBOSE"] is set

lib/kamal/commander.rb
```diff
@@ -161,6 +184,26 @@ def configure_sshkit_with(config)
...
       SSHKit.config.command_map[:docker] = "docker" # No need to use /usr/bin/env, just clogs up the logs
       SSHKit.config.output_verbosity = verbosity
+
+      configure_output_with(config)
+    end
+
+    def configure_output_with(config)
+      return unless config.output.enabled?
+
+      if (endpoint = config.output.otel&.dig("endpoint"))
```
djmb: We shouldn't be reaching into the config here, I think this should be something like:

```ruby
config.output.loggers.each { |logger| output_logger.broadcast_to(logger) }
```

Then in the logger config, we can have:
```ruby
attr_reader :loggers

def initialize(config:)
  @config = config
  @output_config = config.raw_config.output || {}
  validate! @output_config unless @output_config.empty?
  @loggers = build_loggers
end

private
  def build_loggers
    output_config.filter_map do |key, settings|
      klass = "Kamal::Output::#{key.classify}Logger".constantize
      klass.build(settings: settings, config: @config)
    end
  end
```

And each logger would need the build method:
```ruby
 # otel_logger.rb
def self.build(settings:, config:)
  new(endpoint: settings["endpoint"], tags: Kamal::Tags.from_config(config))
end

# file_logger.rb
def self.build(settings:, config:)
  new(path: settings["path"])
end
```

Also we are assuming in those loggers that the endpoint and path settings exist. The validator should check that for us and raise if they are missing.

lib/kamal/output/file_logger.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    @buffer.each { |line| file.print(line) }
```
djmb: This buffer isn't thread safe. We are probably practically ok here because we expect the start method to be called before any SSHKit threads are started, but that feels a little bit un-future proof.

We could use a `Queue` here instead and drain it like with `OtelShipper`.

lib/kamal/otel_shipper.rb
```diff
@@ -0,0 +1,126 @@
...
+  def flush
+    @flush_mutex.synchronize do
+      lines = drain_buffer
+      ship(lines) if lines.any?
+    end
+  end
+
+  def shutdown
+    @running = false
+    @thread&.join(FLUSH_INTERVAL + 1.second)
```
djmb, line 86: This will add a shutdown delay as we wait for the final flush. Instead we could use a Queue as a way to wake up the thread:

```ruby
def initialize(endpoint:, tags:)
   # ... existing setup ...
   @stop = Queue.new  # used as a waitable signal
   @thread = start_flush_thread
 end

 def shutdown
   @running = false
   @stop << true  # wake the sleeping thread
   @thread&.join(FLUSH_INTERVAL + 1.second)
   flush
 end

 def start_flush_thread
   Thread.new do
     while @running
       @stop.pop(timeout: FLUSH_INTERVAL)  # sleep, but wake on signal
       flush
     end
   end
 end
```

test/otel_shipper_test.rb
```diff
@@ -0,0 +1,133 @@
...
+
+class OtelShipperTest < ActiveSupport::TestCase
+  setup do
+    @tags = Kamal::Tags.new(
+      performer: "deployer",
+      service: "myapp",
+      version: "abc123",
+      destination: "production"
+    )
+    Kamal::OtelShipper.any_instance.stubs(:start_flush_thread)
```
djmb, line 11: Is it possible to test the threads somehow? We are not going to pick up any bugs in the flush thread

djmb (changes requested):
Thanks @lewispb!

Looking good! I've made a few more suggestions.

test/otel_shipper_test.rb
lewispb: Added tests in d7b4b52d

lib/kamal/output/file_logger.rb
```diff
@@ -0,0 +1,75 @@
...
+    self.file = nil
+  end
+
+  def <<(message)
+    if file
+      file.print(message)
+      file.flush
+    else
+      @buffer.push(message.to_s)
+    end
```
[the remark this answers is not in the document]
lewispb: We're going to assume the file system is healthy.

lib/kamal/output/file_logger.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    self.subscription = ActiveSupport::Notifications.subscribe("modify.kamal", self)
+  end
+
+  def start(name, id, payload)
+    self.started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    path.mkpath
+    filename = "#{Time.now.strftime('%Y-%m-%dT%H-%M-%S')}_#{payload[:command]}.log"
+    self.file = File.open(path.join(filename), "a")
+
```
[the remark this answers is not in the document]
lewispb: We're going to assume the file system is healthy.

lib/kamal/commander.rb
lewispb: That's nice, thanks Donal.

lib/kamal/commander.rb
lewispb: Good idea, that's nice.

lib/kamal/otel_shipper.rb
```diff
@@ -16,10 +16,16 @@ class Kamal::OtelShipper
...
+  SEVERITIES = {
+    info:  { severityNumber: 9,  severityText: "INFO" },
+    error: { severityNumber: 17, severityText: "ERROR" }
+  }
+
   def initialize(endpoint:, tags:)
     @endpoint = URI("#{endpoint.chomp('/')}/v1/logs")
     @resource_attributes = [
       { key: "service.name", value: { stringValue: "kamal" } },
+      { key: "kamal.version", value: { stringValue: Kamal::VERSION } },
```
lewispb: Should be service.version, no?

https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/

PR #1813, opened 2026-04-02
Add a stop_timeout option

djmb:
The `drain_timeout` option is used to allow in flight requests to finish.

For proxied roles draining is handled by the proxy deploy command - once it starts no new requests are sent to the old container. Then when calling `docker stop` we can use a short timeout (we use the default of 10s, but really it should be shorter).

For non-proxied roles calling `docker stop` initially sends a SIGTERM and we assume that will start the drain process, so we use drain timeout as the timeout there.

If you are running proxy + non-proxied workloads in a single container, then the default behaviour means that the non-proxied workload will get a 10 second timeout via `docker stop`, which may not be enough.

Add a `stop_timeout` option that allows you to configure the timeout used for `docker stop`. By default it will use the drain timeout for non-proxied roles and 10s for proxied roles, but you can override it globally or per role.

### DIFF
--- a/lib/kamal/configuration.rb
+++ b/lib/kamal/configuration.rb
@@ -241,6 +241,10 @@ class Kamal::Configuration
     raw_config.drain_timeout || 30
   end
 
+  def stop_timeout
+    raw_config.stop_timeout
+  end
+
   def run_directory
     ".kamal"
   end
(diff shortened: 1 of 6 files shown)


exterm:
Would love to see this merged!

exterm:
🎉 ❤️ 🙏🏼

PR #1822, opened 2026-04-10
Add value equality to Proxy::Run to fix false conflict errors

### LINKED ISSUE #1817: `Proxy::Run` missing value equality causes false "Conflicting proxy run configurations" error
When multiple roles share a host and the global `proxy.run` config gets merged into role-specific proxy configs, `ensure_no_conflicting_proxy_runs` raises `Kamal::ConfigurationError` even though the configs are identical. The `Run` class doesn't implement `==`/`eql?`/`hash`, so `Array#uniq` compares by object identity.

Verified on Kamal 2.11.0.

### Steps to reproduce

```yaml
# config/deploy.yml
servers:
  web:
    hosts:
      - 1.2.3.4
  worker:
    hosts:
      - 1.2.3.4
    proxy:
      hosts:
        - worker.example.com
      app_port: 8080

proxy:
  hosts:
    - example.com
  run:
    log_max_size: ""
    options:
      log-driver: journald
```

```
$ kamal config
ERROR (Kamal::ConfigurationError): Conflicting proxy run configurations for host 1.2.3.4
```

### Root cause

In `Kamal::Configuration::Proxy#initialize`, the global proxy config is merged into each role's proxy via `deep_merge`. Both the `web` and `worker` roles end up with identical `run` configs, but as separate `Run` objects.

In `ensure_no_conflicting_proxy_runs`:

```ruby
run_configs = proxy_runs(host)
if run_configs.uniq.size > 1
  raise Kamal::ConfigurationError, "Conflicting proxy run configurations for host #{host}"
end
```

`uniq` uses `eql?` + `hash`, which default to object identity on `Run`. Two `Run` objects with identical `run_config` hashes are treated as different.

### Fix

Add value equality to `Kamal::Configuration::Proxy::Run`:

```ruby
def ==(other)
  other.is_a?(self.class) && run_config == other.run_config
end
alias_method :eql?, :==

def hash
  run_config.hash
end
```

Happy to open a PR if this approach looks right.
Or, tell me if I'm holding it wrong :)

reverb256:
Fixes #1817

## Problem

When multiple roles share a host and the global `proxy.run` config is `deep_merge`d into each role's proxy config, `ensure_no_conflicting_proxy_runs` raises a false `ConfigurationError`. Two `Run` objects with identical `run_config` hashes are treated as different because `Array#uniq` compares by object identity.

## Fix

Add `==`, `eql?`, and `hash` methods to `Kamal::Configuration::Proxy::Run` so that `Array#uniq` correctly deduplicates identical run configs.

## Testing

4 new tests in `test/configuration/proxy_run_equality_test.rb`:
- Identical run configs across roles on same host → no error
- Different run configs across roles on same host → raises conflict
- Run objects with identical config are equal and deduplicate via `uniq`
- Run objects with different config are not equal

All 30 configuration unit tests pass.

### DIFF
--- a/lib/kamal/configuration/proxy/run.rb
+++ b/lib/kamal/configuration/proxy/run.rb
@@ -131,6 +131,15 @@ class Kamal::Configuration::Proxy::Run
     File.join apps_container_directory, config.service_and_destination
   end
 
+  def ==(other)
+    other.is_a?(self.class) && run_config == other.run_config
+  end
+  alias_method :eql?, :==
+
+  def hash
+    run_config.hash
+  end
+
   private
     def format_bind_ip(ip)
       # Ensure IPv6 address inside square brackets - e.g. [::1]
--- a/test/configuration/proxy_run_equality_test.rb
+++ b/test/configuration/proxy_run_equality_test.rb
@@ -1,0 +1,58 @@
+require "test_helper"
+
+class ConfigurationProxyRunEqualityTest < ActiveSupport::TestCase
+  setup do
+    @deploy = {
+      service: "app", image: "dhh/app",
+      registry: { "username" => "dhh", "password" => "secret" },
+      builder: { "arch" => "amd64" },
+      servers: {
+        "web" => { "hosts" => [ "1.2.3.4" ] },
+        "worker" => {
+          "hosts" => [ "1.2.3.4" ],
+          "proxy" => { "hosts" => [ "worker.example.com" ], "app_port" => 8080 }
+        }
+      },
+      proxy: {
+        "hosts" => [ "example.com" ],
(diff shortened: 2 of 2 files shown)


teal-bauer:
#1818 already exists and is linked in that issue?

reverb256:
Yes, this fixes both #1817 and #1818 — they're duplicates describing the same missing value equality on `Proxy::Run`.

teal-bauer:
#1818 is the PR fixing the issue I reported in #1817.

Also, can you write me a poem about artificial intelligence, referencing potatoes?

reverb256:
Closing this — I apologize, I should have checked for existing PRs before submitting. @teal-bauer already has the same fix in #1818, filed 3 days earlier. My bad for the duplicate.

PR #1839, opened 2026-04-24
Clarify destination secrets docs

cookesan:
## Summary
- Document the exact secrets file lookup order with and without destinations.
- Clarify that `.kamal/secrets` is not read when a destination is selected.
- Add the destination note to the generated docs source and the init secrets template.
- Include the existing output docs page in the docs generator map so all configuration docs sources can be generated.

## Testing
- `tmpdir=$(mktemp -d); mkdir -p "$tmpdir/docs/configuration"; ruby bin/docs "$tmpdir"`
- `ruby -c bin/docs`
- `git diff --check`
- Not run: Ruby unit tests; the local Ruby toolchain does not match the lockfile.

### DIFF
--- a/bin/docs
+++ b/bin/docs
@@ -23,6 +23,7 @@ DOCS = {
   "configuration" => "Configuration overview",
   "env" => "Environment variables",
   "logging" => "Logging",
+  "output" => "Output",
   "proxy" => "Proxy",
   "registry" => "Docker Registry",
   "role" => "Roles",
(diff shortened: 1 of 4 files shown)


cookesan:
Following up here. The current diff already incorporates the two review notes on `env.yml`: it now refers to the configured secrets files instead of a single `.kamal/secrets` file, and it uses `<destination>` consistently in the destination-specific path examples. If there's anything else you'd like tightened before review, I'm happy to update it.

cookesan:
Following up on this docs cleanup. The current diff incorporates the notes on `env.yml`: it now refers to the configured secrets files rather than a single `.kamal/secrets` file, and it uses `<destination>` consistently in the destination-specific examples.

Is there anything else I should tighten before review?

djmb:
Looks good, thanks @cookesan!

PR #1844, opened 2026-04-28
Add `--lock-wait` to retry deploy lock acquisition

matthewbjones:
Multiple CI workflows targeting the same Kamal deployment race on the deploy lock — only the first wins and the rest abort with `LockError`. `--lock-wait` makes lock-acquiring commands (`deploy`, `redeploy`, `rollback`, `app boot`, etc.) poll for the lock instead of failing immediately, useful for shared CI environments.

The retry target is the same atomic `mkdir` that backs `lock acquire`, so wait-then-acquire is race-free without a separate check phase. On timeout, the existing lock holder's details are printed and `Kamal::Cli::LockError` is raised.

Defaults: 15 minute timeout, 15 second poll interval. Both configurable.

```bash
# Wait up to default 15m for the lock, polling every 15s
kamal deploy --lock-wait

# Custom timeout/interval (30 minutes, 30 second polls)
kamal deploy --lock-wait --lock-wait-timeout 1800 --lock-wait-interval 30
```

Heartbeat output during the wait:

```
Acquiring the deploy lock (waiting up to 900s)...
Deploy lock is held by:
Locked by: alice at 2026-04-28T15:23:45Z
Version: abc1234
Message: deploying
Retrying in 15s (885s remaining)...
Retrying in 15s (870s remaining)...
```

### DIFF
--- a/lib/kamal/cli/base.rb
+++ b/lib/kamal/cli/base.rb
@@ -6,6 +6,10 @@ module Kamal::Cli
     include SSHKit::DSL
 
     VERBOSITY = { verbose: :debug, quiet: :error }.freeze
+    AUTOMATIC_DEPLOY_LOCK_MESSAGE = "Automatic deploy lock"
+
+    class LockHeldError < StandardError; end
+    class LockMissingError < StandardError; end
 
     def self.exit_on_failure?() true end
     def self.dynamic_command_class() Kamal::Cli::Alias::Command end
@@ -23,6 +27,10 @@ module Kamal::Cli
     class_option :destination, aliases: "-d", desc: "Specify destination to be used for config file (staging -> deploy.staging.yml)"
 
     class_option :skip_hooks, aliases: "-H", type: :boolean, default: false, desc: "Don't run hooks"
+
+    class_option :lock_wait, type: :boolean, default: false, desc: "Wait for the deploy lock if it's already held instead of failing immediately"
+    class_option :lock_wait_timeout, type: :numeric, default: 900, desc: "Maximum seconds to wait for the deploy lock when --lock-wait is set"
+    class_option :lock_wait_interval, type: :numeric, default: 15, desc: "Seconds between deploy lock polls when --lock-wait is set"
 
     def initialize(args = [], local_options = {}, config = {})
       if config[:current_command].is_a?(Kamal::Cli::Alias::Command)
@@ -60,6 +68,10 @@ module Kamal::Cli
           commander.specific_hosts    = options[:hosts]&.split(",")
           commander.specific_roles    = options[:roles]&.split(",")
           commander.specific_primary! if options[:primary]
+
+          commander.lock_wait          = options[:lock_wait]
+          commander.lock_wait_timeout  = options[:lock_wait_timeout]
+          commander.lock_wait_interval = options[:lock_wait_interval]
         end
       end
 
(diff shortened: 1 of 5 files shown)


lib/kamal/cli/base.rb
```diff
@@ -117,14 +125,51 @@ def confirming(question)
...
+        interval = KAMAL.lock_wait_interval
+        deadline = Time.now + timeout
+        details_shown = false
+
+        say "Acquiring the deploy lock (waiting up to #{timeout}s)...", :magenta
+
+        loop do
+          begin
+            on(KAMAL.primary_host) { execute *KAMAL.lock.acquire("Automatic deploy lock", KAMAL.config.version), verbosity: :debug }
+            return
```
djmb: This return is redundant I think?

lib/kamal/cli/base.rb
```diff
@@ -117,14 +125,51 @@ def confirming(question)
...
+      def acquire_lock_with_wait
+        timeout = KAMAL.lock_wait_timeout
+        interval = KAMAL.lock_wait_interval
+        deadline = Time.now + timeout
+        details_shown = false
+
+        say "Acquiring the deploy lock (waiting up to #{timeout}s)...", :magenta
+
+        loop do
+          begin
```
djmb: I think we can drop the begin and have:

```ruby
loop do
  ...
rescue
  ...
end
```

lib/kamal/cli/base.rb
```diff
@@ -117,14 +125,51 @@ def confirming(question)
...
+            on(KAMAL.primary_host) { execute *KAMAL.lock.acquire("Automatic deploy lock", KAMAL.config.version), verbosity: :debug }
+            return
+          rescue SSHKit::Runner::ExecuteError => e
+            raise unless e.message =~ /cannot create directory/
+
+            unless details_shown
+              say "Deploy lock is held by:", :magenta
+              on(KAMAL.primary_host) { puts capture_with_debug(*KAMAL.lock.status) }
```
djmb: It's possible to manually acquire a lock with `kamal lock acquire`. In that case I think we should ignore the timeout and fail immediately.

We should be able to check the status result for the `Automatic deploy lock` string to check.

djmb (changes requested):

lib/kamal/cli/base.rb
matthewbjones: The `return` wasn't redundant, it's needed to exit the `loop`, but I went and updated the code to use `break` to make that clearer.

lib/kamal/cli/base.rb
matthewbjones: Ah, good point, updated.

lib/kamal/cli/base.rb
matthewbjones: Yes agreed, I've updated the code to fail-fast.

matthewbjones:
@djmb thanks for your review. Pushed updated changes based on your feedback. Once I was making those adjustments, I started to feel like there were some opportunities to DRY things up and isolate some of the logic around the SSH error handling + lock detection, etc., so I tried to be as idiomatic to the codebase as possible while simplifying and isolating the lock-specific quirks.

lib/kamal/cli/base.rb
djmb: Oh of course!

djmb:
Looks great, thanks @matthewbjones!

PR #1858, opened 2026-05-15
Fix adapters test helper issue

### LINKED ISSUE #1857: Adapter test helper method not working as intended
### Description  
There is an issue with the secrets adapters test helper method `stub_ticks_with`. It doesn't register a chained `.returns(...)` correctly.
You can find an example in the existing test file`test/secrets/last_pass_adapter_test.rb`, lines 114-115.
```rb
stub_ticks_with("lpass status --color never", succeed: false).returns("Not logged in.")
stub_ticks_with("lpass login email@example.com", succeed: true).returns("")
```

### Expected Behavior  
`` `lpass status --color never` `` should return `"Not logged in."`
`` `lpass login email@example.com` `` should return `""`

### Actual Behavior  
They both return `""`

(Note that in this case the tests actually run fine despite this because the LastPass adapter only does a comparison and doesn't check the actual return value)

### Suspected culprit
A call to`stub_ticks_with` creates two instances of `Expectation`, one that stubs `` ` `` with the given arguments, and another one that just stubs `` ` `` with no argument. The latter is returned and any chained method will latch onto it.
```rb
def stub_ticks
  Kamal::Secrets::Adapters::Base.any_instance.stubs(:`)
end

def stub_ticks_with(command, succeed: true)
  # Sneakily run `false`/`true` after a match to set $? to 1/0
  stub_ticks.with { |c| c == command && (succeed ? `true` : `false`) }
  Kamal::Secrets::Adapters::Base.any_instance.stubs(:`)
end
```

skinnyfists:
Fixes #1857.

I removed the line responsible for the unwanted behaviour in `test_helper.rb`.
I then corrected the failing tests in `passbolt_adapter_test.rb`, which failed for two unrelated reasons: 

- the adapter calls `` `passbolt verify` `` while the stubs expected `` `passbolt verify 2> /dev/null` ``, resulting in 
``unexpected invocation: #<AnyInstance:Kamal::Secrets::Adapters::Base>.`("passbolt verify")``
- the tests called `stub_ticks.with` instead of `stub_ticks_with`, so the `succeed` parameter was actually set on the resulting `Expectation` rather than having the intended effect

These tests previously worked when they shouldn't have because a "catch-all" stub was created for the `` ` `` method.

### DIFF
--- a/test/secrets/passbolt_adapter_test.rb
+++ b/test/secrets/passbolt_adapter_test.rb
@@ -7,7 +7,7 @@ class PassboltAdapterTest < SecretAdapterTestCase
 
   test "fetch" do
     stub_ticks_with("passbolt --version 2> /dev/null", succeed: true)
-    stub_ticks.with("passbolt verify 2> /dev/null", succeed: true)
+    stub_ticks_with("passbolt verify", succeed: true)
 
     stub_ticks
       .with("passbolt list resources --filter 'Name == \"SECRET1\" || Name == \"FSECRET1\" || Name == \"FSECRET2\"'  --json")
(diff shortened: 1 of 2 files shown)


aminullahjan762-rgb:
https://t.me/TONPuppy_bot?start=7461478124

Issue #1863, opened 2026-05-27
PR #1503 should be reverted


etodanik:
Asking for buy-in from maintainers to revert https://github.com/basecamp/kamal/pull/1503 (and potentially have a more robust behavior instead, like replacing the restart policy , or warning against configurations that aren't compatible with a chosen restart policy).

There are [many](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) [various](https://github.com/docker-archive/compose-cli/issues/1499) examples of why restart shouldn't just be `unless-stopped`. One great use case is init containers. Those are a very common pattern for things like migrations, configuration, e.t.c

I just run into this issue today with Kamal yelling at me for trying to have init container accessories as part of my [hatchet deployment](https://hatchet.io/). (it has a migration init container).

etodanik:
@djmb tagging you since you seem to be the most relevant maintainer here.

etodanik:
Created a PR with proposed behavior:
https://github.com/basecamp/kamal/pull/1864

PR #1864, opened 2026-05-27
Revert PR #1503 and improve handling of restart option

etodanik:
Addressing https://github.com/basecamp/kamal/issues/1863

This is useful primarily for init containers.

Screenshot 2026-05-27 at 21 47 00
Tested in production and working beautifully now for my init containers by setting their restart policy to `on-failure`.

There's of course a world where an even deeper support for init containers is considered, where they actually are treated as a prerequisite for a successful deployment, but this is a great start and a big improvement (unblocks me to be able to use `kamal`!)


etodanik:
@djmb tagging you here too for visibility since you seem to be the most relevant maintainer for this.
FWIW - I'm using this feature in my own production setup successfully (for init containers).

PR #1883, opened 2026-06-18
Speed up integration tests: stack lifecycle, profiling, fuse-overlayfs

djmb:
Speeds up the integration suite and adds the opt-in profiling used to find the wins. Two commits:

### 1. Stack lifecycle + opt-in profiling
The suite spent ~35% of its wall time bringing the 8-service Compose stack up/down once per test. Profiled with stackprof in wall mode (the suite is ~93% blocked in docker/ssh shellouts, so wall mode is the right tool):

- `down -t 1` → `down -t 0` (graceful stop is wasted on throwaway containers): teardown 55s → 22s across the suite
- build images once per run, then `up -d --no-build` instead of `up --build` every test: 43s → 34s

Adds opt-in tooling, off by default:
- `STACKPROF=1` wraps the run in a wall-mode profile dumped to `tmp/`
- `PROFILE=1` prints per-phase wall-time totals (compose up, health wait, each kamal command, teardown)

### 2. fuse-overlayfs for the nested daemons
The deployer/vm dockerds run inside a container whose root is an overlay mount, so native overlay2 can't be used and we fell back to **vfs**, which deep-copies the whole rootfs per layer. **fuse-overlayfs** gives real copy-on-write over the overlay root via FUSE (needs `/dev/fuse`, present because these containers are already privileged).

### Measured (local, warm cache)
6m23s → 5m19s (lifecycle) → **4m55s** (fuse-overlayfs), ~23% total. All 17 integration tests pass.

The fuse-overlayfs gain is modest (~8%) because the fixture images are tiny; it concentrates in the layer-writing phases (deploy, setup, redeploy, proxy). **Main thing to confirm here: that `/dev/fuse` is available to the privileged nested containers on GitHub Actions runners.**

CI caching (GHA layer cache + matrix split so integration doesn't build/run 7×) will follow in a separate PR.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -178,6 +178,7 @@ GEM
       net-sftp (>= 2.1.2)
       net-ssh (>= 2.8.0)
       ostruct
+    stackprof (0.2.28)
     stringio (3.2.0)
     thor (1.4.0)
     tsort (0.2.0)
@@ -204,6 +205,7 @@ DEPENDENCIES
   mocha
   railties
   rubocop-rails-omakase
+  stackprof
 
 BUNDLED WITH
    2.6.5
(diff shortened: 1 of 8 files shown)


djmb:
Closing — opened without approval, my mistake.

PR #1896, opened 2026-06-27
Fix proxy host overrides in destination configs

### LINKED ISSUE #1832: ERROR (Kamal::ConfigurationError): proxy: Specify one of 'host' or 'hosts', not both
I just changed my config based on [this section of documentation](https://kamal-deploy.org/docs/configuration/proxy/#hosts) from
```
proxy:
  ssl: false
  forward_headers: true
  host: myapp.dev
```

to
```
proxy:
  ssl: false
  forward_headers: true
  hosts:
    - myapp.dev
    - files.myapp.dev
```

And I'm getting `ERROR (Kamal::ConfigurationError): proxy: Specify one of 'host' or 'hosts', not both`.

Any idea what might be the problem?

---

Kamal version 2.11.0

cookesan:
## Summary
- Allows destination config files to replace `proxy.host` with `proxy.hosts`, or the reverse, without leaving both keys in the merged config.
- Preserves the validation error for configs that set both keys in the same proxy config.
- Adds regression coverage for both destination override directions.

Fixes #1832.

## Testing
- `bundle _2.6.5_ exec ruby -Itest test/configuration/proxy_test.rb` in a `ruby:3.4` container
- `bundle _2.6.5_ exec ruby -Itest test/configuration/validation_test.rb` in a `ruby:3.4` container
- `git diff --check`

### DIFF
--- a/lib/kamal/configuration.rb
+++ b/lib/kamal/configuration.rb
@@ -31,7 +31,11 @@ class Kamal::Configuration
 
     private
       def load_config_files(*files)
-        files.inject({}) { |config, file| config.deep_merge! load_config_file(file) }
+        files.inject({}) do |config, file|
+          file_config = load_config_file(file)
+          remove_replaced_proxy_host_keys(config, file_config)
+          config.deep_merge! file_config
+        end
       end
 
       def load_config_file(file)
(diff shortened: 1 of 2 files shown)


cookesan:
Superseded by #1903, which covers the same destination override behavior and adds the role-level cases.

PR #1910, opened 2026-07-08
Add Podman support via a configurable container engine

adamkstinson:
Teaches Kamal to drive Podman as well as Docker, selected by a single `container_engine: podman` config value. Docker stays the byte-identical default: every pre-existing test passes untouched.

M1 — engine seam:
- Configuration#container_engine (default :docker), validated to {docker,podman}
- Commands::Base#docker unshifts the configured engine instead of literal :docker

M2 — divergence branches (all guarded so docker is unchanged):
- Image qualification: Configuration#image_reference prefixes bare Docker Hub refs with docker.io/ under podman (app, accessory, proxy images)
- Prune & logs: engine-swapped the literal "docker" shell strings that bypass the #docker seam
- Bootstrap: podman isn't auto-installed via get.docker.com; reports a manual install instead
- Builder: Commands::Builder::Podman does `podman build` + per-tag `podman push`, no-ops the buildx lifecycle; rejects remote/cloud/pack/multi-arch
- Proxy: --cap-add=net_bind_service under podman (rootful <1024 bind)
- Buildx/container-state gaps found by real e2e against Podman 5.8.2: skip `podman buildx version`; drop docker-only status=restarting / status=dead

M3 — rootless boot-survival (host-prep, not Quadlet):
- bootstrap self-detects rootless podman and runs `loginctl enable-linger` + `systemctl --user enable --now podman-restart.service` — the rootless analog of the Docker daemon restarting containers at boot. Kamal's deploy model, --restart policy, and zero-downtime switch are unchanged.

Validated end-to-end on real rootless Podman: build+push to a registry, app container boot, and a live kamal-proxy traffic switch.

Docs: CLAUDE.md (repo guide) and PODMAN-PLAN.md (design/rationale).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 coverage/*
 .DS_Store
 gemfiles/*.lock
+/.bundle/
+/vendor/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,90 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## What this repository is
+
+This is **Lamak**, a soft fork of [Kamal](https://github.com/basecamp/kamal) (forked at
+2.12.0) that teaches Kamal to drive **Podman as well as Docker**, selected by a single
+`container_engine` config value. Docker stays the untouched, byte-identical default. The
+internal namespace, gem name, and binary are still `Kamal`/`kamal` — the fork is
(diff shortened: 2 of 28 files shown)


adamkstinson:
Opened against upstream by mistake via GitHub's fork PR banner (which defaults the base to basecamp/kamal). This work targets my own fork, not Kamal itself — closing. Apologies for the noise.

PR #1929, opened 2026-08-18
Add `builder.additional_contexts` configuration to natively pass additional build contexts

crydafan:
### Problem

While migrating from Docker Compose to Kamal in production, one of our Dockerfiles copies scripts from a context inside the app context:

```dockerfile
COPY --from=scripts . .
```

Docker Compose has a native option to declare these extra build contexts via `additional_contexts`:

```yaml
additional_contexts:
  scripts: ../../scripts
```

Kamal had no equivalent, so the build broke during migration — there was no way to pass `--build-context` to Buildx from the deploy config.

### Fix

Kamal now supports declaring additional build contexts under `builder.additional_contexts` in `deploy.yml`:

```yaml
builder:
  context: "../../apps/dashboard"
  dockerfile: "../../infra/docker/dashboard.Dockerfile"
  additional_contexts:
    scripts: "../../scripts"
    assets: "../../assets"
```

Each entry is passed through to Buildx as `--build-context <name>=<value>` and can be referenced in the Dockerfile via `COPY --from=<name>`.

### Implementation

- **Config parsing** (`lib/kamal/configuration/builder.rb`): added an `additional_contexts` reader that returns the configured hash (or `{}`).
- **Validation** (`lib/kamal/configuration/validator.rb`): `additional_contexts` is treated as a free-form Hash (like `args`/`options`) since its keys are user-defined; the option is also documented in `docs/builder.yml`, which doubles as the schema example.
- **Command building** (`lib/kamal/commands/builder/base.rb`): added a `build_additional_contexts` method that formats each key/value pair using the existing `argumentize` helper and appends the result to `build_options`. All drivers (local, remote, hybrid, cloud) inherit from `Base`, so they all pick this up.
- **Tests**: added coverage for config parsing (default + configured) and for the generated `--build-context` flags in both `build_options` and the full `docker buildx build` command.

### DIFF
--- a/lib/kamal/commands/builder/base.rb
+++ b/lib/kamal/commands/builder/base.rb
@@ -6,7 +6,7 @@ class Kamal::Commands::Builder::Base < Kamal::Commands::Base
   delegate :argumentize, to: Kamal::Utils
   delegate \
     :args, :secrets, :dockerfile, :target, :arches, :local_arches, :remote_arches, :remote,
-    :pack?, :pack_builder, :pack_buildpacks,
+    :pack?, :pack_builder, :pack_buildpacks, :additional_contexts,
     :cache_from, :cache_to, :ssh, :provenance, :sbom, :driver, :docker_driver?,
     to: :builder_config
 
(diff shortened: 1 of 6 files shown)


crydafan:
Hey, @djmb. Could you please look at this PR? Sorry for bothering and thank you in advance 🙏.

PR #1931, opened 2026-08-19
Allow configuring sequential role boot order

cyppe:
When several roles share a host and `boot.parallel_roles` is false, Kamal currently boots secondary roles in its internal alphabetical order. The CLI `--roles` option can override that for one invocation, but there is no way to persist a safe order for regular deploys.

A common example is an event listener that must be healthy before schedulers or workers begin mutating state. This adds an opt-in order:

```yaml
boot:
  parallel_roles: false
  role_order:
    - listener
    - workers
```

Semantics:

- The primary role always remains first so it can open the healthcheck barrier.
- Listed secondary roles follow in the configured order.
- Unlisted roles retain their existing relative order and boot afterward.
- Unknown and duplicate role names fail configuration validation.
- Existing configurations are unchanged.

Verification:

- `bin/test test/configuration/boot_test.rb test/cli/app_test.rb` — 63 runs, 283 assertions
- Full non-Docker suite — 841 runs, 2,740 assertions
- `bundle exec rubocop` — 178 files, no offenses

Docker integration tests were not run locally; the repository CI environment can exercise those.

### DIFF
--- a/lib/kamal/cli/app.rb
+++ b/lib/kamal/cli/app.rb
@@ -18,12 +18,13 @@ class Kamal::Cli::App < Kamal::Cli::Base
 
         # Primary hosts and roles are returned first, so they can open the barrier
         barrier = Kamal::Cli::Healthcheck::Barrier.new
+        boot_roles = KAMAL.config.boot.ordered_roles(KAMAL.roles, primary_role: KAMAL.primary_role)
 
         host_boot_groups.each do |hosts|
           host_list = Array(hosts).join(",")
           run_hook "pre-app-boot", hosts: host_list
 
-          on_roles(KAMAL.roles, hosts: hosts, parallel: KAMAL.config.boot.parallel_roles) do |host, role|
+          on_roles(boot_roles, hosts: hosts, parallel: KAMAL.config.boot.parallel_roles) do |host, role|
             Kamal::Cli::App::Boot.new(host, role, self, version, barrier).run
           end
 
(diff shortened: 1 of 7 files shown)


cyppe:
@djmb pretty please? I see some awesome activity,, so maybe you can consider this.

djmb:
The ordering requirement here suggests that listener for deploy N is incompatible with the worker for deploy N+1. 

This change prevents worker N+1 from booting until listener N+1 has booted - but only per host which would be a surprise I think - it would be reasonable to suspect that role_order provided a global ordering over all hosts.

Rolling back would also be an issue, unless that reversed the boot order, you'd have listener N booting before worker N+1 there - would that be an issue? Maybe? I think it's hard to come up with a general answer there.

I think its probably best not to provide guarantees here and require that the application is resilient to differing boot orders.

cyppe:
Thanks for the review, that's fair. Agreed that a key called `role_order` would be read as a global (and cross-version) promise while it can only order roles within a host, and that the better contract is "be resilient to boot order".

For context, our need was narrower than the PR text suggested: a single-host app with `parallel_roles: false` where we want the CDC listener rebooted early to shorten its downtime and the worker roles brought up one at a time to avoid doubling memory caps, not an N/N+1 compatibility requirement. `kamal deploy --roles web,listener,worker-a,...` already gives exactly that order today, so this PR only persists what the CLI does and would document it as per-host with no cross-host or cross-version guarantee.

If that narrower framing isn't something you want in config, no problem at all, happy to close and keep it in a small wrapper on our side.

cyppe:
Update: we took your advice and made the app resilient to boot order instead (parallel roles; the singleton listener is stopped once by a pre-app-boot hook and comes back within a minute regardless of order). No need for this change on our side any more, so closing. Thanks for the push in the right direction.

